PR #16, opened 2020-04-27
Add externals for Boost, Alpaka and Cupla

fwyzard:


makortel:
(for future reference) I was able to build all the added `external` targets both directly and via dependencies from `external_cupla`.

PR #18, opened 2020-04-27
Build a local copy of Boost only if the system version is too old

fwyzard:


makortel:
I tested that the mechanism works.

PR #19, opened 2020-04-28
Minimal Skeleton for Alpaka pixel reco

felicepantaleo:
A minimal skeleton for the Alpaka pixel reconstruction was created. 
Only the `Framework` and `DataFormats` were copied from the `cuda` folder.


src/alpaka/Makefile
```diff
@@ -0,0 +1,286 @@
+TARGETS = alpaka
```
makortel, line 1: This file looks mostly a copy from the pixel-standalone. Do I guess correctly that this is a work in progress and will be updated to be part of the build system of this repo at some point soon?

makortel:
Just to note that `DataFormats` and `Framework` are equal to `cuda` program, `AlpakaCore` is added, `plugin-SiPixelClusterizer` has minimal content, and `CondFormats` and `bin` are missing (as are other plugin directories than the clusterizer).

makortel:
Merging anyway

src/alpaka/Makefile
felicepantaleo: yes, precisely

felicepantaleo:
> Just to note that `DataFormats` and `Framework` are equal to `cuda` program, `AlpakaCore` is added, `plugin-SiPixelClusterizer` has minimal content, and `CondFormats` and `bin` are missing (as are other plugin directories than the clusterizer).

I'm trying to follow the minimal approach of adding things only when needed. Now we are working on the clusterizer. As soon as we have something else we will add the other folders

PR #22, opened 2020-05-02
Improve boost version check in Makefile

makortel:
With the earlier syntax I get
```
/bin/sh: 1: 106501: not found
```
on my laptop with boost 1.65.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ export EIGEN_LDFLAGS :=
 ifeq ($(wildcard /usr/include/boost/version.hpp),)
 NEED_BOOST := true
 else
-NEED_BOOST := $(shell (( $$(cat /usr/include/boost/version.hpp | grep '\#define BOOST_VERSION\>' | awk '{ print $$3 }') > 106300 )) || echo true)
+NEED_BOOST := $(shell test $$(cat /usr/include/boost/version.hpp | grep '\#define BOOST_VERSION\>' | awk '{ print $$3 }') -ge 106300 || echo true)
 endif
 ifeq ($(NEED_BOOST),true)
 BOOST_BASE := $(EXTERNAL_BASE)/boost


Makefile
```diff
@@ -58,7 +58,7 @@ export EIGEN_LDFLAGS :=
 ifeq ($(wildcard /usr/include/boost/version.hpp),)
 NEED_BOOST := true
 else
-NEED_BOOST := $(shell (( $$(cat /usr/include/boost/version.hpp | grep '\#define BOOST_VERSION\>' | awk '{ print $$3 }') > 106300 )) || echo true)
+NEED_BOOST := $(shell test $$(cat /usr/include/boost/version.hpp | grep '\#define BOOST_VERSION\>' | awk '{ print $$3 }') -ge 106300 || echo true)
```
makortel, line 61: This should actually be
```suggestion
NEED_BOOST := $(shell test $$(cat /usr/include/boost/version.hpp | grep '\#define BOOST_VERSION\>' | awk '{ print $$3 }') -gt 106300 || echo true)
```
to preserve the earlier intent.

fwyzard:
I guess your laptop (like mine) runs a Debian or Ubuntu derivative ?

RedHat uses `bash` for its `/bin/sh` shell, so Make is able to use "bash-ism" in its `$(shell ...)` function there.

Debian uses `dash` for its `/bin/sh` shell, which is much closer to the original `sh` behaviour.

Makefile
fwyzard, line 61: Simplified to use only `awk`:
```suggestion
NEED_BOOST := $(shell awk '/\#define BOOST_VERSION\>/ { print ($$3 > 106300) ? "false" : "true" }' /usr/include/boost/version.hpp )
```

fwyzard: Actually, re-reading the Alpaka documentation, it looks like the minimum version of Boost they require is 1.65.1:
Simplified to use only `awk`:
```suggestion
NEED_BOOST := $(shell awk '/\#define BOOST_VERSION\>/ { if ($$3 < 106501) print "true" }' /usr/include/boost/version.hpp )
```

makortel:
> I guess your laptop (like mine) runs a Debian or Ubuntu derivative ?

Correct.

makortel:
Closing in favor of #27

PR #26, opened 2020-05-04
First version of alpakatest program

makortel:
Includes
* build rules to compile code in `alpaka` subdirectory to CPU serial, CPU TBB, and CUDA backends
   * `main()` has command line arguments to select which backend(s) to use
* all three test producers exist
* there is some code running kernel through Alpaka

Missing (i.e. to be done later)
* complete the kernel calls (wrt. `cudatest`)
* pass data from one producer to another via Event in some way
* device framework

FYI @fwyzard


src/alpakatest/plugin-Test1/alpaka/alpakaAlgo1.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "AlpakaCore/alpakaConfig.h"
+
+// In principle this could be ALPAKA_ARCHITECTURE_NAMESPACE, but how
+// to compile it only once for CPU? Current build rules build separate
+// objects for serial and TBB accelerators. Do we need to have one set
+// of build rules for files to be compiled by accelerator, and
+// different set for files to be compiled by architecture?
+namespace ALPAKA_ACCELERATOR_NAMESPACE {
+  void alpakaAlgo1();
```
makortel, line 12: @fwyzard Do you have any thoughts about the comment above?

makortel:
Merging now, we can discuss the details later.

PR #31, opened 2020-05-11
[Alpaka] Add prefix scan and its test

felicepantaleo:


makortel:
I don't have anything specific to comment, so I'll wait for @waredjeb to review before merging.

Currently the build system does not support any (unit) tests. Would (limited) support for (unit) tests be useful?

PR #33, opened 2020-05-11
Update Kokkos to 3.1.1 and initialize and use only the backend(s) given at the command line

makortel:
The Kokkos update itself was seamless, and initializing only the selected backend(s) was straightforward. I also had to give the execution space object as a parameter to `Kokkos::deep_copy()` in order to have it call the `fence()` of the execution space instead of the global `Kokkos::fence()`.

Note that the update of Kokkos requires `make clean; rm -fR external/kokkos`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,7 +250,7 @@ $(CUPLA_BASE)/lib: $(CUPLA_BASE) $(ALPAKA_DEPS) $(BOOST_DEPS) $(TBB_DEPS) $(CUDA
 external_kokkos: $(KOKKOS_LIB)
 
 $(KOKKOS_SRC):
-	git clone --branch 3.0.00 https://github.com/kokkos/kokkos.git $@
+	git clone --branch 3.1.01 https://github.com/kokkos/kokkos.git $@
 	cd $(KOKKOS_SRC) && patch -p1 < ../../../nvcc_wrapper.patch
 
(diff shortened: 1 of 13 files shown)


makortel:
It turned out that Kokkos 3.1.1 does not run on Turing (https://github.com/kokkos/kokkos/issues/3026), so I'll close this PR and make another one that calls the `fence()` only on the execution space.

crtrott:
Note: you can compile with Kokkos_ENABLE_TURING75 to get Turing compilation. And in 3.0 the deep_copy behavior wasn't correct with regards to fencing. I.e. you will get more fencing than you like for the version which takes an execution space argument.

makortel:
Thanks @crtrott for the additional details (I also clarified my comment above). For time being I prefer to get away with a common build configuration that runs on both Volta and Turing. If we manage to get to any performance measurements before 3.2 I'll revisit this short cut to avoid the extra fences.

makortel:
Hmm, seems that passing the execution space to `deep_copy()` leads to
```
Error: Kokkos::deep_copy of overlapping views: d_a(198489152,198493152) d_a(198489152,198493152)
```
with the serial backend (with Kokkos 3.0). I did not see such an error in 3.1.1, so I'll give up on the `deep_copy()` changes as well.

PR #34, opened 2020-05-11
porting countModules to Kokkos

jtchilders:
I updated the `gpuClustering::countModules()` function from cuda to kokkos as a first test for me.

### DIFF
--- a/src/kokkos/plugin-SiPixelClusterizer/kokkos/gpuClustering.h
+++ b/src/kokkos/plugin-SiPixelClusterizer/kokkos/gpuClustering.h
@@ -32,8 +32,8 @@ namespace KOKKOS_NAMESPACE {
         --j;
       if (j < 0 or id[j] != id[index]) {
         // boundary... replacing atomicInc with explicit logic
-        auto loc = Kokkos::atomic_compare_exchange(moduleStart,::gpuClustering::MaxNumModules,0);
-        Kokkos::atomic_increment(moduleStart);
+        auto loc = Kokkos::atomic_fetch_add(&moduleStart[0],1);
+        assert(moduleStart[0] < ::gpuClustering::MaxNumModules);
         moduleStart[loc + 1] = index;
       }
     }
--- a/src/kokkos/plugin-SiPixelClusterizer/kokkos/gpuClustering.h
+++ b/src/kokkos/plugin-SiPixelClusterizer/kokkos/gpuClustering.h
@@ -15,31 +15,34 @@
 namespace KOKKOS_NAMESPACE {
   namespace gpuClustering {
 
-#ifdef TODO
+
 #ifdef GPU_DEBUG
     __device__ uint32_t gMaxHit = 0;
 #endif
 
-    KOKKOS_INLINE_FUNCTION void countModules(uint16_t const* __restrict__ id,
-                                             uint32_t* __restrict__ moduleStart,
-                                             int32_t* __restrict__ clusterId,
-                                             int numElements) {
-      int first = blockDim.x * blockIdx.x + threadIdx.x;
-      for (int i = first; i < numElements; i += gridDim.x * blockDim.x) {
-        clusterId[i] = i;
-        if (InvId == id[i])
-          continue;
-        auto j = i - 1;
-        while (j >= 0 and id[j] == InvId)
+    KOKKOS_INLINE_FUNCTION void countModules(
+            Kokkos::View<uint16_t const*,KokkosExecSpace> id,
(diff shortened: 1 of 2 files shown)


src/kokkos/plugin-SiPixelClusterizer/kokkos/gpuClustering.h
```diff
@@ -15,31 +15,32 @@
...
+      while (j >= 0 and id[j] == ::gpuClustering::InvId)
+        --j;
+      if (j < 0 or id[j] != id[index]) {
+        // boundary... replacing atomicInc with explicit logic
+        auto loc = moduleStart[0];
+        if (moduleStart[0] >= ::gpuClustering::MaxNumModules)
+          moduleStart[0] = 0;
+        else
+          moduleStart[0] = moduleStart[0] + 1;
+        moduleStart[loc + 1] = index;
```
makortel, line 40: I'm pretty sure the "reading of `loc`" and "updating `moduleStart[0]`" need to stay in one atomic transaction.

jtchilders: Yes, you're right. I will change this. 

The Cuda `atomicInc` does
`((old >= val) ? 0 : (old+1))`
in one operation. This was used in the original. 

However, I don't see a similar Kokkos atomic. There is an `atomic_increment()` but it does not do the comparison. 
``` c
          auto loc = moduleStart[0];
          if( moduleStart[0] >= ::gpuClustering::MaxNumModules)
            Kokkos::atomic_assign(moduleStart,0);
          else:
            Kokkos::atomic_increment(moduleStart,0);
          moduleStart[loc + 1] = index;
```

jtchilders: I think this will ensure  the proper atomics and behavior of the cuda version.
```c
auto loc = Kokkos::atomic_compare_exchange(moduleStart,::gpuClustering::MaxNumModules,0);
Kokkos::atomic_increment(moduleStart);
moduleStart[loc + 1] = index;
```

makortel: Thanks, but I'm not sure if this is sufficient. E.g. if two threads hit into `atomic_compare_exchange` before either of them go to `atomic_increment`, they get the same value for `loc`, right?

The `atomic_fetch_add` would correctly combine the "return old value and increase", but misses the wrapping over `MaxNumModules`. On the other hand, I think the wrapping is mostly a sanity check to to prevent out-of-bounds access on `moduleStart`. And in fact, looking from the in-house CUDA-to-CPU translation header
https://github.com/cms-patatrack/cmssw/blob/a75d7c6abdf37434a737fc09aef05171b5e4d81a/HeterogeneousCore/CUDAUtilities/interface/cudaCompat.h#L30-L36
it looks to me that the even the wrapping behavior is not needed, but preventing out-of-bounds is sufficient.

So maybe combining `atomic_fetch_add` with a local check `if (loc >= ::gpuClustering::MaxNumModules) loc = 0` (or `loc = ::gpuClustering::MaxNumModules-1`) could be sufficient?

@VinInn Could you confirm if I understood the intended use of `atomicInc()` correctly?

jtchilders: OK, I wasn't quite sure if the wrapping check was just a precaution or if it was expected that wrapping is necessary.

VinInn: I think it was historical. it should be an assert actually. I personally do not like the behaviour of atomicInc...
I am not sure what your aim is here: targetting GPU, CPU serial or CPU MT? or trying CPU SIMD?

the logic is simple: if moduleId changes the module changes nd just increase the counter 
How? depends on the architecture and the concurrency model.

makortel: Thanks @VinInn.

> I think it was historical. it should be an assert actually. I personally do not like the behaviour of atomicInc...

Ok, so the following would actually be the preferred behavior.
```cpp
auto loc = Kokkos::atomic_fetch_add(moduleStart, 1);
assert(loc < ::gpuClustering::MaxNumModules);
```

> I am not sure what your aim is here: targetting GPU, CPU serial or CPU MT? or trying CPU SIMD?

Portable code that would ideally support all of those. In practice for now (with Kokkos) that means supporting CPU serial and CUDA.

PR #38, opened 2020-05-13
[alpaka] code format

makortel:


makortel:
@waredjeb @felicepantaleo any objections?

felicepantaleo:
nope!

PR #39, opened 2020-05-15
Add support for unit tests, include several for CUDA program

### LINKED ISSUE #35: Add support for unit tests
Unit tests would help porting kernels that use more complex data structures like `HistoContainer`.

makortel:
This PR adds support for unit tests for all programs (except `alpaka`, which is not yet integrated in the common build system), and some unit tests (dummy for all `*test` programs to exercise the build rules, some real ones for `cuda` and `kokkos`).

The unit tests are built while building the corresponding program (technically each program executable is set to depend on the unit test executables). Running these unit test programs is included in the `make test` target, and also properly in the `test_cpu` and `test_cuda` targets.

Resolves #35.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 external
 obj
 lib
+./test
 .original_env
 env.sh
 data/*.tar.gz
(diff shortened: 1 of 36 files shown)


makortel:
Passed tests, but the build rules are not fully correct (something gets built with `make test` after `make all`). I'll investigate those tomorrow.

PR #46, opened 2020-05-18
Adding Atomic pair counter

jtchilders:
includes class definition and unittest.


makortel:
(just to document) Out of the box this PR fails to link
```
nvlink error   : Undefined reference to '_ZN6Kokkos4Impl25g_device_cuda_lock_arraysE' in '..../pixeltrack-standalone/obj/kokkos/test/kokkos/AtomicPairCounter_t.cc.cuda.o'
```
because Kokkos is currently built as a shared library, and device link ignores code from shared objects.

src/kokkos/test/kokkos/AtomicPairCounter_t.cc
```diff
@@ -0,0 +1,75 @@
+#include "KokkosCore/kokkosConfigCommon.h"
```
makortel, line 1: After #47 this needs to be changed to
```suggestion
#include "KokkosCore/InitializeScopeGuard.h"
```

makortel:
Now the test builds, but both `serial` and `cuda` versions fail to run
```
$ ./test/kokkos/AtomicPairCounter_t.serial
size 8
Kokkos::abort: Requested Team Size is too large!Aborted (core dumped)

$ ./test/kokkos/AtomicPairCounter_t.cuda
.../pixeltrack-standalone/src/kokkos/test/kokkos/AtomicPairCounter_t.cc:45: lambda [](const Kokkos::Impl::CudaTeamMember &)->auto::operator()(const Kokkos::Impl::CudaTeamMember &)->auto: block: [0,0,0], thread: [0,0,0] Assertion `dc_d(0).get().m == n` failed.
  what():  cudaDeviceSynchronize() error( cudaErrorAssert): device-side assert triggered .../pixeltrack-standalone/external/kokkos/source/core/src/Cuda/Kokkos_Cuda_Instance.cpp:139
```

jtchilders:
OK, serial and cuda both work now and give consistent results with CUDA version of the unit test.

src/kokkos/KokkosCore/AtomicPairCounter.h
```diff
@@ -0,0 +1,46 @@
+#ifndef HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
+#define HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
+
+#include <cstdint>
+#include "KokkosCore/InitializeScopeGuard.h"
```
makortel, line 5: This `#include` should go in the unit test source file, and the `#include "KokkosCore/kokkosConfigCommon.h"` can be removed there.

src/kokkos/test/kokkos/AtomicPairCounter_t.cc
```diff
@@ -0,0 +1,81 @@
...
+  Kokkos::View<uint32_t*,KokkosExecSpace> m_d("m_d",M);
+
+  const uint32_t n = 10000;
+#ifdef KOKKOS_BACKEND_CUDA
+  auto league_size = n/10;
+#elif KOKKOS_BACKEND_SERIAL
+  auto league_size = n*2;
+#else
+  auto league_size = n/2;
+#endif
```
makortel, line 31: I'm a bit unhappy about having to do this (especially with `#ifdef`). Do I understand correctly that both `league_size` and `team_size` can not be `Kokkos::AUTO()`? Maybe this calls for some abstraction to hide the `#ifdef` from user code (to be figured out later).

jtchilders: We don't have to do this. I'll remove this. Mainly we just need to ensure there are enough thread executions to reach `10000` in all cases. However, since I cannot know what `Kokkos::AUTO()` will be, I can only ensure this by making `league_size = n` for all cases. This means for parallel cases we sorely over-execute the work.

jtchilders: It really raises the question of whether or not I should use `Kokkos::AUTO()` since I cannot predict how much work it will produce. If I need to execute my `parallel_for` N times, and I can set the `league_size` but the `team_size` will be set to some arbitrary value, how do I ensure the loop executes N times?

PR #48, opened 2020-05-22
[kokkos] Go back to a dynamic library per package by disabling relocatable device code

makortel:
A dynamic library per package fits better into CMSSW, so let's see how far we get without relocatable device code.

An existing build requires `make clean external_kokkos_clean`.


makortel:
FYI @fwyzard in case you'd be interested.

PR #52, opened 2020-05-27
Add BeamSpotToKokkos

PointKernel:
Do I need to create a unit test as well?

### DIFF
--- a/src/kokkos/plugin-BeamSpotProducer/kokkos/BeamSpotToKokkos.cc
+++ b/src/kokkos/plugin-BeamSpotProducer/kokkos/BeamSpotToKokkos.cc
@@ -15,16 +15,15 @@ namespace KOKKOS_NAMESPACE {
 
   private:
     edm::EDPutTokenT<BeamSpotKokkos<KokkosExecSpace>> bsPutToken_;
-
-    typename Kokkos::View<BeamSpotKokkos<KokkosExecSpace>::Data, KokkosExecSpace>::HostMirror bsHost;
+    // remove bsHost for now
+    // typename Kokkos::View<BeamSpotPOD, KokkosExecSpace>::HostMirror bsHost;
   };
 
   BeamSpotToKokkos::BeamSpotToKokkos(edm::ProductRegistry& reg)
-      : bsPutToken_{reg.produces<BeamSpotKokkos<KokkosExecSpace>>()},
-        bsHost{"bsHost"} {}
+      : bsPutToken_{reg.produces<BeamSpotKokkos<KokkosExecSpace>>()} {}
 
   void BeamSpotToKokkos::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
-    auto bsRaw = iSetup.get<BeamSpotKokkos<KokkosExecSpace>::Data>();
+    auto const& bsRaw = iSetup.get<BeamSpotPOD>();
     BeamSpotKokkos<KokkosExecSpace> bs{&bsRaw};
 
     iEvent.emplace(bsPutToken_, std::move(bs));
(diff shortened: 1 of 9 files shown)


PointKernel:
Just realized I didn't use `BeamSpotToKokkos::bsHost` at all in the current code. Do we still need it in `BeamSpotToKokkos`? If yes, maybe we need a `BeamSpotKokkos` constructor with host mirror as input. What do you think?

makortel:
> Do I need to create a unit test as well?

No, unit test is not needed.

makortel:
> Just realized I didn't use `BeamSpotToKokkos::bsHost` at all in the current code. Do we still need it in `BeamSpotToKokkos`? If yes, maybe we need a `BeamSpotKokkos` constructor with host mirror as input. What do you think?

I think currently the `bsHost` does not provide any added value, so better remove now.

Alternatively one could argue that the `BeamSpotKokkos` is too complicated and could be replaced with something equivalent to `Kokkos::View<BeamSpotData::Kokkos, KokkosExecSpace>`, but I would leave it to later time.

src/kokkos/plugin-BeamSpotProducer/kokkos/BeamSpotToKokkos.cc
```diff
@@ -0,0 +1,34 @@
...
+
+    typename Kokkos::View<BeamSpotKokkos<KokkosExecSpace>::Data, KokkosExecSpace>::HostMirror bsHost;
+  };
+
+  BeamSpotToKokkos::BeamSpotToKokkos(edm::ProductRegistry& reg)
+      : bsPutToken_{reg.produces<BeamSpotKokkos<KokkosExecSpace>>()},
+        bsHost{"bsHost"} {}
+
+  void BeamSpotToKokkos::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
+    auto bsRaw = iSetup.get<BeamSpotKokkos<KokkosExecSpace>::Data>();
```
makortel, line 27:
```suggestion
    auto const& bsRaw = iSetup.get<BeamSpotKokkos<KokkosExecSpace>::Data>();
```
The `BeamSpotKokkos<KokkosExecSpace>::Data` looks weird but is technically correct. Maybe it would actually be clearer to rename `BeamSpotKokkos<...>::Data` to e.g. `BeamSpotPOD`, and use `Kokkos::View<BeamSpotPOD, KokkosExecSpace>` and use that everywhere.

makortel:
Could you also add
```cpp
edmodules.emplace_back(prefix + "BeamSpotToKokkos");
```
as a first element to
https://github.com/cms-patatrack/pixeltrack-standalone/blob/0071083b6c91d8abc99073d68d7c8e6e08e3b471/src/kokkos/bin/main.cc#L98-L99
and
```
kokkos_serial::BeamSpotToKokkos pluginBeamSpotProducer.so
kokkos_cuda::BeamSpotToKokkos pluginBeamSpotProducer.so
```
to https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/kokkos/plugins.txt

makortel:
The `BeamSpotESProducer`
https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/cuda/plugin-BeamSpotProducer/BeamSpotESProducer.cc
needs to be ported as well. In `main.cc` it needs to be added in `esmodules`
https://github.com/cms-patatrack/pixeltrack-standalone/blob/0071083b6c91d8abc99073d68d7c8e6e08e3b471/src/kokkos/bin/main.cc#L100
and it also needs an entry in `plugins.txt`.

Note that you can test running quickly with `./kokkos --maxEvents 2 --serial --cuda` (uses both Serial and Cuda backends).

Issue #43, opened 2020-05-18
Investigate the performance impact of CUDA managed memory


makortel:
I'm thinking to make a full version of `cuda` program that uses the CUDA managed memory (kind of follow-up to https://github.com/cms-patatrack/cmssw/issues/85). The question of the performance impact seems to affect both the portability technology exploration and the evolution of the CUDA products in EDM, so maybe it is time to make a proper evaluation.

Plan
- [x] Clone `cuda` to `cudauvm` done in #44 
- [x] Clone `CachingHostAllocator` to `CachingManagedAllocator`, create `cms::managed::unique_ptr` done in #49 
- [x] Migrate conditions done in #53 
- [x] Migrate beamspot transfer done in #54
- [x] Migrate raw-to-cluster done in #123 
- [x] Migrate rechits done in #162
- [x] Migrate CA done in #182 
- [x] Migrate vertexing done in #182
- [x] Evaluate performance, at least on T4 and V100

I'm thinking to have compile-time switches to
* enable/disable `cudaMemPrefetchAsync()`
* use device/managed memory for (temporary) device-only data structures

makortel:
The evaluation was reported in vCHEP 2021

PR #53, opened 2020-05-29
[cudauvm] Move conditions to use managed memory

makortel:
Part of #43. Kind of equivalent to https://github.com/cms-patatrack/cmssw/pull/157.

By default `cudaMemAdvise()` and `cudaMemPrefetchAsync()` are enabled, but there are compile-time switches to disable them.

### DIFF
--- a/run-scan.py
+++ b/run-scan.py
@@ -12,7 +12,8 @@ import multiprocessing
 n_events_unit = 1000
 n_blocks_per_stream = {
     "fwtest": 1,
-    "cuda": {"": 400, "transfer": 350}
+    "cuda": {"": 400, "transfer": 350},
+    "cudauvm": {"": 400, "transfer": 350},
 }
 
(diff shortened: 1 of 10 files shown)


makortel:
Passes tests, I want to assess performance impact (on V100) first before merging.

makortel:
Here is a throughput scan on a single V100 (all cases process 400k events / concurrent event, "1 concurrent event" case takes about 9 minutes)

cudauvm_pr53

Conclusions
* Managed memory gives 0-4 % smaller throughput than explicit memory 
  * Consistent with observations in https://github.com/cms-patatrack/cmssw/pull/157
* `cudaMemAdvise(cudaMemAdviseSetReadMostly)` has a 2-6 % effect until the throughput plateau is reached
   * at the plateau it does not make a difference
* Prefetch has no impact in this test
  * as expected, conditions data is transferred once and then used many many times

PR #54, opened 2020-06-02
[cudauvm] Migrate BeamSpot transfer to managed memory

makortel:
Part of #43.


makortel:
Here is a throughput scan on a single V100 (all cases process 400k events / concurrent event, "1 concurrent event" case takes about 9 minutes)
cudauvm_pr54

Conclusions
* With advise and prefech (top-right) there is no significant difference wrt. #53
* In the "advise disabled" case (middle-right) #54 actually improves throughput for small thread count
* In the "prefetch disabled" case (bottom-right) #54 decreases throughput up to the plateau (<= 10 threads), and increases above that

The throughput degradations (either wrt. explicit memory, or managed memory with advise and prefetch) appear to be clear in the regime where there are free "GPU resources", but become fuzzy when "GPU is fully utilized" (whetever these terms really mean in practice).

PR #55, opened 2020-06-08
[cudauvm] Need to unset memory advise in BeamSpotCUDA destructor

makortel:
While (slowly) working on the raw-to-cluster part, it occurred to me that it `cudaMemAdvise()` is used with a cached allocation, it better be un-done before returning the memory to the cache.

At first I got weird errors
```
  what():  
.../pixeltrack-standalone/src/cudauvm/CUDADataFormats/BeamSpotCUDA.cc, line 22:
cudaCheck(cudaMemAdvise(data_d_.get(), sizeof(Data), cudaMemAdviseUnsetReadMostly, device_));
cudaErrorDeviceUninitialized: invalid device context
```
(usually with `./cudauvm --validation`). It turned out that in these cases the destructor was run in a different thread than the constructor (i.e. the process was started with one thread, and TBB created a second thread because of`enqueue()` call), and the other thread did not have CUDA context initialized yet. By quick experimentation I guess the `cudaSetDevice()` would be the simplest CUDA runtime API function call that leads to CUDA context being initialized for a thread.

This PR also renames the `CUDAUVM_DISABLE_ADVICE` macro to `CUDAUVM_DISABLE_ADVISE` to be consistent with the CUDA runtime API function name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,16 @@ This program contains developments after CMSSW_11_1_0_pre4.
 
 The purpose of this program is to test the performance of the CUDA managed memory.
 
+To disable `cudaMemAdvise(cudaMemAdviseSetReadMostly)`, compile with
+```
+make cudauvm ... USER_CXXFLAGS="-DCUDAUVM_DISABLE_ADVISE"
+```
+
+To disable `cudaMemPrefetchAsync`, compile with
+```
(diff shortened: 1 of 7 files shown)


makortel:
I'm merging this now and leaving performance measurements for later time (would require larger refresh anyway)

makortel:
Part of #43

PR #62, opened 2020-06-25
Histo container

jtchilders:
I have ported the three unit tests that use `HistoContainer` and used those to port the class itself. I verified that the unit tests give the same outputs compared to the cuda versions.


src/kokkos/KokkosCore/HistoContainer.h
```diff
@@ -0,0 +1,339 @@
...
+    b += histOff(nh);
+    assert(b < totbins());
+    auto w = atomicDecrement(off[b]);
+    assert(w > 0);
+    bins[w - 1] = j;
+  }
+
+  #pragma hd_warning_disable
+  template<typename Histo>
+  static KOKKOS_INLINE_FUNCTION void finalize(Kokkos::View<Histo*,KokkosExecSpace> histo) {
```
makortel, line 305: Is this function supposed to be called only from host? In that case I belive the `KOKKOS_INLINE_FUNCTION` would better be dropped (I'd also consider making this a free function, but am not feeling strongly about it).

Do I understand/recall correctly that the difference wrt. CUDA code is that the (handcrafted)  `blockPrefixScan()` is designed to be called from device code, whereas Kokkos' `parallel_scan()` (in this case) needs to be invoked from host?

src/kokkos/KokkosCore/HistoContainer.h
```diff
@@ -0,0 +1,339 @@
...
+    // assert(off[totbins() - 1] == 0);
+    // for(uint32_t i=0;i<totbins();++i)
+    //   printf("1 %04i off[%04i] = %04i\n",teamMember.team_rank(),i,off[i]);
+    Kokkos::parallel_scan("finalize",Histo::totbins(),
+      KOKKOS_LAMBDA(const int& i, uint32_t& update, const bool& final){
```
makortel, line 310:
```suggestion
      KOKKOS_LAMBDA(const int i, uint32_t& update, const bool final){
```

src/kokkos/test/kokkos/HistoContainer_t.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "KokkosCore/HistoContainer.h"
+
+template <typename T>
+void go() {
+  std::mt19937 eng;
+  std::uniform_int_distribution<T> rgen(std::numeric_limits<T>::min(), std::numeric_limits<T>::max());
+
+  constexpr int N = 12000;
+  Kokkos::View<T*,KokkosExecSpace> v_d("v_d",N);
+  typename Kokkos::View<T*,KokkosExecSpace>::HostMirror v_h("v_h",N);
```
makortel, line 18: The mirror View can be created also with
```suggestion
  auto v_h = Kokkos::create_mirror_view(v_d);
```

src/kokkos/test/kokkos/HistoContainer_t.cc
```diff
@@ -0,0 +1,153 @@
...
+  constexpr uint32_t nParts = 10;
+  constexpr uint32_t partSize = N / nParts;
+//  uint32_t offsets[nParts + 1];
+
+  using Hist = HistoContainer<T, 128, N, 8 * sizeof(T), uint32_t, nParts>;
+  std::cout << "HistoContainer " << (int)(offsetof(Hist, off)) << ' ' << Hist::nbins() << ' ' << Hist::totbins() << ' '
+            << Hist::capacity() << ' ' << Hist::wsSize() << ' '
+            << (std::numeric_limits<T>::max() - std::numeric_limits<T>::min()) / Hist::nbins() << std::endl;
+
+  Kokkos::View<Hist*,KokkosExecSpace> h_d("h_d",1);
```
makortel, line 29: I would do
```suggestion
  Kokkos::View<Hist,KokkosExecSpace> h_d("h_d");
```

src/kokkos/test/kokkos/HistoContainer_t.cc
```diff
@@ -0,0 +1,153 @@
...
+      off_h(3) = 32 + off_h(2);
+      off_h(4) = 123 + off_h(3);
+      off_h(5) = 256 + off_h(4);
+      off_h(6) = 311 + off_h(5);
+      off_h(7) = 2111 + off_h(6);
+      off_h(8) = 256 * 11 + off_h(7);
+      off_h(9) = 44 + off_h(8);
+      off_h(10) = 3297 + off_h(9);
+    }
+    Kokkos::deep_copy(off_d,off_h);
```
makortel, line 56: Please pass the execution space object, e.g.
```suggestion
    Kokkos::deep_copy(KokkosExecSpace(), off_d,off_h);
```
so that the serial build of the unit test can be run on a machine without GPU.

src/kokkos/test/kokkos/OneHistoContainer_t.cc
```diff
@@ -0,0 +1,143 @@
...
+        v_h(j) = 4;
+
+    Kokkos::deep_copy(v_d,v_h);
+    
+    using TeamHist = HistoContainer<T, NBINS, N, S, uint16_t>;
+
+    Kokkos::View<TeamHist*,KokkosExecSpace> histo_d("histo_d",1);
+    auto histo_h = Kokkos::create_mirror_view(histo_d);
+
+    Kokkos::parallel_for("set_zero",Kokkos::RangePolicy<KokkosExecSpace>(0,TeamHist::totbins()),
```
makortel, line 50: Please pass the execution space object, e.g.
```suggestion
    Kokkos::parallel_for("set_zero",Kokkos::RangePolicy<KokkosExecSpace>(KokkosExecSpace(), 0,TeamHist::totbins()),
```
so that the serial build of the unit test can be ran on a machine without a GPU.

jtchilders:
So should I put all that code in `KOKKOS_NAMESPACE` or `cms::kokkos`? Preference?

src/kokkos/KokkosCore/HistoContainer.h
```diff
@@ -0,0 +1,339 @@
...
+      uint32_t first = teamMember.league_rank() * teamMember.team_size() + teamMember.team_rank();
+      uint32_t total_threads = teamMember.league_size() * teamMember.team_size();
+      for (uint32_t i = first, nt = offsets(nh); i < nt; i += total_threads) {
+        uint32_t index = 0;
+        for(uint32_t j = 0;j<=nh;++j){
+          if(offsets(j) > i){
+             index = j;
+             break;
+          }
+        }
```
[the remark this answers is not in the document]
jtchilders: I don't think `cuda_std::upper_bound()` would work directly since we have a Kokkos::View object instead of a real pointer to an array. Unless there is a way to get the actual pointer to the array in memory, perhaps there is an I'm not familiar with it yet.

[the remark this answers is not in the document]
makortel: `offsets.data()` gives a pointer to the underlying array (even though its use is discouraged IIRC).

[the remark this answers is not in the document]
jtchilders: I'll just write an `upper_bound` that operates on the view if that is OK. That will clarify the code and keep the kokkos methodology intact.

makortel:
> So should I put all that code in `KOKKOS_NAMESPACE` or `cms::kokkos`? Preference?

I don't have a strong opinion at the moment. `KOKKOS_NAMESPACE` might lead to fewer issues with "surprising template argument deduction" like to what @PointKernel encountered in https://github.com/makortel/pixel-standalone/pull/43.

src/kokkos/KokkosCore/HistoContainer.h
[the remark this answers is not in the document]
makortel: An `upper_bound()` with `Kokkos::View` would be fine as well, thanks.

src/kokkos/KokkosCore/HistoContainer.h
```diff
@@ -0,0 +1,339 @@
...
+
+#include "KokkosCore/AtomicPairCounter.h"
+
+using team_policy = Kokkos::TeamPolicy<KokkosExecSpace>;
+using member_type = Kokkos::TeamPolicy<KokkosExecSpace>::member_type;
+
+namespace cms {
+  namespace kokkos {
+    template <typename Histo, typename T>
+    KOKKOS_FUNCTION void countFromVector(Kokkos::View<Histo*,KokkosExecSpace> h,
```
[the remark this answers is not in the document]
jtchilders: I have only ever seen Kokkos views with arrays. How does it work with class objects or singletons? 
Can I do this?
``` python
Kokkos::View<int> x('x');
x = 5; 
```
This doesn't seem to compile.

makortel:
By the way, should this PR also change the `-I` to `-isystem` in 
https://github.com/cms-patatrack/pixeltrack-standalone/blob/274f5c72f98576895e8111eca96ba174fb6d497f/Makefile#L122
? (I did not try to build this PR yet)

src/kokkos/KokkosCore/HistoContainer.h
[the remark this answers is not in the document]
jtchilders: Just for documentation, the proper use is:
```C++
Kokkos::View<int> x("x");
x() = 5;
```

src/kokkos/KokkosCore/HistoContainer.h
```diff
@@ -0,0 +1,339 @@
...
+        int32_t ih = &offsets(index) - &offsets(0) - 1;
+        assert(ih >= 0);
+        assert(ih < int(nh));
+        h(0).fill(v(i), i, ih);
+      }
+    }
+
+    template <typename Histo>
+    inline void launchZero(Kokkos::View<Histo*,KokkosExecSpace> h) {
+      Kokkos::parallel_for(Histo::totbins(),KOKKOS_LAMBDA(const size_t i) {
```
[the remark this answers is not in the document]
jtchilders: I think this would be
```c++
Kokkos::parallel_for(Kokkos::RangePolicy<execSpace>(0,Histo::totbins()), KOKKOS_LAMBDA(const size_t& i) {
```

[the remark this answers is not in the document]
makortel: It can't be `Kokkos::RangePolicy<execSpace>(...)` because `execSpace` is a variable and a type. The point is to pass the `execSpace` object to the `RangePolicy` constructor.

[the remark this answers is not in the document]
jtchilders: Sorry I meant this:
```python
template <typename Histo, typename ExecSpace>
    inline void launchZero(Kokkos::View<Histo*,ExecSpace> h, ExecSpace const& execSpace) {
      Kokkos::parallel_for(Kokkos::RangePolicy<ExecSpace>(0,Histo::totbins()), 
        KOKKOS_LAMBDA(const size_t i) {
```
Doesn't this tell the range policy to execute in the same space as the histo view is located? And why are we passing the `execSpace` object to the `RangePolicy`?

If I follow your example which has `RangePolicy<KokkosExecSpace>(execSpace)` doesn't using the `KokkosExecSpace` negate the point?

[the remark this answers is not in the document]
makortel: When execution space object is passed to the execution policy constructor, the policy internally will call `fence()` only on that execution space instead of global `Kokkos::fence()`. It follows that passing execution space object always is the only way to allow running a serial-only job of a serial+CUDA build on a machine without a GPU (otherwise `Kokkos::fence()` calls `fence()` also on CUDA space, which fails without a GPU).

Furthermore, passing execution space object is necessary for asynchronous execution (to which we'll hopefully get some day), and crafting everything towards that now saves later effort.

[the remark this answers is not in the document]
jtchilders: OK, updated this to use `ExecSpace` in template and `execSpace` object passed to constructor.

src/kokkos/test/kokkos/HistoContainer_t.cc
jtchilders: per our discussion with Kokkos team, I've change all the single objects from pointers to objects.

src/kokkos/KokkosCore/HistoContainer.h
jtchilders: The Kokkos Documentation for `parallel_scan` lambdas uses pass by reference for all three variables in the function. Do you have a reason we should not follow that convention? I figured it would avoid a copy.

src/kokkos/test/kokkos/HistoContainer_t.cc
jtchilders: Added this in all the unit tests

jtchilders:
For the `-isystem` build issue, we can, but then I should remove my work around. Let me know.

src/kokkos/KokkosCore/HistoContainer.h
makortel: Just that usually taking primitive types by value is faster than going through the pointer (which reference internally is). E.g. C++ core guidelines recommend that https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in.

makortel:
> For the `-isystem` build issue, we can, but then I should remove my work around. Let me know.

Somehow I missed you had a workaround. Could you point me to it? It's likely fine, I'm mostly curious.

src/kokkos/test/kokkos/OneToManyAssoc_t.cc
```diff
@@ -0,0 +1,324 @@
...
+        }
+        assert(n_d() <= MaxElem);
+        assert(j <= N);
+      }
+    });
+
+  Kokkos::deep_copy(KokkosExecSpace(),n_h,n_d);
+  Kokkos::deep_copy(KokkosExecSpace(),nz_h,nz_d);
+  Kokkos::deep_copy(KokkosExecSpace(),ave_h,ave_d);
+  Kokkos::deep_copy(KokkosExecSpace(),imax_h,imax_d);
```
jtchilders, line 226: This is the work around. In the CUDA version these variables `n`, `nz`, `ave`, and `imax` were initialized serially on the host device, then copied to their device versions. Since I had the issue with `deep_copy` of a `int16**` View (that triggered your finding `-isystem`)  I decided to initialize the device side objects instead so I could avoid the `deep_copy`.

src/kokkos/test/kokkos/OneToManyAssoc_t.cc
```diff
@@ -0,0 +1,324 @@
...
+  std::default_random_engine generator (1234);
+  std::geometric_distribution<int> rdm(0.8);
+  constexpr uint32_t rdm_N = 16000;
+  Kokkos::View<int*,KokkosExecSpace> rdm_d("rdm_d",rdm_N);
+  auto rdm_h = Kokkos::create_mirror_view(rdm_d);
+  for(uint32_t i=0;i<rdm_N;++i){
+    rdm_h(i) = rdm(eng);
+    //printf("00 rdm(eng) = %04d\n",rdm_h(i));
+  }
+  Kokkos::deep_copy(rdm_d,rdm_h);
```
makortel, line 153:
```suggestion
  Kokkos::deep_copy(KokkosExecSpace(), rdm_d,rdm_h);
```

src/kokkos/test/kokkos/OneToManyAssoc_t.cc
makortel: Ok, thanks for the explanation.I think it's good as it is. Could you add a short comment here about that?

makortel:
I tried to run the tests, and the GPU versions work fine, but the CPU serial ones have issues
* `HistoContainer_t.serial` dies with `Kokkos::abort: Requested Team Size is too large!`
  * This happens because serial backend requires the team size to be 1, right?
* `OneHistoContainer_t.serial` dies with `Kokkos::Cuda::Cuda instance constructor : ERROR device not initialized`
  * This is a symptom of execution space object not being passed to some Kokkos operation (by quick glance I did not spot it though)
* `OneToManyAssoc_t.serial` segfaults in `countMultiLocal()`
  * I did not take a closer look

In principle I could accept these failures for now to enable further development depending on `HistoContainer` while these are being investigated/addressed.

jtchilders:
I found the problem with `OneHistoContainer.serial` so that works now.

Not sure how to dynamically set the team size to 1 for serial backend. Do you?

makortel:
Thanks, much better now. The `HistoContainer_t` and `OneHistoContainer_t` work now. `OneToManyAssoc_t.serial` works some times, and fails in assert in other times `OneToManyAssoc_t.cc:301: int main(): Assertion 'x == 4 || x == 3' failed.`.

makortel:
@jtchilders I debugged a bit the remaining failure in `OneToManyAssoc_t` and it seems to me that the `dc_d` here
https://github.com/cms-patatrack/pixeltrack-standalone/blob/b09a6612f8b0b5f4e5ad9bde2e57c70941472022/src/kokkos/test/kokkos/OneToManyAssoc_t.cc#L271
is uninitialized. It gets initialized a few lines below
https://github.com/cms-patatrack/pixeltrack-standalone/blob/b09a6612f8b0b5f4e5ad9bde2e57c70941472022/src/kokkos/test/kokkos/OneToManyAssoc_t.cc#L279-L281

Comparing to the CUDA version of the test
https://github.com/cms-patatrack/pixeltrack-standalone/blob/b09a6612f8b0b5f4e5ad9bde2e57c70941472022/src/cuda/test/OneToManyAssoc_t.h#L227-L240
it seems that the initialization of `dc_d`/`dc_h` would be missing. I made a PR to your branch https://github.com/jtchilders/pixeltrack-standalone/pull/1 with a fix suggestion. At least with that I was able run `OneToManyAssoc_t.serial` many times without crashes. What do you think?

jtchilders:
OK, merged the PR

PR #69, opened 2020-07-27
add OneAPI test

fwyzard:
First attempt of introducing a `oneapitest` directory and binary.

Current status:
  - builds with Intel DPC++ beta 07;
  - runs on the HOST device;
  - crashes on the CPU and GPU OpenCL devices.


Makefile
```diff
@@ -3,9 +3,9 @@ export BASE_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
...
+HOST_CXXFLAGS := -O2 -fPIC -fdiagnostics-show-option -felide-constructors -fmessage-length=0 -fno-math-errno -ftree-vectorize -fvisibility-inlines-hidden -msse3 -pipe -pthread -Xassembler --compress-debug-sections -Werror=address -Wall -Werror=array-bounds -Wno-attributes -Werror=conversion-null -Werror=delete-non-virtual-dtor -Wno-deprecated -Werror=format -Wno-long-long -Werror=main -Werror=missing-braces -Werror=narrowing -Wnon-virtual-dtor -Werror=overflow -Werror=overlength-strings -Wparentheses -Werror=pointer-arith -Werror=reorder -Wreturn-type -Werror=return-type -Werror=sign-compare -Werror=strict-aliasing -Wstrict-overflow -Werror=switch -Werror=type-limits -Wunused -Wno-unused-local-typedefs -Wno-unused-function -Werror=unused-value -Wno-error=unused-variable -Wno-vla -Werror=write-strings
```
makortel, line 6: Are these changed because `dpcpp` takes different flags than `g++`? How about defining `ONEAPI_CXXFLAGS` that filters out offending flags or just overwrites them, and would then be used in `src/oneapitest/Makefile` instead of `CXXFLAGS`?

src/oneapitest/Makefile
```diff
@@ -0,0 +1,117 @@
+TARGET_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+TARGET_NAME := $(notdir $(TARGET_DIR))
+TARGET := $(BASE_DIR)/$(TARGET_NAME)
+include Makefile.deps
+EXTERNAL_DEPENDS := $(oneapitest_EXTERNAL_DEPENDS)
+
+$(TARGET):
+test_cpu:
+test_sycl: $(TARGET)
```
makortel, line 9: Just to clarify, the intention is that the suffix corresponds a platform to be tested, e.g. `cpu` for all CPU backends (like `serial` or `tbb` with Kokkos or Alpaka), `cuda` for NVIDIA GPUs etc.

src/oneapitest/Makefile
```diff
@@ -0,0 +1,117 @@
...
+-include $(ALL_DEPENDS)
+
+# Build targets
+$(TARGET): $(EXE_OBJ) $(LIBS) $(PLUGINS) | $(TESTS_EXE)
+	dpcpp $(EXE_OBJ) $(LDFLAGS) $(MY_LDFLAGS) -o $@ -L$(LIB_DIR)/$(TARGET_NAME) $(patsubst %,-l%,$(LIBNAMES)) $(foreach dep,$(EXTERNAL_DEPENDS),$($(dep)_LDFLAGS))
+
+define BUILD_template
+$(OBJ_DIR)/$(2)/%.cc.o: $(SRC_DIR)/$(2)/%.cc
+	@[ -d $$(@D) ] || mkdir -p $$(@D)
+	dpcpp $(CXXFLAGS) $(MY_CXXFLAGS) $$(foreach dep,$(EXTERNAL_DEPENDS),$$($$(dep)_CXXFLAGS)) -c $$< -o $$@ -MMD
```
makortel, line 81: Does this mean that we essentially need to build everything with `dpcpp`, or was this just an easy way to get started?

src/oneapitest/SYCLCore/ScopedContext.cc
```diff
@@ -0,0 +1,88 @@
...
+
+      if (dataStream != stream()) {
+        // Different streams, need to synchronize
+        if (not available) {
+          // Event not yet occurred, so need to add synchronization
+          // here. Sychronization is done by making the SYCL stream to
+          // wait for an event, so all subsequent work in the stream
+          // will run only after the event has "occurred" (i.e. data
+          // product became available).
+          dataEvent.wait();
```
makortel, line 45: This does a blocking wait, right?

src/oneapitest/Makefile
fwyzard: I don't know if it's a requirement, but every example I have seen uses only `dpcpp`.

src/oneapitest/SYCLCore/ScopedContext.cc
```diff
@@ -0,0 +1,88 @@
...
+  ////////////////////
+
+  ScopedContextProduce::~ScopedContextProduce() {
+    // Intentionally not checking the return value to avoid throwing
+    // exceptions. If this call would fail, we should get failures
+    // elsewhere as well.
+    /*
+    DPCT1012:15: Detected kernel execution time measurement pattern and generated an initial code for time measurements in SYCL. You can change the way time is measured depending on your goals.
+    */
+    auto event__get_ct1 = clock();
```
makortel, line 82: Do I interpret correctly that this code is sort of a placeholder?

In SYCL e.g. the `queue.submit()` returns directly the event, right? Ideally we would capture these events and keep the last one. A straightforward way would be to provide the `submit()` and `memcpy()` API in the ScopedContext, which would be towards passing the ScopedContext around instead of the queue. Anyway, something for the longer term.

src/oneapitest/Makefile
fwyzard: I didn't intentionally make any changes here - think I simply copied this from `cudatest` and did a global replace for `cuda` wih `sycl`...

What should the `Makefile` look like ?

Keep in mind that with `dpcpp` (or rather, with the open source `clang` fork) we would build a single binary, and choose the backend type (host, opencl, cuda, etc.) and device (cpu, gpu, etc.) at runtime,

src/oneapitest/SYCLCore/ScopedSetDevice.h
```diff
@@ -0,0 +1,29 @@
...
+#include <CL/sycl.hpp>
+#include <dpct/dpct.hpp>
+
+namespace cms {
+  namespace sycl {
+    class ScopedSetDevice {
+    public:
+      explicit ScopedSetDevice(int newDevice) {
+        prevDevice_ = dpct::dev_mgr::instance().current_device_id();
+        dpct::dev_mgr::instance().select_device(newDevice);
```
makortel, line 13: I see SYCL also has the notion of the "current device" (is it per thread?). Does something actually need it? Or can the device be always given explicitly (or implicitly via the queue) in the API calls?

fwyzard: Actually this comes from `dpct`, the Intel compatibility library for porting CUDA code (which I'll slowly try to migrate away from).

Given that all operations require a queue, and a queue knows which devices it runs on, I'd simply remove the "current device" from everywhere.

src/oneapitest/SYCLCore/ScopedContext.cc
fwyzard: > Do I interpret correctly that this code is sort of a placeholder?

Yes - any `DPCT` comment you see left around means I didn't actually go through that part of the code :-)

src/oneapitest/Makefile
makortel: The runtime choice has been the goal for `kokkos` and `alpaka` as well. Both are basically running the same binary,  but with command line arguments specifying which backends to use. See e.g. for alpaka
https://github.com/cms-patatrack/pixeltrack-standalone/blob/c2e2642c9ccd1209626aba72869df1e39bb6a9a0/src/alpakatest/Makefile#L8-L18

For unit tests the approach needs to differ from `kokkos` and `alpaka`. For them, a separate binary is built for each backend. I suppose we could use command line arguments there as well for the single binary of SYCL.

src/oneapitest/SYCLCore/ScopedSetDevice.h
makortel: > Given that all operations require a queue, and a queue knows which devices it runs on, I'd simply remove the "current device" from everywhere.

I would prefer that as well.

src/oneapitest/Makefile
fwyzard: If needed, we can also force the selection of a specific backend and device using a dedicated device selector and compilation flags (or tie the selector to command line arguments).

PR #76, opened 2020-08-28
[kokkos] Port CAHitNtupletGeneratorKernels::buildDoublets function

PointKernel:


src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
```diff
@@ -167,20 +167,20 @@ namespace KOKKOS_NAMESPACE {
...
-      cudaCheck(cudaGetLastError());
-    }
+
+#ifdef KOKKOS_BACKEND_CUDA
+    int teamSize = 128;
+    // at least one league
+    int leagueSize = (std::max(1U, nhits) + teamSize - 1) / teamSize;
+    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, teamSize};
+#else
+    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, 1, Kokkos::AUTO()};
```
makortel, line 177: I think it would be better to do have these the other way around, if `KOKKOS_BACKEND_SERIAL`, then use `1, Kokkos::AUTO()`, and else use many teams. Then the "many teams" would get applied automatically for all parallel backends.

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
```diff
@@ -191,37 +191,51 @@ namespace KOKKOS_NAMESPACE {
...
+    Kokkos::TeamPolicy<KokkosExecSpace,
+                       Kokkos::LaunchBounds<gpuPixelDoublets::getDoubletsFromHistoMaxBlockSize,
+                                            gpuPixelDoublets::getDoubletsFromHistoMinBlocksPerMP>>
+        tempPolicy{execSpace, leagueSize, teamSize * stride};
+#else
+    int stride = 1;
+    Kokkos::TeamPolicy<KokkosExecSpace,
+                       Kokkos::LaunchBounds<gpuPixelDoublets::getDoubletsFromHistoMaxBlockSize,
+                                            gpuPixelDoublets::getDoubletsFromHistoMinBlocksPerMP>>
+        tempPolicy{execSpace, 1, Kokkos::AUTO()};
```
makortel, line 216: Same here.

src/kokkos/plugin-PixelTriplets/kokkos/gpuPixelDoublets.h
```diff
@@ -3,8 +3,10 @@
 
 #include "gpuPixelDoubletsAlgos.h"
 
-#ifdef TODO
+#ifdef KOKKOS_BACKEND_CUDA
 #define CONSTANT_VAR __constant__
+#else
+#define CONSTANT_VAR
```
makortel, line 9: Did you try creating a functor class with the constants as member data? (I vaguely recall we discussed that option)

(I'm also fine if we try it later)

makortel:
The unit test `kokkos/OneToManyAssoc_t.cuda` fails now with
```
found with ave occupancy 3.9105 4
  what():  cudaFuncGetAttributes( &attr, cuda_parallel_launch_local_memory<DriverType>) error( cudaErrorInvalidDeviceFunction): invalid device function .../pixeltrack-standalone/external/kokkos/install/include/Cuda/Kokkos_Cuda_KernelLaunch.hpp:448
```
I need to investigate it more, so consider this as note to myself.

src/kokkos/KokkosDataFormats/TrackingRecHit2DKokkos.h
```diff
@@ -33,7 +33,7 @@ class TrackingRecHit2DKokkos {
   Kokkos::View<uint32_t const*, MemorySpace> hitsModuleStart() const { return m_hitsModuleStart; }
   Kokkos::View<uint32_t*, MemorySpace> hitsLayerStart() { return m_hitsLayerStart; }
 #ifdef TODO
-  auto phiBinner() { return m_hist; }
+  Kokkos::View<Hist*, MemorySpace> phiBinner() { return m_hist; }
 #endif
```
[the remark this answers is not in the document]
PointKernel: I think so, but they are not used in `buildDoublets` function. We can remove `#ifdef TODO` when it's needed

src/kokkos/KokkosDataFormats/TrackingRecHit2DKokkos.h
```diff
@@ -76,7 +76,7 @@ class TrackingRecHit2DKokkos {
 
 #ifdef TODO
   // needed as kernel params...
-  Hist* m_hist;
+  Kokkos::View<Hist*, MemorySpace> m_hist;
 #endif
```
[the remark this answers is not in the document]
PointKernel: the same as above

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
PointKernel: sure, this kernel can be easily reorganized in that way

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
PointKernel: Here to have the same flexibility is less straightforward. Cause current multi-team kernels are basically a copy-paste from CUDA implementations which sometimes use CUDA-specific algorithms. In this `getDoubletsFromHisto`, for example, it set the blockDim as 64 (or 128 I don't remember) and hardcoded `stride size = 4`. In a multi-threading use case on CPU, it's possible that we have less than 4 hardware threads or we launch an odd number of threads. Also, I need to investigate if the code is flexible enough to use a different stride size other than 4. Just FYI, all my previous porting had the same issue also since I thought we only need to take care of CUDA and Serial.

src/kokkos/plugin-PixelTriplets/kokkos/gpuPixelDoublets.h
PointKernel: will do.

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
makortel: Thanks for the explanation. Let's keep it like this then, but could you add the explanation in a comment in the file itself (once in the first occurrence is enough)?

src/kokkos/KokkosCore/HistoContainer.h
```diff
@@ -8,7 +8,8 @@
 
 #include <Kokkos_Core.hpp>
 
-#include "KokkosCore/AtomicPairCounter.h"
+#include "AtomicPairCounter.h"
+#include "kokkos_assert.h"
```
makortel, line 12: The `kokkos/OneToManyAssoc_t.cuda` comes from `kokkos_assert.h`. Or really the logic inside `kokkos_assert.h`, the standard `<cassert>` itself works. I'm exploring some ways how deal with that (at the system level)

src/kokkos/KokkosDataFormats/TrackingRecHit2DKokkos.h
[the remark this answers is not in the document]
makortel: Ok.

src/kokkos/KokkosCore/HistoContainer.h
makortel: #77 should provide a proper fix.

makortel: Yup, that did it, the unit tests work now with this PR.

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
PointKernel: Actually a question here: how can we properly set `N` in `TeamPolicy(N, AUTO)` for different backends? I thought Kokkos may provide an auto-tuning abstraction for the whole team policy (i.e.,  `TeamPolicy(AUTO)`) but looks like `N` has to be set depending on each specific backend. In that case, a generic parallel solution seems impossible and we still need `#ifdef ...` to specify `N` manually.

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
PointKernel: done

src/kokkos/plugin-PixelTriplets/PixelDoubletsConstants.h
```diff
@@ -0,0 +1,86 @@
...
+        4, 5, 7, 8,                    // FPIX1 (8)
+        2, 3, 2, 4, 2, 7, 5, 6, 8, 9,  // BPIX3 & FPIX2 (13)
+        0, 2, 1, 3,                    // Jumping Barrel (15)
+        0, 5, 0, 8,                    // Jumping Forward (BPIX1,FPIX2)
+        4, 6, 7, 9                     // Jumping Forward (19)
+    };
+
+  public:
+    KOKKOS_INLINE_FUNCTION uint8_t operator[](int i) const { return layerPairs[i]; };
+  };
```
makortel, line 25: The idea was actually to define the constants as members of the functor itself. I.e. implement `getDoubletsFromHisto` itself as a class with the function call operator called directly by the `parallel_for`.

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
makortel: I think this is a good question for the notes :) You're right (in the other comment) that for immediate purposes Serial and CUDA backends are sufficient, but we should also keep in mind the other backends. For practical purposes this means just taking `N` as the number of blocks in CUDA for others-than-serial, and `1` for serial.

I'm probably biased from the flexibility in TBB and CUDA for not even thinking why `N` teams would not work more than `N` worker threads (in CPU case).

src/kokkos/plugin-PixelTriplets/PixelDoubletsConstants.h
PointKernel: yeah, I found it's weird also. Is the latest update correct?

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
PointKernel: Thanks, I will discuss this in the notes.

src/kokkos/plugin-PixelTriplets/PixelDoubletsConstants.h
makortel: Thanks, it looks like what I was after.

PointKernel: Thanks

PR #81, opened 2020-09-04
added a kokkos version of TrackingRecHit2DCuda test

alexstrel:
experimental Kokkos version of RecHit module

### DIFF
--- a/src/kokkos/KokkosDataFormats/TrackingRecHit2DKokkos.h
+++ b/src/kokkos/KokkosDataFormats/TrackingRecHit2DKokkos.h
@@ -32,6 +32,8 @@ public:
 
   Kokkos::View<uint32_t const*, MemorySpace> hitsModuleStart() const { return m_hitsModuleStart; }
   Kokkos::View<uint32_t*, MemorySpace> hitsLayerStart() { return m_hitsLayerStart; }
+
+  Kokkos::View<TrackingRecHit2DSOAView, MemorySpace> mView() { return m_view; }
 #ifdef TODO
   Kokkos::View<Hist*, MemorySpace> phiBinner() { return m_hist; }
 #endif
--- a/src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
+++ b/src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
@@ -10,10 +10,10 @@ namespace testTrackingRecHit2DKokkos {
   void fill(Kokkos::View<TrackingRecHit2DSOAView, MemorySpace> hits) {
 
     assert(hits.data());
-    auto &hits_ = *hits.data();
+    auto hits_ = &hits();
 
     Kokkos::parallel_for("fill", Kokkos::RangePolicy<KokkosExecSpace>(KokkosExecSpace(), 0, 1024), KOKKOS_LAMBDA(const size_t i) {
-              assert(hits_.nHits() == 200);
+              assert(hits_->nHits() == 200);
               if (i > 200) return;
             });
 
@@ -25,10 +25,10 @@ namespace testTrackingRecHit2DKokkos {
 
     assert(hits.data());
 
-    auto const &hits_ = *hits.data();
+    auto const hits_ = &hits();
 
     Kokkos::parallel_for("fill", Kokkos::RangePolicy<KokkosExecSpace>(0, 1024), KOKKOS_LAMBDA(const size_t i) {
-              assert(hits_.nHits() == 200);
+              assert(hits_->nHits() == 200);
               if (i > 200) return;
             });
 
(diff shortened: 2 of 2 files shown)


src/kokkos/KokkosDataFormats/TrackingRecHit2DKokkos.h
```diff
@@ -32,6 +32,8 @@ class TrackingRecHit2DKokkos {
 
   Kokkos::View<uint32_t const*, MemorySpace> hitsModuleStart() const { return m_hitsModuleStart; }
   Kokkos::View<uint32_t*, MemorySpace> hitsLayerStart() { return m_hitsLayerStart; }
+
+  Kokkos::View<TrackingRecHit2DSOAView, MemorySpace> mView() { return m_view; }
```
makortel, line 36: I would have justs used the pointer from `view()`, but this is ok as well.

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "CondFormats/pixelCPEforGPU.h"
+#include "KokkosDataFormats/TrackingRecHit2DKokkos.h"
+
+namespace testTrackingRecHit2DKokkos {
+
+  template<typename MemorySpace>
+  void fill(Kokkos::View<TrackingRecHit2DSOAView, MemorySpace> hits) {
+
+    assert(hits.data());
+    auto &hits_ = *hits.data();
```
makortel, line 13: The preferred way in Kokkos would be
```suggestion
    auto &hits_ = hits();
```
(I'm personally fine with either way)

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
```diff
@@ -0,0 +1,70 @@
...
+
+    Kokkos::parallel_for("fill", Kokkos::RangePolicy<KokkosExecSpace>(0, 1024), KOKKOS_LAMBDA(const size_t i) {
```
makortel, line 15: This is ok here, but for functions called in the "production" code the policies should take also the execution space object (that would be one of the function's parameters).

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
```diff
@@ -0,0 +1,70 @@
...
+    assert(hits.data());
+
+    fill(hits);
+    verify(hits);    
+  }
+}  // namespace testTrackingRecHit2DKokkos
+
+namespace testTrackingRecHit2DKokkos {
+  template<typename MemorySpace>  
+  void runKernels(Kokkos::View<TrackingRecHit2DSOAView, MemorySpace> hits);
```
makortel, line 49: Just to note that this declaration looks unnecessary, but it is in the CUDA version as well (who knows why), so it can stay here as well.

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
```diff
@@ -0,0 +1,70 @@
...
+
+int main() {	
+
+  kokkos_common::InitializeScopeGuard kokkosGuard({KokkosBackend<KokkosExecSpace>::value});
+
+  {
+    auto nHits = 200;
+
+    Kokkos::View<pixelCPEforGPU::ParamsOnGPU, KokkosExecSpace> _cpeParams("cpeparams");
+    Kokkos::View<uint32_t*, KokkosExecSpace> _hitsModuleStart("hitsmodulestart");
```
makortel, line 61: Both serial and CUDA versions of the test fail with
```
Constructor for Kokkos View 'hitsmodulestart' has mismatched number of arguments. Number of arguments = 0 but dynamic rank = 1 
Aborted
```
The constructor of `_hitsModuleStart` would either need one additional parameter for the size, or be changed to call the default constructor. The default constructor (also for `_cpeParams`) would be closer to the CUDA test which passes `nullptr` for both.

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
makortel: Actually this might require the execution space object being passed in along
```suggestion
    Kokkos::parallel_for("fill", Kokkos::RangePolicy<KokkosExecSpace>(KokkosExecSpace(), 0, 1024), KOKKOS_LAMBDA(const size_t i) {
```
in order to be able to run the serial backend binaries (because then the possible internal `fence()` calls will be limited to the `KokkosExecSpace` instead of full `Kokkos::fence()` that would try to synchronize also the CUDA backend, that is not initialized for the serial backend binaries).

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
alexstrel: I wondered about this point. Agree that this should be updated synchronously.

src/kokkos/test/kokkos/TrackingRecHit2D_t.cc
makortel: @alexstrel Thanks for the update. Could you address this comment as well? I'm not able to test without.

alexstrel: @makortel, sure. I was able to build without issues in my environment, thought.

makortel: Right, it builds but neither serial or CUDA binaries of the test run.

PR #86, opened 2020-09-11
[kokkos] code format

makortel:
Somehow slipped through in #80 (I thought I checked the formatting).


makortel:
@jtchilders Could you check that this PR does not interfere with your current work?

jtchilders:
Thanks @makortel 
It will require me to do some merging for sure. I think I'm almost done with SiPixelRawToClusterGPUKernel. Could you do this after that?

makortel:
Thanks, then I'll wait for you to finish.

PR #90, opened 2020-09-14
[kokkos] Enable RecHit module

makortel:
~~This PR is work in progress to enable RecHit module.~~

This PR enables the RecHit module to be run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ downloaded automatically during the build process.
 | `cudadev`    | CUDA version (development)       | :heavy_check_mark: | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `cudauvm`    | CUDA version with managed memory | :heavy_check_mark: | :heavy_check_mark: |                    | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | `kokkostest` | Kokkos FW test                   | :heavy_check_mark: | :white_check_mark: | :heavy_check_mark: |                    |                    |                    |                    |                    |
-| `kokkos`     | Kokkos version                   | :heavy_check_mark: |                    | :heavy_check_mark: | :white_check_mark: |                    |                    |                    |                    |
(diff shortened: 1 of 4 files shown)


makortel:
Currently this fails with
```
----------
Caught std::runtime_error
cudaFuncGetAttributes( &attr, cuda_parallel_launch_local_memory<DriverType>) error( cudaErrorInvalidDeviceFunction): invalid device function .../pixeltrack-standalone/external/kokkos/install/include/Cuda/Kokkos_Cuda_KernelLaunch.hpp:448       
```
"caused" by
https://github.com/cms-patatrack/pixeltrack-standalone/blob/790d95bb918a5538e3d64f2fb7899a44df57f444/src/kokkos/plugin-SiPixelRecHits/kokkos/PixelRecHits.cc#L60-L68
("caused" meaning that if I comment that call out, the program succeeds)

makortel:
Both
https://github.com/cms-patatrack/pixeltrack-standalone/blob/aa9b4f5673b24d11bf345ff7bb240eed36bbc48a/src/kokkos/plugin-SiPixelRecHits/kokkos/PixelRecHits.cc#L43-L54
and
https://github.com/cms-patatrack/pixeltrack-standalone/blob/aa9b4f5673b24d11bf345ff7bb240eed36bbc48a/src/kokkos/plugin-SiPixelRecHits/kokkos/PixelRecHits.cc#L76-L77
seem to be ok. Both of them use `TeamPolicy`. Only the lambda with `RangePolicy` appears to be causing the problem.

makortel:
Trying again a single plugin (essentially reverting 57eb24233c49fd2e2bca90ebab1ca8d9433a1479 of #48) did not help. Going a step further and packaging all Kokkos code into a single plugin dynamic library did not help either.

makortel:
Trying CUDA 11.0.2 instead of CUDA 10.2 did not help.

makortel:
Artificially avoiding the load of `pluginSiPixelClusterizer.so` did not help, and also commenting out the team policy kernels mentioned in https://github.com/cms-patatrack/pixeltrack-standalone/pull/90#issuecomment-692271916 did not help.

I think all these indicate that this specific `parallel_for(RangePolicy)` is the one that causes the problem.

makortel:
Commenting out `assert()` in the body seems to fix the problem.

makortel:
I was not able to reproduce the problematic behavior with a unit test (using exactly the same `parallel_for` as in `PixelRecHits.cc`)

PR #96, opened 2020-09-21
[kokkos] Use unit-stride loops for the serial execution

PointKernel:
`stride=1` is required for serial execution since team size must be 1.

### DIFF
--- a/src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
+++ b/src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
@@ -51,8 +51,9 @@ namespace KOKKOS_NAMESPACE {
 #ifdef KOKKOS_BACKEND_CUDA
     Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, teamSize};
 #else  // serial
-    // TODO: does the serial backend give correct results (with non-unit-stride loops)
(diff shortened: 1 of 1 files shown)


src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
```diff
@@ -51,8 +51,9 @@ namespace KOKKOS_NAMESPACE {
 #ifdef KOKKOS_BACKEND_CUDA
     Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, teamSize};
 #else  // serial
-    // TODO: does the serial backend give correct results (with non-unit-stride loops)
-    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, 1, Kokkos::AUTO()};
+    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, 1};
```
makortel, line 54: Would it now make sense to change the `#ifdefs` to
```cpp
#ifdef KOKKOS_BACKEND_SERIAL
    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, 1};
    // unit stride loop for serial execution
    stride = 1;
#else
    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, teamSize};
#endif  
```
?

PointKernel: make sense.

PR #102, opened 2020-10-01
[sycltest] Implement the framework test for the SYCL/oneAPI backend

fwyzard:
Implementation the `sycltest` test of the host and device framework for the SYCL/oneAPI backend,
based on the `cudatest` program, summarily translated by DPCT, fixed and reworked.

Known limitations:
  - caching of device and host memory, queues (streams) and events has been removed
  - there is no explicit support for device selection, so left to be driven by environment variables
  - "make test" needs to be updated


fwyzard:
@makortel here is first draft of the `sycltest` program.

We have more work to do, but this version already compiles and runs on the "host" device and on the Intel gen9 integrated GPUs OpenCL backend.

And, the performance is likely horrible for the moment...

fwyzard:
@lauracappelli @Viola27 FYI

src/sycltest/SYCLCore/chooseDevice.cc
```diff
@@ -0,0 +1,32 @@
...
+  sycl::device chooseDevice(edm::StreamID id) {
+    auto const& devices = enumerateDevices();
+
+    // For startes we "statically" assign the device based on
+    // edm::Stream number. This is suboptimal if the number of
+    // edm::Streams is not a multiple of the number of CUDA devices
+    // (and even then there is no load balancing).
+    //
+    // TODO: improve the "assignment" logic
+    auto const& device = devices[id % devices.size()];
```
makortel, line 28: Do I understand correctly that on a machine with CPU and GPU, this will alternate between them?

src/sycltest/SYCLCore/device_unique_ptr.h
```diff
@@ -0,0 +1,96 @@
...
+                                                                              device::impl::DeviceDeleter{stream}};
+    }
+
+    template <typename T>
+    typename device::impl::make_device_unique_selector<T>::unbounded_array make_device_unique(size_t n,
+                                                                                              sycl::queue stream) {
+      using element_type = typename std::remove_extent<T>::type;
+      static_assert(std::is_trivially_constructible<element_type>::value,
+                    "Allocating with non-trivial constructor on the device memory is not supported");
+      void *mem = sycl::malloc_device(n * sizeof(element_type), stream);
```
makortel, line 65: Do you know (or can you point to documentation) how the `malloc_device()` (and `malloc_host()`) use the `sycl::queue`?

src/sycltest/plugin-Test1/gpuAlgo1.cc
```diff
@@ -0,0 +1,175 @@
+#include <algorithm>
+#include <cmath>
+
+#include <CL/sycl.hpp>
+
+#include "SYCLCore/device_unique_ptr.h"
+#include "SYCLCore/host_unique_ptr.h"
+#include "gpuAlgo1.h"
+
+namespace gpu_algo_1 {
```
makortel, line 10: Does anonymous namespace not work?

src/sycltest/plugin-Test1/gpuAlgo1.cc
```diff
@@ -0,0 +1,175 @@
...
+
+  auto d_c = cms::sycltools::make_device_unique<float[]>(NUM_VALUES, stream);
+  stream.submit([&](sycl::handler &cgh) {
+    auto d_a_get_ct0 = d_a.get();
+    auto d_b_get_ct1 = d_b.get();
+    auto d_c_get_ct2 = d_c.get();
+
+    cgh.parallel_for(sycl::nd_range<3>(sycl::range<3>(1, 1, blocksPerGrid) * sycl::range<3>(1, 1, threadsPerBlock),
+                                       sycl::range<3>(1, 1, threadsPerBlock)),
+                     [=](sycl::nd_item<3> item_ct1) {
```
makortel, line 106: Could one do
```suggestion
    cgh.parallel_for(sycl::nd_range<1>(sycl::range<1>(blocksPerGrid) * sycl::range<1>(threadsPerBlock),
                                       sycl::range<3>(threadsPerBlock)),
                     [=](sycl::nd_item<1> item_ct1) {
```
?

src/sycltest/plugin-Test1/gpuAlgo1.cc
```diff
@@ -0,0 +1,175 @@
...
+#include "SYCLCore/device_unique_ptr.h"
+#include "SYCLCore/host_unique_ptr.h"
+#include "gpuAlgo1.h"
+
+namespace gpu_algo_1 {
+  constexpr int NUM_VALUES = 1000;
+
+  template <typename T>
+  SYCL_EXTERNAL void vectorAdd(const T *a, const T *b, T *c, int numElements, sycl::nd_item<3> item_ct1) {
+    int i = item_ct1.get_group(2) * item_ct1.get_local_range().get(2) + item_ct1.get_local_id(2);
```
makortel, line 15: If I understand correctly, the index `2` refers to the third element of `item_ct1`. I find using that (instead of `0` and first element) a bit counter-intuitive (but ok). Is that perhaps the closest correspondence with CUDA's x-dimension?

src/sycltest/plugin-Test1/gpuAlgo1.cc
```diff
@@ -0,0 +1,175 @@
...
+    auto threadsPerGrid = blocksPerGrid3 * threadsPerBlock3;
+
+    auto d_a_get_ct0 = d_a.get();
+    auto d_b_get_ct1 = d_b.get();
+    auto d_ma_get_ct2 = d_ma.get();
+
+    cgh.parallel_for(
+        sycl::nd_range<3>(sycl::range<3>(threadsPerGrid.get(2), threadsPerGrid.get(1), threadsPerGrid.get(0)),
+                          sycl::range<3>(threadsPerBlock3.get(2), threadsPerBlock3.get(1), threadsPerBlock3.get(0))),
+        [=](sycl::nd_item<3> item_ct1) {
```
makortel, line 127: Ok, now I'm just confused of how the `sycl::range`, `sycl::nd_range`, and `sycl::nd_item` work exactly.

makortel (commented):
I think this is an excellent starting point!

Makefile
```diff
@@ -3,9 +3,9 @@ export BASE_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
...
+HOST_CXXFLAGS := -O2 -g -fPIC -fdiagnostics-show-option -felide-constructors -fmessage-length=0 -fno-math-errno -ftree-vectorize -fvisibility-inlines-hidden -msse3 -pipe -pthread -Werror=address -Wall -Werror=array-bounds -Wno-attributes -Werror=conversion-null -Werror=delete-non-virtual-dtor -Wno-deprecated -Werror=format -Wno-long-long -Werror=main -Werror=missing-braces -Werror=narrowing -Wnon-virtual-dtor -Werror=overflow -Werror=overlength-strings -Wparentheses -Werror=pointer-arith -Wno-reorder-ctor -Wreturn-type -Werror=return-type -Wno-sign-compare -Werror=strict-aliasing -Wstrict-overflow -Werror=switch -Werror=type-limits -Wunused -Wno-unused-local-typedefs -Wno-unused-function -Werror=unused-value -Wno-error=unused-variable -Wno-vla -Werror=write-strings
```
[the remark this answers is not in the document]
fwyzard: Well, I added `-g` to help debug things :-)
Do you prefer that we add it by hand / from the command line when needed ?

I removed `-Xassembler --compress-debug-sections` because compressed debug sections do not seem to be supported by `dpcpp`:
```
/opt/intel/oneapi/compiler/latest/linux/bin/dpcpp -fsycl -I/opt/intel/oneapi/dpcpp-ct/latest/include /data/user/fwyzard/pixeltrack-standalone/obj/sycltest/DataFormats/FEDRawDataCollection.cc.o /data/user/fwyzard/pixeltrack-standalone/obj/sycltest/DataFormats/FEDHeader.cc.o /data/user/fwyzard/pixeltrack-standalone/obj/sycltest/DataFormats/FEDNumbering.cc.o /data/user/fwyzard/pixeltrack-standalone/obj/sycltest/DataFormats/FEDRawData.cc.o /data/user/fwyzard/pixeltrack-standalone/obj/sycltest/DataFormats/FEDTrailer.cc.o -O2 -g -fPIC -pthread -Wl,-E -lstdc++fs -shared -Wl,-z,defs -L/data/user/fwyzard/pixeltrack-standalone/lib/sycltest  -L/data/user/fwyzard/pixeltrack-standalone/external/tbb/lib -ltbb  -o /data/user/fwyzard/pixeltrack-standalone/lib/sycltest/libDataFormats.so
/usr/bin/ld: /tmp/fwyzard/FEDRawDataCollection-d99872.o: unable to initialize decompress status for section .debug_loc
/usr/bin/ld: /tmp/fwyzard/FEDRawDataCollection-d99872.o: unable to initialize decompress status for section .debug_loc
/tmp/fwyzard/FEDRawDataCollection-d99872.o: file not recognized: File format not recognized
dpcpp: error: linker command failed with exit code 1 (use -v to see invocation)
```

By the way, compressed debug sections are also not supported by `cuda-gdb`, so I would really be in favour of removing this option.


Other options are not understood:
```
dpcpp: warning: argument unused during compilation: '--param vect-max-version-for-alias-checks=50' [-Wunused-command-line-argument]
warning: unknown warning option '-Werror=format-contains-nul' [-Wunknown-warning-option]
warning: unknown warning option '-Wno-non-template-friend'; did you mean '-Wno-unsupported-friend'? [-Wunknown-warning-option]
warning: unknown warning option '-Werror=return-local-addr'; did you mean '-Werror=return-stack-address'? [-Wunknown-warning-option]
warning: unknown warning option '-Werror=unused-but-set-variable'; did you mean '-Werror=unused-const-variable'? [-Wunknown-warning-option]
```
I think thse might be options we switch off for the clang builds in CMSSW as well ?

Some more were giving a lot of positives in the `dpct` headers.
Since I managed to get rid of them after the migration, they can be added back.

Makefile
```diff
@@ -36,7 +36,7 @@ export CUDA_TEST_CXXFLAGS := -DGPU_DEBUG
 export CUDA_LDFLAGS := -L$(CUDA_BASE)/lib64 -lcudart -lcudadevrt
 export CUDA_NVCC := $(CUDA_BASE)/bin/nvcc
 define CUFLAGS_template
-$(2)NVCC_FLAGS := $$(foreach ARCH,$(1),-gencode arch=compute_$$(ARCH),code=sm_$$(ARCH)) --expt-relaxed-constexpr --expt-extended-lambda --generate-line-info --source-in-ptx --cudart=shared
+$(2)NVCC_FLAGS := $$(foreach ARCH,$(1),-gencode arch=compute_$$(ARCH),code=sm_$$(ARCH)) -Wno-deprecated-gpu-targets -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored --expt-relaxed-constexpr --expt-extended-lambda --generate-line-info --source-in-ptx --cudart=shared
```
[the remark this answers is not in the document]
fwyzard: Agreed. I didn't want to touch those in this PR, but I can do that if you prefer.

Makefile
```diff
@@ -159,8 +188,12 @@ env.sh: Makefile
 	@echo -n '$(CUDA_LIBDIR):'                                              >> $@
 	@echo -n '$(CUPLA_LIBDIR):'                                             >> $@
 	@echo -n '$(KOKKOS_LIBDIR):'                                            >> $@
+	@echo -n '$(SYCL_LIBDIR):'                                              >> $@
 	@echo '$$LD_LIBRARY_PATH'                                               >> $@
-	@echo 'export PATH=$$PATH:$(CUDA_BASE)/bin'                             >> $@
+	@echo 'export PATH=$$PATH:$(CUDA_BASE)/bin:$(SYCL_BASE)/bin'            >> $@
```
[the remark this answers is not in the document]
fwyzard: Good point - yes.
Also for `$(SYCL_LIBDIR)` ?

src/sycltest/Framework/ESPluginFactory.h
```diff
@@ -0,0 +1,60 @@
...
+
+    std::unique_ptr<ESProducer> create(std::string const& name, std::filesystem::path const& datadir);
+  }  // namespace ESPluginFactory
+}  // namespace edm
+
+#define EDM_ES_PLUGIN_SYM(x, y) EDM_ES_PLUGIN_SYM2(x, y)
+#define EDM_ES_PLUGIN_SYM2(x, y) x##y
+
+#define DEFINE_FWK_EVENTSETUP_MODULE(type) \
+  [[maybe_unused]] static edm::ESPluginFactory::impl::Registrar<type> EDM_ES_PLUGIN_SYM(maker, __LINE__)(#type);
```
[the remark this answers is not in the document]
fwyzard: Yes, compiling with `dpcpp` gave warnings about all these being unused.
I did not investigate the reason why.

src/sycltest/Framework/PluginFactory.h
```diff
@@ -0,0 +1,58 @@
...
+    }  // namespace impl
+
+    std::unique_ptr<Worker> create(std::string const& name, ProductRegistry& reg);
+  }  // namespace PluginFactory
+}  // namespace edm
+
+#define EDM_PLUGIN_SYM(x, y) EDM_PLUGIN_SYM2(x, y)
+#define EDM_PLUGIN_SYM2(x, y) x##y
+
+#define DEFINE_FWK_MODULE(type) [[maybe_unused]] static edm::PluginFactory::impl::Registrar<type> EDM_PLUGIN_SYM(maker, __LINE__)(#type);
```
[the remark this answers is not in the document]
fwyzard: Yep, here as well.

src/sycltest/bin/main.cc
```diff
@@ -0,0 +1,150 @@
...
+  }
+  if (datadir.empty()) {
+    datadir = std::filesystem::path(args[0]).parent_path() / "data";
+  }
+  if (not std::filesystem::exists(datadir)) {
+    std::cout << "Data directory '" << datadir << "' does not exist" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  // Initialse the CUDA runtime
```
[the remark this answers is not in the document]
fwyzard: :+1:

src/sycltest/bin/main.cc
```diff
@@ -0,0 +1,150 @@
...
+  auto diff = stop - start;
+  auto time = static_cast<double>(std::chrono::duration_cast<std::chrono::microseconds>(diff).count()) / 1e6;
+  std::cout << "Processed " << maxEvents << " events in " << std::scientific << time << " seconds, throughput "
+            << std::defaultfloat << (maxEvents / time) << " events/s." << std::endl;
+  return EXIT_SUCCESS;
+}
+catch (sycl::exception const& exc) {
+  std::cerr << exc.what() << "Exception caught at file:" << __FILE__ << ", line:" << __LINE__ << std::endl;
+  std::exit(1);
+}
```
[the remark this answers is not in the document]
fwyzard: > Ok for now, but could be the try-catch scope be smaller?

This was added automatically by the migration - I don't know if because we call some SYCL functions from `main`, or a priori to make sure all SYCL exceptions are handled.

In CMSSW we can rely on the framework catching all exceptions, so we don't need to have such overarching `catch` block.

> I suppose in CMSSW the `cms::sycltools::enumerateDevices(true);` (and possible other initialization) would go into a Service.

Yes, same as for the call to `cudaGetDeviceCount(&numberOfDevices);` in `cudatest` *vs* using `CUDAService` in CMSSW.

src/sycltest/Makefile
```diff
@@ -0,0 +1,117 @@
+TARGET_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+TARGET_NAME := $(notdir $(TARGET_DIR))
+TARGET := $(BASE_DIR)/$(TARGET_NAME)
+include Makefile.deps
+EXTERNAL_DEPENDS := $(sycltest_EXTERNAL_DEPENDS)
+
+$(TARGET):
+test_cpu:
+test_sycl: $(TARGET)
```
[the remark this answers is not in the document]
fwyzard: I think I had the command line part in `pixel-standalone`, so eventually I can import it here.
For the moment I am using environment variables to test the various backends, e.g.
```
# use only the SYCL host "dvice"
SYCL_DEVICE_TYPE=HOST ./sycltest

# offload to the CPU using the OpenCL backend
SYCL_DEVICE_TYPE=CPU ./sycltest

# offload to the GPU using the OpenCL backend
SYCL_BE=PI_OPENCL SYCL_DEVICE_TYPE=GPU ./sycltest

# offload to the GPU using the Level 0 backend
SYCL_BE=PI_LEVEL0 SYCL_DEVICE_TYPE=GPU ./sycltest
```

Is there a way to add something like this to the tests, for the time being ?

src/sycltest/Makefile
```diff
@@ -0,0 +1,117 @@
...
+	@echo "Succeeded"
+test_$(2): run_$(1)
+endef
+$(foreach test,$(TESTS_EXE_CPU),$(eval $(call RUNTEST_template,$(test),cpu)))
+
+-include $(ALL_DEPENDS)
+
+# Build targets
+$(TARGET): $(EXE_OBJ) $(LIBS) $(PLUGINS) | $(TESTS_EXE)
+	$(SYCL_CXX) $(SYCL_FLAGS) $(EXE_OBJ) $(LDFLAGS) $(MY_LDFLAGS) -o $@ -L$(LIB_DIR)/$(TARGET_NAME) $(patsubst %,-l%,$(LIBNAMES)) $(foreach dep,$(EXTERNAL_DEPENDS),$($(dep)_LDFLAGS))
```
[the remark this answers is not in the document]
fwyzard: Sure, though I suspect that disentangling things is more complicated than with CUDA.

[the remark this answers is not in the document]
fwyzard: Do you want to open an issue :-) to remind us of this ?

src/sycltest/SYCLCore/chooseDevice.cc
fwyzard: Yes, currently this will cycle through all available devices: "host device", OpenCL cpu driver, OpenCL gpu driver, maybe also Level 0 gpu driver (i.e. the gpu is overcommited), FPGA emulator device...

The device discovery and selection clearly needs more work.

src/sycltest/SYCLCore/device_unique_ptr.h
fwyzard: Documentation is scarce...

According to the [SYCL 2020 specification](https://www.khronos.org/registry/SYCL/specs/sycl-2020-provisional.pdf#page=200), the `sycl::queue` is used just to get the `sycl::device` and `sycl::context` to be used for the memory allocation.

That is, there is no mention of any synchronisation with the host or device work in the queue.

src/sycltest/plugin-Test1/gpuAlgo1.cc
fwyzard: No :-(

```
/data/user/fwyzard/pixeltrack-standalone/src/sycltest/plugin-Test1/gpuAlgo1.cc:14:3: error: 'sycl_device' attribute cannot be applied to a static function or function in an anonymous namespace
  SYCL_EXTERNAL void vectorAdd(const T *a, const T *b, T *c, int numElements, sycl::nd_item<3> item_ct1) {
  ^
<built-in>:776:38: note: expanded from here
#define SYCL_EXTERNAL __attribute__((sycl_device))
                                     ^
```

src/sycltest/plugin-Test1/gpuAlgo1.cc
fwyzard: Unfortunately yes, I think the order of the indices in SYCL is reversed with respect that in CUDA.

In CUDA, `x` increments faster, followed by `y`, followed by `z`.
In SYCL, `<2>` supposedly increments faster, followed by `<1>`, followed by `<0>`.
I write "supposedly" because I haven't read much about it (I haven't searched much, either), I only remember stumbling on a Codeplay [blog entry](https://www.codeplay.com/portal/blogs/2019/11/18/computecpp-v1-1-6-changes-to-work-item-mapping-optimization.html) related to this.

Anyway, the mapping was done automatically by the conversion tool, so I assumed it to be correct :-)

src/sycltest/plugin-Test1/gpuAlgo1.cc
fwyzard: Welcome to the club :-(

As far as I understand:
  - `sycl::range` is a tuple, *e.g.* `64` (1-dimensional) or `4, 4, 4` (3-dimensional), and describes the "block" size (`blockDim`) OR the "grid" size (`gridDim`) OR the total size (`gridDim * blockDim`);
  - `sycl::nd_range` is two tuples, and describes the whole grid, like `<<<gridDim, blockDim>>>`, but it contains the *total* size and the *block* size; so `32 blocks x 32 threads` in CUDA are configred as `1024 global, 32 local`;
  - `sycl::nd_item` identifies one work item, like (`blockId, threadId`) in CUDA; it has accessors for the global, local and group ids, as well as global, local and group linearised indices.

fwyzard: I think the conversion tools did a poor job here, I'll see if I can clean things up a bit.

src/sycltest/plugin-Test1/gpuAlgo1.cc
fwyzard: Yes, but it should be simplified a lot more than that :-)

src/sycltest/plugin-Test1/gpuAlgo1.cc
fwyzard: I found some more details in the [SYCL documentation(https://www.khronos.org/registry/SYCL/specs/sycl-2020-provisional.pdf#equation.4.7.3):
image

fwyzard:
With the latest commit the test runs on all reasonable backends:
```
 SYCL_DEVICE_ALLOWLIST='DeviceName:{{.*CPU.*}}|DeviceName:{{.*Gen9.*}}' ./sycltest --maxEvents 10 --numberOfThreads 1 --numberOfStreams 3
Found 3 SYCL devices:
  - Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
  - Intel(R) Gen9 HD Graphics NEO
  - SYCL host device

Processing 10 events, of which 3 concurrently, with 1 threads.
EDM stream 0 offload to Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
TestProducer  Event 2 stream 2 ES int 42 FED 1200 size 2456
EDM stream 2 offload to SYCL host device
TestProducer2::acquire Event 2 stream 2 array 0x7fe8a805da80
TestProducer3 Event 2 stream 2
TestProducer  Event 3 stream 1 ES int 42 FED 1200 size 2040
EDM stream 1 offload to Intel(R) Gen9 HD Graphics NEO
TestProducer2::acquire Event 1 stream 0 array 0x17ed5180
TestProducer3 Event 1 stream 0
TestProducer2::produce Event 2 stream 2
TestProducer  Event 4 stream 2 ES int 42 FED 1200 size 2224
EDM stream 2 offload to SYCL host device
TestProducer2::acquire Event 3 stream 1 array 0x7fe8a84b6000
TestProducer3 Event 3 stream 1
TestProducer2::produce Event 1 stream 0
TestProducer2::acquire Event 4 stream 2 array 0x18936d00
TestProducer3 Event 4 stream 2
TestProducer2::produce Event 4 stream 2
TestProducer  Event 5 stream 2 ES int 42 FED 1200 size 1864
EDM stream 2 offload to SYCL host device
TestProducer  Event 6 stream 0 ES int 42 FED 1200 size 1936
EDM stream 0 offload to Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
TestProducer2::acquire Event 5 stream 2 array 0x18936d00
TestProducer3 Event 5 stream 2
TestProducer2::produce Event 3 stream 1
TestProducer  Event 7 stream 1 ES int 42 FED 1200 size 2024
EDM stream 1 offload to Intel(R) Gen9 HD Graphics NEO
TestProducer2::acquire Event 6 stream 0 array 0x7fe8a866a800
TestProducer3 Event 6 stream 0
TestProducer2::produce Event 5 stream 2
TestProducer  Event 8 stream 2 ES int 42 FED 1200 size 1976
EDM stream 2 offload to SYCL host device
TestProducer2::acquire Event 7 stream 1 array 0x18994000
TestProducer3 Event 7 stream 1
TestProducer2::acquire Event 8 stream 2 array 0x7fe8a87a7180
TestProducer2::produce Event 6 stream 0
TestProducer3 Event 8 stream 2
TestProducer2::produce Event 7 stream 1
TestProducer  Event 9 stream 1 ES int 42 FED 1200 size 2320
EDM stream 1 offload to Intel(R) Gen9 HD Graphics NEO
TestProducer  Event 10 stream 0 ES int 42 FED 1200 size 3080
EDM stream 0 offload to Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
TestProducer2::acquire Event 9 stream 1 array 0x7fe8a8108000
TestProducer3 Event 9 stream 1
TestProducer2::produce Event 9 stream 1
TestProducer2::produce Event 8 stream 2
TestProducer3 Event 10 stream 0
TestProducer2::acquire Event 10 stream 0 array 0x19575000
TestProducer2::produce Event 10 stream 0
TestProducer2::endJob processed 10 events
Processed 10 events in 9.827208e+00 seconds, throughput 1.01758 events/s.
```

src/sycltest/plugin-Test1/gpuAlgo1.cc
makortel: Thanks. The blog post explicitly mentions a transpose (even if not explicitly relative to CUDA).

Should we look into the SYCL-provided "linearization functions" (beyond this PR)?

makortel: > Should we look into the SYCL-provided "linearization functions" (beyond this PR)?

Ah, I see now that you did already :)

src/sycltest/SYCLCore/device_unique_ptr.h
makortel: Ok, thanks.

src/cudatest/plugin-Test1/gpuAlgo1.cu
```diff
@@ -40,7 +40,7 @@ namespace {
 
   template <typename T>
   __global__ void matrixMulVector(const T *a, const T *b, T *c, int numElements) {
-    int row = blockIdx.y * blockDim.y + threadIdx.y;
+    int row = blockIdx.x * blockDim.x + threadIdx.x;
```
makortel, line 43: Nice, thanks. (facepalm)

makortel:
> With the latest commit the test runs on all reasonable backends:

Very nice. Now I would just have to figure out a way to test it myself :)

src/sycltest/plugin-Test2/gpuAlgo2.cc
```diff
@@ -0,0 +1,202 @@
...
+  stream.submit([&](sycl::handler &cgh) {
+    auto pd_mc = d_mc.get();
+    auto pd_b = d_b.get();
+    auto pd_c = d_c.get();
+
+    cgh.parallel_for(grid, [=](sycl::nd_item<1> item) { matrixMulVector(pd_mc, pd_b, pd_c, NUM_VALUES, item); });
+  });
+
+  // wait for all asynchronous work to complete before destroying the device memory buffers
+  stream.submit_barrier().wait();
```
makortel, line 199: What happens if this synchronization is not done? Program crashes?

fwyzard: Yes... segmentation fault in one of the kernels.

Makefile
```diff
@@ -123,11 +123,40 @@ export KOKKOS_DEPS := $(KOKKOS_LIB)
...
-export KOKKOS_CUFLAGS := $(KOKKOS_CUDA_CUFLAGS) -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored
+export KOKKOS_CUFLAGS := $(KOKKOS_CUDA_CUFLAGS)
 export KOKKOS_LDFLAGS := -L$(KOKKOS_INSTALL)/lib -lkokkoscore -ldl
 export KOKKOS_DLINKFLAGS := $(KOKKOS_CUDA_DLINKFLAGS)
 export NVCC_WRAPPER_DEFAULT_COMPILER := $(CXX)
 
+# SYCL / oneAPI and compatibility tool
+SYCL_BASE  := /opt/intel/oneapi/compiler/latest/linux
+DPCT_BASE  := /opt/intel/oneapi/dpcpp-ct/latest
+ONEAPI_ENV := /opt/intel/oneapi/setvars.sh
```
makortel, line 134: How about
```suggestion
ONEAPI_BASE := /opt/intel/oneapi
SYCL_BASE  := $(ONEAPI_BASE)/compiler/latest/linux
DPCT_BASE  := $(ONEAPI_BASE)/dpcpp-ct/latest
ONEAPI_ENV := $(ONEAPI_BASE)/setvars.sh
```
? Then it would be sufficient to override just `ONEAPI_BASE` for a different location of oneAPI distribution (while still preserving the ability to override individual variables)

Makefile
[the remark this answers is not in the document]
makortel: > Do you prefer that we add it by hand / from the command line when needed ?

Yes (essentially to mimic the behavior of CMSSW)

> By the way, compressed debug sections are also not supported by `cuda-gdb`, so I would really be in favour of removing this option.

Ok, let's keep it removed.

For the remaining options, I'd suggest to leave them in `CXXFLAGS` , but define later in the `Makefile` e.g.
```Makefile
export SYCL_CXXFLAGS := $(filter-out option1 option2,$(CXXFLAGS))
```
and replace `$(CXXFLAGS)` with `$(SYCL_CXXFLAGS)` in `src/sycltest/Makefile`. Then it would be self-documented which options (that we pass to gcc) need to be removed (even if they are likely removed from CMSSW clang builds as well).

Makefile
fwyzard: The reasons is an attempt of compatibility with the open source llvm/clang/sycl compiler, which does not have `dpct` and `setvars.sh`.

For `dpct` I take it from the latest oneAPI beta, so it need a potentially different `BASE` directory.
For `setvars.sh` just do without.

Let me see if I can play with the variables to use a common default, at least.

fwyzard: ok, it's probably not worth doing anything extra complicated right now... I'll have to revisit this anyway next time I try a different compiler.

Makefile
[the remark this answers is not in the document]
fwyzard: OK, I've done something like what you suggested.

src/sycltest/SYCLCore/ScopedContext.h
```diff
@@ -0,0 +1,233 @@
...
+
+    private:
+      friend class cudatest::TestScopedContext;
+
+      // This construcor is only meant for testing
+      explicit ScopedContextProduce(sycl::queue stream, sycl::event event)
+          : ScopedContextGetterBase(stream), event_{event} {}
+
+      // the barrier should be a no-op on an ordered queue, but is used to initialise the event on the data stream
+      sycl::event event_ = stream().submit_barrier();
```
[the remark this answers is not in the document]
fwyzard: It does:

> `event()`:
> Constructs a host `event` that is immediately ready.  The `event` has no dependencies and no associated commands.   > Waiting on this `event` will return immediately.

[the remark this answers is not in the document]
fwyzard: However that is associated to the wrong device and queue, so I'm not sure it's the best thing to use.

src/sycltest/Makefile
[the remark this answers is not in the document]
makortel: Opened https://github.com/cms-patatrack/pixeltrack-standalone/issues/105

src/sycltest/plugin-Test2/gpuAlgo2.cc
makortel: I opened #106 to discuss how to remove that synchronization.

src/sycltest/SYCLCore/ScopedContext.h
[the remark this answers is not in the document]
makortel: That would actually be fine, because the `event_` is not used for synchronization before the destructor. Wait. That makes me realize that copying the event usage pattern from CUDA does not actually work here.

With CUDA we create the event in the `ScopedContextProduce` constructor, distribute a pointer to it to all the `Product<T>`s, and record the event in the destructor of `ScopedContextProduce`.

Here, an event is created in the `ScopedContextProduce` constructor, it is copied to all the `Product<T>`s, and a new event is assigned in the `ScopedContextProduce` destructor. But that event is now disconnected from the events of the `Product<T>`s.

Instead, we should do (e.g.) something like constructing a `std::shared_ptr<sycl::event>`in the constructor (the default constructor would be fine, IIUC), distributing the `shared_ptr` to the `Product<T>`s, and in the destructor assign the event to the de-referenced `shared_ptr`.

Written that, I would be fine proceeding with the current (broken) mode and fixing it in a subsequent PR (and I could volunteer to do it).

src/sycltest/Makefile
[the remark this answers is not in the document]
makortel: The host "device" could be easily added by
```Makefile
test_cpu:
        @echo
        @echo "Testing $(TARGET)"
        SYCL_DEVICE_TYPE=CPU $(TARGET) --maxEvents 2
        @echo "Succeeded"
```
after which `make test_sycl_cpu` succeeds (assuming https://github.com/cms-patatrack/pixeltrack-standalone/pull/102#discussion_r498563646 is changed as well).

Going beyond that gets tricky. `test_cuda` target makes no sense without CUDA device support, so probably for time being the `test_sycl` is the best we can do. I try to improve (a bit later) how that gets hooked in the top-level `Makefile` (might require a bit more work, because this PR changes the "big picture" quite a lot actually from the earlier "CPU" or "CUDA").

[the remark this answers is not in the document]
fwyzard: Mhm... long term looks like we could have various options:
  - `cpu`: for SYCL, means two options: `host` and `cpu`; also for Alpaka it can map to at least two backends: `serial` and `tbb`
  - `cuda`: for SYCL needs the codeplay and/or open source compiler, I'll look into it in a while
  - `intel` (?):  intel gpu, which can go through two back-ends, OpenCL and Level 0... do we want to test both?
  - `auto` / `all`: all supported backends

Makefile
```diff
@@ -3,9 +3,9 @@ export BASE_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
...
+HOST_CXXFLAGS := -O2 -fPIC -fdiagnostics-show-option -felide-constructors -fmessage-length=0 -fno-math-errno -ftree-vectorize -fvisibility-inlines-hidden --param vect-max-version-for-alias-checks=50 -msse3 -pipe -pthread -Werror=address -Wall -Werror=array-bounds -Wno-attributes -Werror=conversion-null -Werror=delete-non-virtual-dtor -Wno-deprecated -Werror=format-contains-nul -Werror=format -Wno-long-long -Werror=main -Werror=missing-braces -Werror=narrowing -Wno-non-template-friend -Wnon-virtual-dtor -Werror=overflow -Werror=overlength-strings -Wparentheses -Werror=pointer-arith -Wno-psabi -Werror=reorder -Werror=return-local-addr -Wreturn-type -Werror=return-type -Werror=sign-compare -Werror=strict-aliasing -Wstrict-overflow -Werror=switch -Werror=type-limits -Wunused -Werror=unused-but-set-variable -Wno-unused-local-typedefs -Werror=unused-value -Wno-error=unused-variable -Wno-vla -Werror=write-strings
 export CXXFLAGS := -std=c++17 $(HOST_CXXFLAGS) $(USER_CXXFLAGS)
-export LDFLAGS := -pthread -Wl,-E -lstdc++fs
+export LDFLAGS := -O2 -g -fPIC -pthread -Wl,-E -lstdc++fs
```
makortel, line 8: Could the `-g` be removed from here as well?
```suggestion
export LDFLAGS := -O2 -fPIC -pthread -Wl,-E -lstdc++fs
```
Or would it actually do anything if the object files do not contain debug symbols?

makortel:
I'm thinking to being ready to merge this PR as it is (@fwyzard tell me if you think otherwise). Even addressing the testing discussion in https://github.com/cms-patatrack/pixeltrack-standalone/pull/102#discussion_r498563352 would have the implication that `make test` (as `make all` already does now) would require both `nvcc` and SYCL compiler. I'd like to be able to have some simple way to build and test the part that requires only `nvcc`, and that would require some further reorganization of Makefile targets, that could be best done after this PR.

fwyzard:
Yes, I think we can merge this as-is, and work on improving things in future PRs.

PR #109, opened 2020-10-12
various CUDA-related improvements

fwyzard:
Hard code CUDA version 10.2, since CUDA 11.0 fails to run and CUDA 11.1 fails to build.
Silence CUDA warnings about deprecated architectures.
Remove duplicate use of `esa_on_defaulted_function_ignored`.
Fix a bug in the matrixMulVector test.


fwyzard:
I've moved here the CUDA related fixes from #102 .

PR #110, opened 2020-10-12
[cudadev] Get rid of cub prefix-scan

PointKernel:
From https://github.com/cms-patatrack/cmssw/pull/447.

### DIFF
--- a/src/cudadev/CUDACore/HistoContainer.h
+++ b/src/cudadev/CUDACore/HistoContainer.h
@@ -9,10 +9,6 @@
 #include <cstdint>
 #include <type_traits>
 
-#ifdef __CUDACC__
-#include <cub/cub.cuh>
-#endif
-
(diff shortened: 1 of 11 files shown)


makortel:
Could you copy the changes under `cudadev` and leave `cuda` untouched? I'd prefer to keep the latter frozen (for now at least).

PointKernel:
https://github.com/cms-patatrack/cmssw/pull/447/

PointKernel:
Forgot to mention, it's https://github.com/cms-patatrack/cmssw/pull/447 only. I didn't completely remove cub dependency but only replaced cub prefix-scan.

makortel:
Ah, ok, thanks (I corrected the description, should have looked more carefully).

PointKernel:
> Could you copy the changes under `cudadev` and leave `cuda` untouched? I'd prefer to keep the latter frozen (for now at least).

Sure, I can do that.

PR #112, opened 2020-10-16
Make distinction between gcc+nvcc and SYCL toolchains more clear in the all and test make targets

makortel:
First small step for improve the `all` and `test` targets distinguishing the gcc+nvcc and SYCL toolchains.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,14 +163,25 @@ endif
 
 # Targets and their dependencies on externals
 TARGETS := $(notdir $(wildcard $(SRC_DIR)/*))
-TEST_CPU_TARGETS := $(patsubst %,test_%_cpu,$(TARGETS))
-TEST_CUDA_TARGETS := $(patsubst %,test_%_cuda,$(TARGETS))
+TARGETS_SYCL = sycltest
+TARGETS_GCC_NVCC = $(filter-out $(TARGETS_SYCL),$(TARGETS))
+
+TEST_CPU_TARGETS := $(patsubst %,test_%_cpu,$(TARGETS_GCC_NVCC))
+TEST_CUDA_TARGETS := $(patsubst %,test_%_cuda,$(TARGETS_GCC_NVCC))
+TEST_SYCL_TARGETS := $(patsubst %,test_%_cuda,$(TARGETS_SYCL))
+
 all: $(TARGETS)
-test: test_cpu test_cuda
+all_gcc_nvcc: $(TARGETS_GCC_NVCC)
+all_sycl: $(TARGETS_SYCL)
+
+test: test_cpu test_cuda test_sycl
 test_cpu: $(TEST_CPU_TARGETS)
 test_cuda: $(TEST_CUDA_TARGETS)
+test_sycl: $(TEST_SYCL_TARGETS)
+
 # $(TARGETS) needs to be PHONY because only the called Makefile knows their dependencies
-.PHONY: all $(TARGETS) test test_cpu test_cuda $(TEST_CPU_TARGETS) $(TEST_CUDA_TARGETS)
+.PHONY: all all_gcc_nvcc $(TARGETS)
+.PHONY: test test_cpu test_cuda $(TEST_CPU_TARGETS) $(TEST_CUDA_TARGETS) $(TEST_SYCL_TARGETS)
 .PHONY: environment format clean distclean dataclean external_tbb external_cub external_eigen external_kokkos external_kokkos_clean
 
 environment: env.sh


Makefile
```diff
@@ -163,14 +163,25 @@ endif
...
+
 all: $(TARGETS)
-test: test_cpu test_cuda
+all_gcc_nvcc: $(TARGETS_GCC_NVCC)
+all_sycl: $(TARGETS_SYCL)
+
+test: test_cpu test_cuda test_sycl
 test_cpu: $(TEST_CPU_TARGETS)
 test_cuda: $(TEST_CUDA_TARGETS)
+test_sycl: $(TEST_SYCL_TARGETS)
```
makortel, line 180: I'm not completely happy with this as the possible `test_cpu` target from `sycltest` would be included in the `test_cpu` target here as well, therefore bringing in the dependence on SYCL toolchain.

How about using 
* the existence of `$(CUDA_BASE)` directory for `nvcc`
* `$(SYCL_BASE)` being non-empty for SYCL

as criteria for including CUDA and SYCL-dependent programs to `all` and `test_X`?

fwyzard:
hi Matti,
I cannot comment much, mostly because I'm not sure what it is you would
like to achieve.

The targets I can foresee are:

native code
  - cpu (native or cudacompat)
  - nvidia gpu (cuda)

alpaka
  - cpu (serial)
  - cpu (parallel via tbb)
  - nvidia gpu (cuda backend)
  - amd gpus ?

kokkos
  - same as alpaka ?

sycl/oneapi
  - cpu (host device)
  - cpu (parallel via opencl backend)
  - intel gpu (opencl and/or level 0 backend)
  - nvidia gpu (cuda backend)
  - eventually amd gpu ?


So, how should we arrange the tests ?

makortel:
What I'm after are "easy" targets for building and running tests for everything that can be run for a given "toolchain" (or compiler). These would be mainly for PR testing, I presume for actual development the program-specific target is enough for building, and unit tests can be somewhat easily run directly (although we are missing a general "run all unit tests for a given program" target, mainly because of name clash with `test_cuda`, which runs all tests that use CUDA).

Today the toolchains are
* GCC 8 + NVCC (10.2)
* oneAPI

In principle "GCC only" could be there as well (although that would include only `fwtest` program). In the future AMD GPUs could be there as well, both directly via HIP/ROCm and via a SYCL compiler. I suppose we could also have multiple SYCL compilers.

Now I realize that while a toolchain is sufficient to specify the building, the actual hardware adds another dimension for actually running tests. (and cross-compilation and ability to run "everything else than the missing piece of hardware" is something I want to be able to test).

Assuming the use of `CUDA_BASE` and `SYCL_BASE` to deduce whether a toolchain is present would not be too confusing, we could use those to decide which programs to include for the `all` and `test` targets (possibly accompanied with some printouts explaining that). The `test` target could then call hardware-specific test targets, e.g.
* `test_cpu` for all tests that should run on CPU alone
* `test_nvidiagpu` for all tests that require NVIDIA GPU
* `test_intelgpu` for all tests that require Intel GPU
* `test_amdgpu` for all tests that require AMD GPU
* `test_auto` for all tests that use auto-discovery internally (i.e. current `test_sycl` would go there)

fwyzard:
Yes, the tests we could run look more like a matrix.

Does this match what you had in mind ?

  | CPUs | NVIDIA GPUs | Intel GPUs | AMD GPUs | auto
-- | -- | -- | -- | -- | --
**GCC** | `test_cpu` |   |   |   |  
**CUDA** |   | `test_nvidiagpu` |   |  
**Alpaka** | `test_cpu` (serial, tbb) | `test_nvidiagpu` | ? | ? | `test_auto`
**Kokkos** | `test_cpu` | `test_nvidiagpu` | ? | ? | `test_auto`
**SYCL** (*oneAPI*, *CodePlay*, *LLVM*) | `test_cpu` (host, opencl) | `test_nvidiagpu` | `test_intelgpu` (level 0, opencl) | ? | `test_auto`

makortel:
> Does this match what you had in mind ?

Yes, something along that (thanks, the table is a nice summary). Then the remaining question is whether the existence of CUDA and SYCL compilers is good-enough to decide which "technologies" of the first column are available to define the exact behavior of the top-level `all` and `test` (and `test_<arch>`) targets is good enough (I hope it would be, as the the way of explicit coding into the target names along `all_gcc`, `all_nvcc`, `all_sycl` is not very attractive to me).

With the proposed naming convention it would be straightforward to provide also `test_<program>` targets, in addition to the  `test_<program>_<arch>` targets that already exist.

makortel:
Superseded by #113.

PR #113, opened 2020-10-21
Use CUDA_BASE and SYCL_BASE to choose targets to include in all, create hardware-specific test targets

makortel:
New attempt to be more flexible with the build targets, based on the discussion in https://github.com/cms-patatrack/pixeltrack-standalone/pull/112.


makortel:
I'm going to merge this now because I have further developments that would conflict. Problems or improvements can be dealt with later.

PR #116, opened 2020-10-29
[kokkos] Generalize build rules, add support for PTHREAD backend

makortel:
The default set of backends is still kept in Serial+CUDA, because PThread backend is incompatible with an event loop implemented as TBB tasks (which I still prefer to keep the default, to mimic CMSSW).

I have tested that the combinations
* serial only
* serial+pthread
* serial+CUDA
* serial+pthread+CUDA 

compile and run.


src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
```diff
@@ -45,10 +45,10 @@ namespace KOKKOS_NAMESPACE {
...
     teamSize *= stride;
 
-#ifdef KOKKOS_BACKEND_SERIAL
-    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, 1};
-    // unit stride loop for serial execution
-    stride = 1;
+#if defined KOKKOS_BACKEND_SERIAL || defined KOKKOS_BACKEND_PTHREAD
+    // for host backends use the same stride as the number of threads in the team
+    stride = KokkosExecSpace::impl_thread_pool_size();
+    Kokkos::TeamPolicy<KokkosExecSpace> policy{execSpace, leagueSize, stride};
```
makortel, line 51: @PointKernel Could you double-check my changes in this file?

src/kokkos/plugin-PixelVertexFinding/kokkos/gpuVertexFinder.cc
```diff
@@ -130,8 +130,8 @@ namespace KOKKOS_NAMESPACE {
           Kokkos::RangePolicy<KokkosExecSpace>(execSpace, 0, TkSoA::stride()),
           KOKKOS_LAMBDA(const size_t i) { loadTracks(tksoa, vertices_d, workspace_d, ptMin, i); });
 
-#ifdef KOKKOS_BACKEND_SERIAL
-      auto policy = TeamPolicy(execSpace, 1, 1).set_scratch_size(0, Kokkos::PerTeam(8192 * 4));
+#if defined KOKKOS_BACKEND_SERIAL || defined KOKKOS_BACKEND_PTHREAD
+      auto policy = TeamPolicy(execSpace, 1, Kokkos::AUTO()).set_scratch_size(0, Kokkos::PerTeam(8192 * 4));
```
makortel, line 134: @PointKernel Could you double-check my changes in this file?

src/kokkos/plugin-SiPixelClusterizer/kokkos/SiPixelRawToClusterGPUKernel.cc
```diff
@@ -688,14 +688,11 @@ namespace KOKKOS_NAMESPACE {
         Kokkos::deep_copy(
             execSpace, Kokkos::subview(nModules_Clusters_h, 0), Kokkos::subview(clusters_d.moduleStart(), 0));
 
-#ifdef KOKKOS_BACKEND_SERIAL
-        const uint32_t threadsPerBlock = 1;
-#else
-        const uint32_t threadsPerBlock = 256;
-#endif
         const uint32_t blocks = ::gpuClustering::MaxNumModules;
+        Kokkos::TeamPolicy<KokkosExecSpace> teamPolicy(execSpace, blocks, Kokkos::AUTO());
```
makortel, line 692: @jtchilders Could you double-check my changes here and in `gpuClustering.h` and `gpuClusterChargeCut.h`?

src/kokkos/test/kokkos/OneHistoContainer_t.cc
```diff
@@ -65,14 +65,16 @@ void go() {
         });
 
     Kokkos::deep_copy(KokkosExecSpace(), histo_h, histo_d);
-    assert(histo_h().off[-1] == 0);
+    // TODO: fails to compile with GCC, is this comparison really needed?
+    //assert(histo_h().off[-1] == 0);
```
makortel, line 69: @jtchilders Can this assert be removed?

src/kokkos/test/kokkos/OneHistoContainer_t.cc
```diff
@@ -65,14 +65,16 @@ void go() {
...
     assert(0 == histo_h().size());
 
     TeamHist::finalize(histo_d, KokkosExecSpace());
 
     Kokkos::deep_copy(KokkosExecSpace(), histo_h, histo_d);
 
-    assert(histo_h().off[-1] == histo_h().off[-2]);
+    // TODO: fails to compile with GCC, is this comparison really needed?
+    //assert(histo_h().off[-1] == histo_h().off[-2]);
```
makortel, line 77: @jtchilders Can this assert be removed?

src/kokkos/test/kokkos/OneHistoContainer_t.cc
jtchilders: I presume it was in the original which is why I left it. My guess is that it is a sanity check and can be removed.

src/kokkos/test/kokkos/OneHistoContainer_t.cc
jtchilders: same comment here.

src/kokkos/test/kokkos/OneHistoContainer_t.cc
makortel: Thanks, removed.

src/kokkos/plugin-PixelVertexFinding/kokkos/gpuVertexFinder.cc
PointKernel: I don't see any issues here.

src/kokkos/plugin-PixelTriplets/kokkos/CAHitNtupletGeneratorKernels.cc
PointKernel: Does this give the correct numerical results? I can understand if the pthread policy is defined in a CUDA-similar way, i.e. adjusting team size and league size based on a predefined stride size.  Making the stride size equal to team size will theoretically work cause the kernels involved (like `fishbone`) are just to iterate to a certain upper bound. But the algorithm is different from the original CUDA implementation. It's hard to tell without evaluating the eventual output.

makortel: I'm somewhat convinced that the results are correct. At least the pthread backend passes the counting-based validation (which has been rather picky) at least up to 8 threads. I think the `stride` mostly adjusts the balance between the two nested loops in the `getDoublets`, `kernel_connect`, and `fishbone` kernels.

makortel: I could actually just leave the stride 1 for pthread backend as well, on a quick test it in fact appears to scale better

PR #119, opened 2020-11-03
[kokkos] Replace Nx parallel_scan() with one parallel_scan()+2x parallel_for() in HistoContainer finalize()

makortel:
This is a more performance workaround for Kokkos not providing team-wide prefix scan.


makortel:
@PointKernel @jtchilders Please take a look

PointKernel:
Looks good! A minor thing here:https://github.com/cms-patatrack/pixeltrack-standalone/blob/a393add0eb827c2a58382d5fc3e720bc22b8a0c0/src/kokkos/KokkosCore/HistoContainer.h#L89
We forgot to add a label to this `parallel_scan`.

makortel:
> We forgot to add a label to this `parallel_scan`.

Thanks, I added a label to that now.

PR #143, opened 2020-12-03
Implement event batching in the miniframework

fwyzard:
## Introduce a new way of passing information from `acquire` to `produce`:
  - add a template type parameter to the `EDProducerExternalWork` base class, and make it available to inherited classes as `AsyncState`;
  - `acquire` should use an `AsyncState` object hold all the information that needs to be passed to `produce`; it is now marked as ` const` method to guarantee that the information is *not* passed through some internal state;
  - `produce` takes an extra `AsyncState` parameter to receive the state information from `acquire`;
  - if a Producer does not need to pass any state from `acquire` to `produce` it can use a `void` or empty template parameter; in this case `acquire` returns `void` and `produce` does not take any new parameters (as before); `acquire` is still marked as `const`.

This change is propaedeutic to the event batching support, but can be reviewed and adopted independently of it

##  Add a `doneWaiting` overload that does not take an exception pointer argument
This is just a simplification for the cases where `doneWaiting(...)` was being called with an empty exception pointer.

This change is independent from event batching support, and can be reviewed and adopted independently of it

## Implement basic support for event batching, trying to have minimal impact on the existing modules:
  - the `Source` is responsible for reading (up to) N events at a time;
  - the `Event`s are stored in an `std::vector<Event>` and passed to each `Worker` as an `std::vector<Event *> const&` (for `produce()`) or ``std::vector<Event const*> const&` (for `acquire`); the `vector` itself is `const` to avoid adding/dropping events, while the underlying events are `const` or not as before;
  - the `Worker` passes the events to the various `EDProducer`-like base classes in the same way;
  - the existing `EDProducer` and `EDProducerExternalWork` base classes loop over the events in the vector, and call `acquire` and `produce` one event at a time; modules that derive from them can be used without any changes;
  - introduce two new `EDBatchingProducer` and `EDBatchingProducerExternalWork` base classes that pass the vector of `Event`s to the module's `acquire` and `produce` methods; the module's implementation can then loop over the events, process them in parallel, etc.

The framework tests have been reorganised and extended to test all four `EDProdcer`-like base classes:
  - `EDProducer`;
  - `EDProducerExternalWork`;
  - `EDBatchingProducer`;
  - `EDBatchingProducerExternalWork`.


fwyzard:
## Alternative approach for `acquire()`

Instead of `acquire` returning a new `AsyncState` object by value (relying on return value optimisation), the `AsyncState` object could be default-constructed by the `EDProducerExternalWork` base class and passed to `acquire` by `AsyncState &`.

This would make the interface closer to that of `produce()` - but adds the requirement that `AsyncState` be default-constructible.

fwyzard:
## Alternative approach for `produce()`

The `AsyncState` object could be passed to `produce` by `AsyncState &` rather than `AsyncState &&`.
This should make it easier tu support types that are not movable and/or copyable.

fwyzard:
OK, this looks usable now.

The next step will be to try and make the CUDA implementation use it ...

makortel:
> The next step will be to try and make the CUDA implementation use it ...

It will take me some time to digest this. Organizationally I'd prefer if this version of `fwtest` would be made as a separate program (it's kind of a branching point), because I'm expecting eventually to have a multiple attempts of batching while it could be useful to have many/all of them around. I'll anyway go though this PR as it is.

fwyzard:
> Organizationally I'd prefer if this version of `fwtest` would be made as a separate program

Sure. Making the PR on top of the existing `fwtest` makes it easier to see all the changes, but if/when we decide to merge it it can go in a different directory.

fwyzard:
I'm now looking how to apply this to the `cuda` code base.

The first problem I ran into is that the `AsyncState` I need to define for some external work producers is not copyable or movable, because it includes a `cms::cuda::ContextState`.

Just to show how much I didn't know about c++:
  - even when return value optimisation is mandated by the standard, a non-deleted move or copy constructir is still required
  - `std::vector<T>::resize()` and `std::vector<T>::reserve()` require that `T` is copyable or movable (I don't know why)

So I'm using a third approach:
  - the `AsyncState`s are held in a `std::unique_ptr<AsyncSytate[]>` or `std::optional<AsyncState>`
  - they are recreated for every event or batch of events
  - they are passed by `AsyncState &` both to `acquire()` and `produce()`
  - they are reset after each event or batch of events

With this approach the only requirement on `AsyncState` is that it can be default constructed; I think this should be fine, since using a data member to hold some state between `acquire()` and `produce()` probably has the same requirement - and if some type does not satisfy it, it can always be wrapped in a `unique_ptr` ...

fwyzard:
The implementation above incurs a huge performance penalty, because the `AsyncState` variables are now constructed and destructed for every event.

This avoids it:
```diff
diff --git a/src/cuda/Framework/EDProducer.h b/src/cuda/Framework/EDProducer.h
index 17e7e1be1bdb..76e39e93bf62 100644
--- a/src/cuda/Framework/EDProducer.h
+++ b/src/cuda/Framework/EDProducer.h
@@ -38,27 +38,25 @@ namespace edm {
     bool hasAcquire() const { return true; }
 
     void doAcquire(Event const& event, EventSetup const& eventSetup, WaitingTaskWithArenaHolder holder) {
-      state_.reset();
-      state_.emplace();
-      acquire(event, eventSetup, std::move(holder), *state_);
+      acquire(event, eventSetup, std::move(holder), state_);
     }
 
     void doProduce(Event& event, EventSetup const& eventSetup) {
-      produce(event, eventSetup, *state_);
-      state_.reset();
+      produce(event, eventSetup, state_);
     }
 
     virtual void acquire(Event const& event,
                          EventSetup const& eventSetup,
                          WaitingTaskWithArenaHolder holder,
                          AsyncState& state) const = 0;
+
     virtual void produce(Event& event, EventSetup const& eventSetup, AsyncState& state) = 0;
 
     void doEndJob() { endJob(); }
     virtual void endJob() {}
 
   private:
-    std::optional<AsyncState> state_;
+    AsyncState state_;
   };
 
   template <>
```

but I'm not sure that's a good approach ?

src/cuda/Framework/EDProducer.h
```diff
@@ -27,27 +34,157 @@ namespace edm {
...
+    }
+
+    virtual void acquire(ConstEventRange events,
+                         EventSetup const& eventSetup,
+                         WaitingTaskWithArenaHolder holder,
+                         AsyncState& state) const = 0;
+
+    void doProduce(EventRange events, EventSetup const& eventSetup) { produce(events, eventSetup, state_); }
+
+    virtual void produce(EventRange events, EventSetup const& eventSetup, AsyncState& states) = 0;
```
makortel, line 151: Really minor, but is this a typo?
```suggestion
    virtual void produce(EventRange events, EventSetup const& eventSetup, AsyncState& state) = 0;
```

src/cuda/Framework/EventBatch.h
```diff
@@ -0,0 +1,36 @@
...
+
+    void clear() { events_.clear(); }
+
+    size_t size() const { return events_.size(); }
+
+    bool empty() const { return events_.empty(); }
+
+    EventRange range() { return EventRange(&events_.front(), &events_.back() + 1); }
+
+    ConstEventRange range() const { return ConstEventRange(&events_.front(), &events_.back() + 1); }
```
makortel, line 28: Not very relevant for prototyping itself, but i think being able to use something along
```suggestion
    EventRange range() { return EventRange(events_.begin(), events_.end()); }

    ConstEventRange range() const { return ConstEventRange(events_.begin(), events_.end()); }
```
would be clearer to read.

makortel (commented):
I think this is a reasonable approach for a fixed-size batching and having total number of concurrent events to be "batch size * number of streams".

I found only two rather nitpicking-level comments (which may mean that I didn't digest all in the end). I would not worry too much about the interfaces (as long as they are reasonable) before performance is compared against the current non-batched approach.

makortel:
> The previous implementation incurred in a huge performance penalty, because the `AsyncState` variables were now constructed and destructed for every event.

Interesting. The `AsyncState` variables are essentially bunch of smart pointers and numbers (and one `map<unsigned, vector<SiPixelRawDataError>>`), is their construction and destruction really that expensive? (I'm probably missing something)

makortel:
> The first problem I ran into is that the `AsyncState` I need to define for some external work producers is not copyable or movable, because it includes a `cms::cuda::ContextState`.

For prototyping purposes I think making `cms::cuda::ContextState` e.g. movable is perfectly fine. The only reason for denying moving is to catch unintended usage early.

src/cuda/Framework/EventBatch.h
fwyzard: The reason I didn't do that was to avoid making `EventRange` and `ConstEventRange` use `std::vector<Event>::iterator` and `const_iterator`: then it could only work for a range of events stored in a `vector`, while using plain pointers it can work with any consecutive range of events.

An alternative could be to add a constructor like
```c++
template <typename IT>
EventRange(IT begin, IT end) : begin_(std::addressof(*begin)), end_(std::addressof((*end)) { ... }
```
but then one could be tempted to pass non-contiguous iterators.
Maybe one could add something like
```c++
  assert((end - begin) == (end_ - begin_));
```
or,  once we move to C++20 with concepts, we could use `contiguous_iterator` ?

src/cuda/Framework/EDProducer.h
fwyzard: I've used both spellings while prototyping, so I don't mind either way: this should capture the "state" or "states" of multiple events:
  - `state` is more consistent with `AsyncState`
  - `states` is more consistent with `events`

Re-thinking a bit more... maybe `state` is more correct, since it is one object, not *e.g.* a `vector<AsyncState>`.

fwyzard:
> > The previous implementation incurred in a huge performance penalty, because the `AsyncState` variables were now constructed and destructed for every event.
> 
> Interesting. The `AsyncState` variables are essentially bunch of smart pointers and numbers (and one `map<unsigned, vector<SiPixelRawDataError>>`), is their construction and destruction really that expensive? (I'm probably missing something)

I was surprised as well, and digging into the cause I found it to be `SiPixelRawToClusterCUDA` that needs to pass a `pixelgpudetails::SiPixelRawToClusterGPUKernel::WordFedAppender` from `acquire()` to `produce()`.
`WordFedAppender` has two `cms::cuda::host::noncached::unique_ptr<unsigned int[]>` and `cms::cuda::host::noncached::unique_ptr<unsigned char[]>` data members, so re-creating it for every event becomes prohibitely expensive.

I guess we could find a workaround - but avoiding constructing and destructing `AsyncState` on every event seemed the simplest solution.

src/cuda/Framework/EventBatch.h
makortel: For the purposes of the prototype the current pointers are fine for me. With C++20 maybe even passing the range as `std::span` could be an option.

makortel:
> I was surprised as well, and digging into the cause I found it to be `SiPixelRawToClusterCUDA` that needs to pass a `pixelgpudetails::SiPixelRawToClusterGPUKernel::WordFedAppender` from `acquire()` to `produce()`.
> `WordFedAppender` has two `cms::cuda::host::noncached::unique_ptr<unsigned int[]>` and `cms::cuda::host::noncached::unique_ptr<unsigned char[]>` data members, so re-creating it for every event becomes prohibitely expensive.

Ah, `WordFedAppender`, missed that. Right, that one has been intended to be "allocate once per stream and re-use". On the other hand, it would also be straightforward to change it (back) to use the caching allocator given the tiny impact of `cudaHostAllocWriteCombined`. Counter-OTOH, it could also be useful to keep it around as it is to have a use case for a stream-object that should not be constructed for each event.

makortel:
Just to repeat here what I mentioned in the core sw meeting today. I think an object that should live longer than "from `acquire()` to `produce()`" (e.g. `WordFedAppender` currently) should be stored as a stream module's member variable (or in StreamCache of a global module).

And I think I finally realized the issue with that (within this PR) 
* `EDProducerExternalWork<T>` calls first the `acquire()` for all events in the batch, and only then the `produce()` for all events in the batch
* `EDBatchingProducerExternalWork<T>` calls the `acquire()` once for the full batch, and then the `produce()` for the full batch

Therefore `EDBatchingProducerExternalWork<T>` can use the stream member variable/cache to pass something from `acquire()` to `produce()`, because it has to deal with the batching explicitly anyway. But `EDProducerExternalWork<T>` can not, because the values written in `acquire()` call of the first event of the batch would be overwritten by the call of second event of the batch etc. And that would effectively necessitate the concept of "event cache" in addition to "stream cache", because a stream would no longer be a stream of events but a stream of event batches, even if we would solve https://github.com/cms-sw/cmssw/issues/24186.

Would it be fair to say that the `AsyncState` in this PR is effectively such an "event cache"?

fwyzard:
> Would it be fair to say that the `AsyncState` in this PR is effectively such an "event cache"?

Even if that is not the goal of the interface, in practice yes, because it is available (and mutable) in both `acquire()` and `produce()` - which are the only event processing methods.

makortel:
Given that this batching approach needs an "event cache" in any case, how important do you think a separate mechanism to pass data from `acquire()` to `produce()` (https://github.com/cms-sw/cmssw/issues/24186) would be?

fwyzard:
I have the feeling that an "event cache" and a "pass data from acquire to produce" mechanism could be to some extent orthogonal:
  - if we have acquire-to-produce, most existing ExternalWork modules would not need an event cache to support batching
  - even without ExternalWork, modules may need some kind of event cache to support batching - be it something done explicitly by the module or provided by the framework

**If** we introduce an event cache, and **if** the interface for the event cache and the acquire-to-produce data would be the same (i.e. additional template parameter, additional parameter passed to acquire and produce) then of course having both at the same tie would only add complexity.

PR #148, opened 2020-12-11
Use the correct version of Eigen

### LINKED ISSUE #147: Eigen version is incorrect
I noticed that despite of intention the Eigen commit hash used here (`d812f411c`) is not the same as used in CMSSW (`b20a61c3a0dc9a79790cd258130a99b574662272`). I made the mistake on relying on the version reported by `scram tool info eigen` and not checking the actual contents of the spec file
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/eigen.spec#L6

makortel:
This new one is the one used in CMSSW_11_2_0_pre8_Patatrack (and earlier), the one used before is the commit used as the base of patches applied by CMS.

Resolves #147.

Update needs either `make distclean` or `rm -fR external/eigen`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -368,7 +368,7 @@ external_eigen: $(EIGEN_BASE)
 
 $(EIGEN_BASE):
 	git clone https://github.com/cms-externals/eigen-git-mirror $@
-	cd $@ && git checkout -b cms_branch d812f411c3f9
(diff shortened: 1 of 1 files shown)


makortel:
No impact in results or performance (as expected).

PR #157, opened 2021-01-11
[alpakatest] Add return types to alpakatest algos

ghugo83:
Event data can now be stored, and accessed with getToken directly from Acc memory, no need to transfer data to CPU.

NB: At a later stage: There should likely be implementations in AlpakaCore of ScopedContext, similar to what is done in KokkosCore; skipped for now.

### DIFF
--- a/src/alpaka/Makefile
+++ b/src/alpaka/Makefile
@@ -8,7 +8,8 @@ $(TARGET):
 test_cpu: $(TARGET)
 	@echo
 	@echo "Testing $(TARGET)"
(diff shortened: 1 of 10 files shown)


makortel:
I had to separate the tests of Alpaka's `serial` and `tbb` backends into two separate jobs because of a limitation in the framework to identify event data products only by type. I opened an issue #158 to remind myself to improve that.

ghugo83:
Thanks @makortel for the review

PR #159, opened 2021-01-12
[alpaka] Add AtomicPairCounter to alpaka/AlpakaCore and alpaka/test

ghugo83:
Results as expected for the 3 backends.

### DIFF
--- a/src/alpaka/AlpakaCore/AtomicPairCounter.h
+++ b/src/alpaka/AlpakaCore/AtomicPairCounter.h
@@ -3,8 +3,6 @@
 
 #include <cstdint>
 
-#include "AlpakaCore/alpakaConfig.h"   // needed?
-
 namespace cms {
   namespace Alpaka {
 
@@ -12,8 +10,8 @@ namespace cms {
     public:
       using c_type = unsigned long long int;
 
-      ALPAKA_FN_HOST_ACC AtomicPairCounter() {}  //  needed?
-      ALPAKA_FN_HOST_ACC AtomicPairCounter(c_type i) { counter.ac = i; } //  needed?
+      ALPAKA_FN_HOST_ACC AtomicPairCounter() {}
+      ALPAKA_FN_HOST_ACC AtomicPairCounter(c_type i) { counter.ac = i; }
 
       ALPAKA_FN_HOST_ACC AtomicPairCounter& operator=(c_type i) {
 	counter.ac = i;
@@ -36,9 +34,10 @@ namespace cms {
 
       // increment n by 1 and m by i.  return previous value
       template<typename T_Acc>
-	ALPAKA_FN_ACC ALPAKA_FN_INLINE Counters add(T_Acc const & acc, uint32_t i) {  // not ALPAKA_FN_HOST_ACC ?
+	ALPAKA_FN_ACC ALPAKA_FN_INLINE Counters add(const T_Acc& acc, uint32_t i) {
 	c_type c = i;
 	c += incr;
+	
 	Atomic2 ret;
 	ret.ac = alpaka::atomic::atomicOp<alpaka::atomic::op::Add>(acc, &counter.ac, c);
 	return ret.counters;
--- a/src/alpaka/AlpakaCore/AtomicPairCounter.h
+++ b/src/alpaka/AlpakaCore/AtomicPairCounter.h
@@ -1,0 +1,54 @@
+#ifndef HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
+#define HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
+
+#include <cstdint>
+
+#include "AlpakaCore/alpakaConfig.h"   // needed?
+
+namespace cms {
+  namespace Alpaka {
+
+    class AtomicPairCounter {
+    public:
+      using c_type = unsigned long long int;
+
+      ALPAKA_FN_HOST_ACC AtomicPairCounter() {}  //  needed?
+      ALPAKA_FN_HOST_ACC AtomicPairCounter(c_type i) { counter.ac = i; } //  needed?
+
+      ALPAKA_FN_HOST_ACC AtomicPairCounter& operator=(c_type i) {
+	counter.ac = i;
+	return *this;
+      }
+
+      struct Counters {
+        uint32_t n;  // in a "One to Many" association is the number of "One"
+        uint32_t m;  // in a "One to Many" association is the total number of associations
+      };
+
+      union Atomic2 {
+        Counters counters;
+        c_type ac;
+      };
+
+      static constexpr c_type incr = 1UL << 32;
+
+      ALPAKA_FN_HOST_ACC Counters get() const { return counter.counters; }
+
+      // increment n by 1 and m by i.  return previous value
+      template<typename T_Acc>
+	ALPAKA_FN_ACC ALPAKA_FN_INLINE Counters add(T_Acc const & acc, uint32_t i) {  // not ALPAKA_FN_HOST_ACC ?
+	c_type c = i;
+	c += incr;
+	Atomic2 ret;
+	ret.ac = alpaka::atomic::atomicOp<alpaka::atomic::op::Add>(acc, &counter.ac, c);
+	return ret.counters;
+      }
+
+    private:
+      Atomic2 counter;
+    };
+
+  }  // namespace Alpaka
+}  // namespace cms
+
+#endif  // HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
--- a/src/alpaka/AlpakaCore/AtomicPairCounter.h
+++ b/src/alpaka/AlpakaCore/AtomicPairCounter.h
@@ -14,8 +14,8 @@ namespace cms {
       ALPAKA_FN_HOST_ACC AtomicPairCounter(c_type i) { counter.ac = i; }
 
       ALPAKA_FN_HOST_ACC AtomicPairCounter& operator=(c_type i) {
-	counter.ac = i;
-	return *this;
+        counter.ac = i;
+        return *this;
       }
 
       struct Counters {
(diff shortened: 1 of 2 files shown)


ghugo83:
NB: Also let me know if there is some formatting to be done, can do it myself not to disturb.

ghugo83:
Had accidentally called the commits [alpakatest], while here it is [alpaka]

src/alpaka/AlpakaCore/AtomicPairCounter.h
```diff
@@ -0,0 +1,53 @@
+#ifndef HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
+#define HeterogeneousCore_CUDAUtilities_interface_AtomicPairCounter_h
+
+#include <cstdint>
+
+namespace cms {
+  namespace Alpaka {
```
makortel, line 7: To be compatible with CMSSW convention
```suggestion
  namespace alpaka {
```

makortel:
> NB: Also let me know if there is some formatting to be done, can do it myself not to disturb.

If you have LLVM 10, you could do `make format CLANG_FORMAT=<name or path to clang-format if not clang-format-10>` (or `make format_alpaka ...` to format only `src/alpaka`).

src/alpaka/test/alpaka/AtomicPairCounter_t.cc
```diff
@@ -0,0 +1,138 @@
...
+  Vec elementsPerThread(Vec::all(1));
+  Vec threadsPerBlock(Vec::all(512));
+  const Vec blocksPerGrid(Vec::all(2000));
+#if defined ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED || ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED || \
+  ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED || ALPAKA_ACC_CPU_BT_OMP4_ENABLED
+  // on the GPU, run with 32 threads in parallel per block, each looking at a single element
+  // on the CPU, run serially with a single thread per block, over 32 elements
+  std::swap(threadsPerBlock, elementsPerThread);
+#endif
+  const WorkDiv workDiv(blocksPerGrid, threadsPerBlock, elementsPerThread);
```
makortel, line 100: (possibly beyond this PR)

I'm wondering if would make sense already now to abstract this with something along (pseudocode)
```cpp
WorkDiv make_workdiv(int blocksPerGrid, int threadsOrElements) {
#if defined ALPAJA_ACC_CPU ...
  return WorkDiv(blocksPerGrid, 1, threadsOrElements);
#else
  return WorkDiv(blocksPerGrid, threadsOrElements, 1);
#endif
}
```
and call here
```cpp
  const WorkDiv workDiv = cms::alpaka::make_workdiv(2000, 512);
```

src/alpaka/test/alpaka/AtomicPairCounter_t.cc
```diff
@@ -0,0 +1,138 @@
...
+  template <typename T_Acc>
+  ALPAKA_FN_ACC void operator()(const T_Acc& acc, cms::Alpaka::AtomicPairCounter *dc, uint32_t *ind, uint32_t *cont, uint32_t n) const {
+    // Global thread index in grid
+    const uint32_t threadIdxGlobal(alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u]);
+    const uint32_t threadDimension(alpaka::workdiv::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc)[0u]);
+
+    // Global element index (obviously relevant for CPU only, for GPU, i = threadIndexGlobal only)
+    const uint32_t firstElementIdxGlobal = threadIdxGlobal * threadDimension;
+    const uint32_t endElementIdxGlobalUncut = firstElementIdxGlobal + threadDimension;
+    const uint32_t endElementIdxGlobal = std::min(endElementIdxGlobalUncut, n);
```
makortel, line 20: (possibly beyond this PR)

I'm wondering if it would make sense already now to abstract these with something along
```cpp
template <typename T, typename T_Acc>
std::pair<T, T> make_global_index_range_max(const T_Acc& acc, T max) {
  // Global thread index in grid
  const T threadIdxGlobal(alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u]);
  const T threadDimension(alpaka::workdiv::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc)[0u]);

  // Global element index (obviously relevant for CPU only, for GPU, i = threadIndexGlobal only)
  const T firstElementIdxGlobal = threadIdxGlobal * threadDimension;
  const T endElementIdxGlobalUncut = firstElementIdxGlobal + threadDimension;
  const T endElementIdxGlobal = std::min(endElementIdxGlobalUncut, max);

  return {firstElementIdxGlobal, endElementIdxGlobal};
}
```
and call here
```cpp
  const auto [firstElementIdxGlobal, endElementIdxGlobal] = cms::alpaka::make_global_index_range_max(acc, n);
```

(or instead of `pair` return a wrapper that provides `begin()` and `end()` so the returned object could be used with range-for)

src/alpaka/AlpakaCore/AtomicPairCounter.h
ghugo83: Ha yes, the issue is that when I rename cms::Alpaka to cms::alpaka, any function from alpaka library (for example here, alpaka::atomic::atomicOp) is looked for within cms::alpaka.
Hmm maybe not ideal to have the same name here?

For Kokkos, an issue with cms::kokkos does not appear, because the namespace from Kokkos has an upperletter: Kokkos::atomic_fetch_add

src/alpaka/test/alpaka/AtomicPairCounter_t.cc
ghugo83: @makortel Yes absolutely, would make total sense.
Actually was thinking doing it in another PR.

src/alpaka/test/alpaka/AtomicPairCounter_t.cc
ghugo83: Yep absolutely

src/alpaka/test/alpaka/AtomicPairCounter_t.cc
makortel: Another PR is fine.

src/alpaka/AlpakaCore/AtomicPairCounter.h
makortel: > Ha yes, the issue is that when I rename cms::Alpaka to cms::alpaka, any function from alpaka library (for example here, alpaka::atomic::atomicOp) is looked for within cms::alpaka.
> Hmm maybe not ideal to have the same name here?

Fair point. I'd expect qualifying the Alpaka function calls with leading `::` along `::alpaka::atomic::atomic::atomicOp` to work though.

ghugo83: Hmm yes there must be a way around, but then whenever we are within cms::alpaka and we call a function from alpaka library, we need to use it, which makes it not very elegant, no?

makortel: I agree having to use leading `::` is slightly annoying, but so would be to remember that one of our namespaces starts with capital letter, and, conversely, that the name `cms::Alpaka` is a namespace and not a class. Thinking about a possible use in CMSSW, I'd expect `cms::[aA]lpaka` to be used more than `::alpaka` inside functions/classes in `cms::[aA]lpaka`, and therefore I'd value interface consistency over slight annoyance in utility function/class implementations.

Written that, this project is all about prototyping and certain shortcuts are allowed, and the choice of the namespace name is not that relevant for the question "what would it take to use Alpaka in CMSSW" (even if it is part of it). I could accept the capital version now, and it would be straightforward to change it later, but it would be easier to do it now than later.

ghugo83: @makortel Ok thanks for the explanation, yes will change it to `cms::alpaka` and just use `::alpaka` for alpaka library calls then.

ghugo83: Ok I have just pushed sth at https://github.com/cms-patatrack/pixeltrack-standalone/pull/160/commits/8e026919b02f298c1392c5d6396e1acd330a2aef to get rid of the cms::Alpaka namespace, and follow CMSSW conventions.
In the end, I found it nice to simply rename `cms::Alpaka` to describe better what was behind the "Alpaka" anyway: `cms::Alpaka` namespace was only used in `AlpaCore` for helper functions, hence I have renamed `cms::Alpaka` to `cms::alpaka_helpers`.
This is in the same spirit as other 'home-made' alpaka namespaces we already have for other purposes: `alpaka_common`, `alpaka_cuda_async`, etc.
Let me know if you are ok with this solution, or I should rather change the alpaka library namespace.

makortel: The namespaces in `AlpakaCore/alpakaConfig.h` special and should not be used outside explicitly (with the exception of `alpaka_common` that appears to be useful in #160). The macro `ALPAKA_ACCELERATOR_NAMESPACE` should be used instead, and it gets replaced with the namespace corresponding the backend for which the translation unit is being compiled for.

That reminds me (sorry for not before) that for `sycltest` program we had a similar discussion on the namespaces in #102 (ok, I see the discussion is not in the PR but I'm pretty sure I discussed about that with @fwyzard; found it in our private chat). There we decided to go with `cms::sycltools` and figure out better name later (if needed). So how about `cms::alpakatools`? I could accept `alpaka_helpers" as well for now, but I wanted ask about `cms::alpakatools` first.

ghugo83: Ok sure, will put cms::alpakatools rather than cms::alpaka_helpers if you prefer, will make it more consistent with the sycl side

ghugo83: @makortel Just renamed from cms::alpaka_helpers to cms::alpakatools as you suggested (+ few other cleanings, templating) (I put the changes in https://github.com/cms-patatrack/pixeltrack-standalone/pull/160 so that all the code is directly properly modified).

makortel:
@ghugo83 

> Just renamed from `cms::alpaka_helpers` to `cms::alpakatools` as you suggested (+ few other cleanings, templating) (I put the changes in #160 so that all the latest code is directly properly modified).

Do you prefer to continue in #160 then? (if yes, please close this PR, if no, please propagate the namespace change here)

ghugo83:
@makortel As you want, this PR and #160 add different (but dependent) features. The namespace point is fix on the fly everywhere after #160 is merged.
#160 is built on top of this branch.
One can either:
- (1) merge this PR, then merge #160.
- (2) or I close this PR and one can just merge #160.
Do you prefer (2) ?

makortel:
On a second thought let's go with (1).

PR #152, opened 2021-01-06
[kokkos] Add work-in-progress support for HIP backend

makortel:
This PR adds support for HIP backend to `kokkostest` and `kokkos` programs. The `kokkostest` is fully functional (on Vega 10 XL/XT at least), while `kokkos` does not run yet. There is some issue with `assert()` calls (setup working with CUDA does not work with HIP, and vice versa), and there is some runtime problem in CA pixel tracking that I have not managed to figure out yet. Nevertheless, I think there is value on getting this development in because of the impact in Makefiles.

Note that there is a patch for Eigen (more `EIGEN_DEVICE_FUNC`), which strictly speaking means that this PR implies `make distclean` or `rm -fR external/eigen`.

PR #161, opened 2021-01-14
[hip] Add work-in-progress version of hip program

makortel:
This is essentially a copy of `cuda` program with
* copying already-hipified code and build rules from `hiptest`
* running `hipify-perl` with `hipconvertinplace-perl.sh` for the rest
* fixing compilation errors and warnings
* rename `cms::cuda*` namespaces to `cms::hip*`
* generalize warp size from 32 to `warpSize`

There are still several unit tests that fail to run in addition of `hip` program itself (similar to https://github.com/cms-patatrack/pixeltrack-standalone/pull/152). The runtime errors do not seem to be trivial to fix, so I think it is more beneficial to include the (incomplete) developments here than letting the ~~rot~~ sit in my private branch.


makortel:
For future reference, the failures I see are
```
$ ./hip --maxEvents 1
Found 1 devices
Processing 1 events, of which 1 concurrently, with 1 threads.
Clusters/Hits Overflow 49152 >= 49152
:0:rocdevice.cpp            :2303: 1460541978325 us: Device::callbackQueue aborting with status: 0x29
Aborted (core dumped)

$ /test/hip/testEigenGPUNoFit
TEST EIGENVALUES
TEST INVERSE 3x3
TEST INVERSE 4x4
TEST INVERSE 5x5
# hangs at this point

$ ./test/hip/prefixScan_t 
warp level
block level
:0:rocdevice.cpp            :2303: 1460699916491 us: Device::callbackQueue aborting with status: 0x100f
Aborted (core dumped)

$ ./test/hip/eigenSoA_t
0.808429
:0:rocdevice.cpp            :2303: 1460714637372 us: Device::callbackQueue aborting with status: 0x100f
Aborted (core dumped)

$ ./test/hip/radixSort_t
using Global memory
Will sort 81920 signed 'ints' of size 1 using 1 significant bytes
lauch for 81755
done for 81755
/home/matti/pixeltrack-standalone/src/hip/CUDACore/radixSort.h:179: : Device-side assertion `' failed.
...
Aborted (core dumped)

$ ./test/hip/gpuVertexFinder_t 
v,t size 26 301
M eps, pset 2 0.1 0
nt,nv 301 0,0
nt,nv 301 19,19
after fit nv, min max chi2 19 0.526189 15.6285
before splitting nv, min max chi2 19 0.526189 13.9668
after split 21
# hangs at this point
```

PR #162, opened 2021-01-15
[cudauvm] Migrate rechit to use managed memory

kakwok:
Part of #43


makortel:
Looks otherwise good, but could you also update the README by adding the new macro to the table in https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/README.md#cudauvm ?

kakwok:
Attaching relative throughput before and after migration
image

PR #166, opened 2021-01-29
[kokkos] Rewrite Hist::finalize() with Teamscan

PointKernel:
I tried to fuse multiple kernels into one since `finalize()` is now a device function but got a sequence of different errors. Will do it next week.

### DIFF
--- a/src/kokkos/test/kokkos/OneHistoContainer_t.cc
+++ b/src/kokkos/test/kokkos/OneHistoContainer_t.cc
@@ -61,7 +61,6 @@ void go() {
 
     Kokkos::parallel_for(
         "count",
-
 // XXX Team prefix scan used in Hist::finalize() requires power-of-two blockDim
 // Kokkos::AUTO() in CUDA backend will somehow use blockDim = 96 thus fail the execution
 #ifdef KOKKOS_BACKEND_CUDA
--- a/src/kokkos/test/kokkos/OneHistoContainer_t.cc
+++ b/src/kokkos/test/kokkos/OneHistoContainer_t.cc
@@ -46,60 +46,60 @@ void go() {
 
     using TeamHist = cms::kokkos::HistoContainer<T, NBINS, N, S, uint16_t>;
 
-    Kokkos::View<TeamHist[1], KokkosExecSpace> histo_d("histo_d");
+    Kokkos::View<TeamHist*, KokkosExecSpace> histo_d("histo_d", 1);
     auto histo_h = Kokkos::create_mirror_view(histo_d);
 
     Kokkos::parallel_for(
         "set_zero",
         Kokkos::RangePolicy<KokkosExecSpace>(KokkosExecSpace(), 0, TeamHist::totbins()),
-        KOKKOS_LAMBDA(const int& i) { histo_d().off[i] = 0; });
+        KOKKOS_LAMBDA(const int& i) { histo_d(0).off[i] = 0; });
 
     Kokkos::parallel_for(
         "set_zero_bin",
         Kokkos::RangePolicy<KokkosExecSpace>(KokkosExecSpace(), 0, TeamHist::capacity()),
-        KOKKOS_LAMBDA(const int& i) { histo_d().bins[i] = 0; });
+        KOKKOS_LAMBDA(const int& i) { histo_d(0).bins[i] = 0; });
 
     Kokkos::parallel_for(
         "count",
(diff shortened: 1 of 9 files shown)


src/kokkos/test/kokkos/OneHistoContainer_t.cc
```diff
@@ -46,7 +46,7 @@ void go() {
 
     using TeamHist = cms::kokkos::HistoContainer<T, NBINS, N, S, uint16_t>;
 
-    Kokkos::View<TeamHist, KokkosExecSpace> histo_d("histo_d");
+    Kokkos::View<TeamHist[1], KokkosExecSpace> histo_d("histo_d");
```
makortel, line 49: Why was this change needed?

makortel (commented):
Thanks, looks good. I have only one question.

makortel:
`VertexFinder_t` unit test fails with
```
Running test /global/cscratch1/sd/matti/cce/prtest/pixeltrack-standalone/test/kokkos/VertexFinder_t.device
v,t size 26 301
M eps, pset 2 0.1 0
nt,nv 301 0,0
:0: : block: [0,0,0], thread: [0,32,0] Assertion `Cuda::cuda_intra_block_scan requires power-of-two blockDim` failed.
:0: : block: [0,0,0], thread: [0,33,0] Assertion `Cuda::cuda_intra_block_scan requires power-of-two blockDim` failed.
:0: : block: [0,0,0], thread: [0,34,0] Assertion `Cuda::cuda_intra_block_scan requires power-of-two blockDim` failed.
...
  what():  cudaStreamSynchronize(m_stream) error( cudaErrorAssert): device-side assert triggered .../pixeltrack-standalone/external/kokkos/source/core/src/Cuda/Kokkos_Cuda_Instance.cpp:314
```

src/kokkos/test/kokkos/OneHistoContainer_t.cc
PointKernel: In `Hist::finalize()`, each hist view has to take an argument, e.g. `update += histo(leagueRank).off[i];`. This expression cannot be applied to "scalar" view. For example,
```c++
Kokkos::View<TeamHist, KokkosExecSpace> histo_d("histo_d");
Kokkos:: parallel_scan( ...
  update += histo(0).off[i]; // this will not work, result in compile time error
  update += histo().off[i]; // this will work, "scalar" view takes no argument
);
```
Since the original `forEachInWindow` function works over a scalar view only and the view in `finalize()` has to take an argument, I just make a minimal change to make both of them run. Let me know if a thorough solution to this is required.

PointKernel:
> `VertexFinder_t` unit test fails with
> 
> ```
> Running test .../pixeltrack-standalone/test/kokkos/VertexFinder_t.device
> v,t size 26 301
> M eps, pset 2 0.1 0
> nt,nv 301 0,0
> :0: : block: [0,0,0], thread: [0,32,0] Assertion `Cuda::cuda_intra_block_scan requires power-of-two blockDim` failed.
> :0: : block: [0,0,0], thread: [0,33,0] Assertion `Cuda::cuda_intra_block_scan requires power-of-two blockDim` failed.
> :0: : block: [0,0,0], thread: [0,34,0] Assertion `Cuda::cuda_intra_block_scan requires power-of-two blockDim` failed.
> ...
> terminate called after throwing an instance of 'std::runtime_error'
>   what():  cudaStreamSynchronize(m_stream) error( cudaErrorAssert): device-side assert triggered .../pixeltrack-standalone/external/kokkos/source/core/src/Cuda/Kokkos_Cuda_Instance.cpp:314
> Traceback functionality not available
> ```

src/kokkos/test/kokkos/OneHistoContainer_t.cc
makortel: Thanks. Just to check that I understood, the need to be able to call `HistoContainer::finalize()` for an array of `HistoContainer` necessitates the indexing via `histo(leagueRank)`, and therefore all calls need to look like arrays (i.e. mimic a scalar with 1-element array). 

Then I'm wondering why
https://github.com/cms-patatrack/pixeltrack-standalone/blob/95d56aad8feb461165fe7563bedc564ac3c4f816/src/kokkos/KokkosCore/HistoContainer.h#L349
takes a scalar and not `Kokkos::View<Histo*, ExecSpace>`?

PointKernel: Oh no, that's a mistake. Thanks for pointing this out.

PR #175, opened 2021-02-22
CUDA: use the stream-ordered memory allocations if the caching allocator is disabled

fwyzard:
If the caching allocator is disabled and CUDA version is 11.2 or greater is detected, device allocations and deallocations will use the stream-ordered CUDA functions `cudaMallocAsync` and `cudaFreeAsync`.

Their use can be disabled explicitly at compile time setting also the `CUDA_DISABLE_ASYNC_ALLOCATOR` preprocessor symbol.

### DIFF
--- a/src/cuda/CUDACore/allocate_device.cc
+++ b/src/cuda/CUDACore/allocate_device.cc
@@ -1,4 +1,7 @@
+#include <cassert>
 #include <limits>
+
+#include <cuda_runtime.h>
 
 #include "CUDACore/ScopedSetDevice.h"
 #include "CUDACore/allocate_device.h"
@@ -14,12 +17,17 @@ namespace {
 namespace cms::cuda {
   void *allocate_device(int dev, size_t nbytes, cudaStream_t stream) {
     void *ptr = nullptr;
-    if constexpr (allocator::useCaching) {
+    if constexpr (allocator::policy == allocator::Policy::Caching) {
       if (nbytes > maxAllocationSize) {
         throw std::runtime_error("Tried to allocate " + std::to_string(nbytes) +
                                  " bytes, but the allocator maximum is " + std::to_string(maxAllocationSize));
       }
       cudaCheck(allocator::getCachingDeviceAllocator().DeviceAllocate(dev, &ptr, nbytes, stream));
+#if CUDA_VERSION >= 11020
+    } else if constexpr (allocator::policy == allocator::Policy::Asynchronous) {
+      ScopedSetDevice setDeviceForThisScope(dev);
+      cudaCheck(cudaMallocAsync(&ptr, nbytes, stream));
+#endif
     } else {
       ScopedSetDevice setDeviceForThisScope(dev);
       cudaCheck(cudaMalloc(&ptr, nbytes));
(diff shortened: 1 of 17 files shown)


fwyzard:
Here is a comparison of the performance using
  - `cudaMalloc`/`cudaFree`
  - `cudaMallocAsync`/`cudaFreeAsync`
  - the CUB-based caching allocator

obtained running `./cuda --numberOfThreads 10 --maxEvents 10000` on a T4 with CUDA 11.2 Update 1:

allocator | throughput [ev/s]
--- | ---
`cudaMalloc` | 120.4 ± 0.5
`cudaMallocAsync` | 522.3 ± 3.0
caching allocator | 922.2 ± 3.1

src/cuda/CUDACore/allocate_device.cc
```diff
@@ -22,17 +22,25 @@ namespace cms::cuda {
       cudaCheck(allocator::getCachingDeviceAllocator().DeviceAllocate(dev, &ptr, nbytes, stream));
     } else {
       ScopedSetDevice setDeviceForThisScope(dev);
+#if CUDA_VERSION >= 11020
+      cudaCheck(cudaMallocAsync(&ptr, nbytes, stream));
+#else
```
makortel, line 27: I would like to be able to force `cudaMalloc()` also on CUDA >= 11.2. How about an explicit macro, e.g. something along `CUDA_DISABLE_MALLOC_ASYNC` (possibly combined with the version check if the macro would disable the "async" version)?

fwyzard: I've added `CUDA_DISABLE_ASYNC_ALLOCATOR` to follow the same naming scheme as `CUDA_DISABLE_CACHING_ALLOCATOR`

src/cuda/CUDACore/allocate_device.cc
```diff
@@ -14,22 +14,28 @@ namespace {
...
-    if constexpr (allocator::useCaching) {
+    if constexpr (allocator::policy == allocator::Policy::Caching) {
       if (nbytes > maxAllocationSize) {
         throw std::runtime_error("Tried to allocate " + std::to_string(nbytes) +
                                  " bytes, but the allocator maximum is " + std::to_string(maxAllocationSize));
       }
       cudaCheck(allocator::getCachingDeviceAllocator().DeviceAllocate(dev, &ptr, nbytes, stream));
+    } else if constexpr (allocator::policy == allocator::Policy::Asynchronous) {
+      ScopedSetDevice setDeviceForThisScope(dev);
+      cudaCheck(cudaMallocAsync(&ptr, nbytes, stream));
```
makortel, line 25: This fails to compile on CUDA 11.1, so needs to be protected with `#ifdef` as well (or the implementation moved to a templated function to allow `if constexpr` to avoid checking the discarded statement).

fwyzard: I don't understand - this *is* within an  `if constexpr` that should not be taken if CUDA < 11.2

makortel: Right, but
> Outside a template, a discarded statement is fully checked. `if constexpr` is not a substitute for the `#if` preprocessing directive

https://en.cppreference.com/w/cpp/language/if

fwyzard: Sigh.

makortel: Yeah...

fwyzard: So, what do you prefer ?

1. add a declaration for the new functions if the CUDA version is older than 11.2.0:
    ```c++
    #include <cuda_runtime.h>
    #if CUDA_VERSION < 11020
   // declare the stream ordered memory allocation functions, that are available only starting from CUDA 11.2.0
    __host__ cudaError_t cudaFreeAsync(void* ptr, cudaStream_t stream);
    __host__ cudaError_t cudaMallocAsync(void** ptr, size_t size, cudaStream_t stream);
    #endif  // CUDA_VERSION
    ```

2. add an `#if CUDA_VERSION >= 11020` around the use of `cudaMallocAsync`/`cudaFreeAsync`

3. add some fake templating to avoid checking inside the `if constexpr`

makortel: I don't have strong feeling. I'd probably go myself with 2, but I'd be fine with any of them (for the purposes of this repo).

fwyzard: I've added some `#if` ...

PR #176, opened 2021-03-02
Use consistently curl instead of wget

fwyzard:
This solves two potential problems:
  - running on a system where `wget` is not available, but `curl` is;
  - running behind a SOCKS5 proxy, that `curl` supports but `wget` does not.


fwyzard:
Note `-L` tells `curl` to follow redirects like HTTP 302 messages.

fwyzard:
Should we use `curl -s -S` to make curl silent, but still show errors ?

makortel:
> Should we use `curl -s -S` to make curl silent, but still show errors ?

I'd be fine with that.

fwyzard:
done

PR #177, opened 2021-03-02
Update Eigen to the upstream master branch

fwyzard:
Upate Eigen to the master branch, using the same version used by TensorFlow 2.4.1: commit [011e0db3](https://gitlab.com/libeigen/eigen/-/commit/011e0db31d1bed8b7f73662be6d57d9f30fa457a) from October 14th, 2020.

This do not require any more the use of the CMS-specific changes.


fwyzard:
It looks like this update incurs in a ~2% overall drop in throughput with respect to the older version.
This should be investigated before applying the change.

Makefile
```diff
@@ -436,9 +436,9 @@ $(TBB_LIB): $(TBB_BASE) $(TBB_LIBDIR)
 external_eigen: $(EIGEN_BASE)
 
 $(EIGEN_BASE):
-	git clone https://github.com/cms-externals/eigen-git-mirror $@
-	cd $@ && git checkout -b cms_branch b20a61c3a0dc9a79790cd258130a99b574662272
-	cd $@ && patch -p1 < ../../eigen_hip.patch
+	git clone -b master https://gitlab.com/libeigen/eigen.git $@
+	cd $@ && git reset --hard 011e0db31d1bed8b7f73662be6d57d9f30fa457a	# from TensorFlow 2.4.1
+	#cd $@ && patch -p1 < ../../eigen_hip.patch
```
makortel, line 441: This version appears to work with HIP as well, so the patch can be removed
```suggestion
```

fwyzard: :+1:

makortel:
I did not observe (either) any significant impact on performance on `cuda`, `cudacompat`, or `kokkos`.

PR #180, opened 2021-04-20
[alpaka] Alpaka port operational up to clusterizer

ghugo83:
- Add Digi, DigiErrors, Clusters AlpakaData formats.
- Add needed CondFormats.
- Add BeamSpot Producer.
- Add SiPixelFedCablingMapESProducer.
- Add SiPixelGainCalibrationForHLTESProducer.
- Completed plugin-SiPixelClusterizer, including SiPixelRawToCluster, SiPixelRawToClusterGPUKernel and gpuCalibPixel kernel.
- Add validation plugin (evth in place, can just un-comment the rest when more products become available).

**Compiles & runs smoothly on all events and for all backends (Alpaka 0.6.0)**:
- SERIAL
- TBB
- CUDA

**Possible to produce histograms with `--histogram`. Digis and clusters results are identical to legacy.
Also run smoothly with `--validation`.
Possible to run concurrent events with `--numberOfStreams 2` etc** : increase throughput as expected.

Present, temporary, perf on my local (tested on 1000 events):
- SERIAL: throughput ~71 +- 3 events/s with default Alpaka, **throughput ~119 +- 3 events/s** by deactivating the grid atomics at https://github.com/alpaka-group/alpaka/blob/0.6.1/include/alpaka/acc/AccCpuSerial.hpp#L63 (as noticed during tests. Grid atomics are not needed here anyway).
- TBB: **throughput ~89 +- 3 events/s** with TBB default number of threads. (Grid atomics not desactivated).
- CUDA: **throughput ~258 +- 3 events/s**

NB: Will squash all commits into 1 after corrections.

@makortel @fwyzard


Makefile
```diff
@@ -403,6 +403,15 @@ distclean: | clean
...
 	rm -fR data/*.tar.gz data/*.bin data/data_ok
+	
+alpakaclean:
+	rm -fR lib/alpaka obj/alpaka test/alpaka alpaka
+	
+cudaclean:
+	rm -fR lib/cuda obj/cuda test/cuda cuda
+	
+kokkosclean:
+	rm -fR lib/kokkos obj/kokkos test/kokkos kokkos
```
makortel, line 414: Is this level of precision for cleaning useful? (in my use cases cleaning everything has been sufficient)

If it is, this would be better done in a general way.

src/alpaka/AlpakaCore/alpaka/alpakaDevAcc.cc
```diff
@@ -0,0 +1,5 @@
+#include "AlpakaCore/alpakaDevAcc.h"
+
+namespace ALPAKA_ACCELERATOR_NAMESPACE {
+  const DevAcc1 device = alpaka::getDevByIdx<PltfAcc1>(0u);
```
makortel, line 4: I wonder why this is needed as a global object (I may find the reason while reading code)? Anyway, ok for now as it can be improved later.

src/alpaka/AlpakaCore/alpakaConfig.h
```diff
@@ -1,154 +1,7 @@
 #ifndef alpakaConfig_h_
 #define alpakaConfig_h_
 
-#include <alpaka/alpaka.hpp>
```
makortel, line 4: Why was it useful to split this file?

src/alpaka/AlpakaCore/alpakaMemoryHelper.h
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    template <typename TData>
+    auto allocDeviceBuf(const Extent& extent) {
+      return alpaka::allocBuf<TData, Idx>(ALPAKA_ACCELERATOR_NAMESPACE::device, extent);
+    }
+
+    template <typename TData>
+    auto createDeviceView(const TData* data, const Extent& extent) {
+      return alpaka::ViewPlainPtr<ALPAKA_ACCELERATOR_NAMESPACE::DevAcc1, TData, Dim1, Idx>(
```
makortel, line 29: Would
```suggestion
      return alpaka::ViewPlainPtr<ALPAKA_ACCELERATOR_NAMESPACE::DevAcc1, const TData, Dim1, Idx>(
```
work?

src/alpaka/AlpakaCore/alpakaMemoryHelper.h
```diff
@@ -0,0 +1,44 @@
...
+          const_cast<TData*>(data),
+          ALPAKA_ACCELERATOR_NAMESPACE::device,
+          extent);  // TO DO: Obviously aweful: why no view constructor inside alpaka library with a const TData* argument?
+    }
+
+    template <typename TData>
+    auto createDeviceView(TData* data, const Extent& extent) {
+      return alpaka::ViewPlainPtr<ALPAKA_ACCELERATOR_NAMESPACE::DevAcc1, TData, Dim1, Idx>(
```
makortel, line 37: It would be cleaner here if these functions would take the `device` as template argument, then the `ALPAKA_ACCELERATOR_NAMESPACE` would not be needed.

src/alpaka/AlpakaDataFormats/BeamSpotAlpaka.h
```diff
@@ -0,0 +1,29 @@
...
+#include <cstring>
+
+namespace ALPAKA_ACCELERATOR_NAMESPACE {
+
+  class BeamSpotAlpaka {
+  public:
+    BeamSpotAlpaka() = default;
+
+    BeamSpotAlpaka(BeamSpotPOD const* data, Queue& queue) : data_d{cms::alpakatools::allocDeviceBuf<BeamSpotPOD>(1u)} {
+      auto data_h{cms::alpakatools::createHostView<const BeamSpotPOD>(data, 1u)};
```
makortel, line 16: Does `createHostView()` create a copy, or only a non-owning "View" that can be passed to `alpaka::memcpy()` in lieu of the `data` pointer?

src/alpaka/AlpakaDataFormats/BeamSpotAlpaka.h
```diff
@@ -0,0 +1,29 @@
+#ifndef AlpakaDataFormats_BeamSpot_interface_BeamSpotAlpaka_h
+#define AlpakaDataFormats_BeamSpot_interface_BeamSpotAlpaka_h
+
+#include "AlpakaCore/alpakaCommon.h"
+#include "DataFormats/BeamSpotPOD.h"
+
+#include <cstring>
+
+namespace ALPAKA_ACCELERATOR_NAMESPACE {
```
makortel, line 9: This is more for further thought, I have an uneasy feeling about this that I'm unable to express in words. It's probably good enough for now.

If I read the template definitions correctly, as an alternative it would be possible to template the `BeamSpotAlpaka` over the Accelerator. (to repeat, I don't suggest to change this now, it is also an interesting to see how this approach pans out)

makortel (commented):
Reviewed up to BeamSpotAlpaka.

src/alpaka/AlpakaCore/alpaka/alpakaDevAcc.cc
makortel: It certainly makes some things simpler to avoid passing the device objects around. These may become easier with the `ScopedContext` etc where we can more easily hide these details without resorting to globals.

I would expect this global to require that a device for each accelerator actually exists, and that the program would not be able support multiple devices of the same type.

Makefile
ghugo83: Yes, find it nice to be able to clean for example src/alpaka, without having to recompile all the rest.
But yes will push a more generic way

src/alpaka/AlpakaCore/alpaka/alpakaDevAcc.cc
ghugo83: >These may become easier with the ScopedContext etc where we can more easily hide these details without resorting to globals.

Yes exactly.

src/alpaka/AlpakaCore/alpakaConfig.h
ghugo83: There is now a source file in `src/alpaka/AlpakaCore`, which implied having a `src/alpaka/AlpakaCore/alpaka` repo to be able to compile this with -DALPAKA_ACC_GPU_CUDA_ENABLED etc:
```
 #ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
#define ALPAKA_ACC_GPU_CUDA_ASYNC_BACKEND
namespace alpaka_cuda_async {
  using namespace alpaka_common;
  using Acc1 = alpaka::AccGpuCudaRt<Dim1, Extent>;
  using Acc2 = alpaka::AccGpuCudaRt<Dim2, Extent>;
  using Queue = alpaka::QueueCudaRtNonBlocking;
}  // namespace alpaka_cuda_async
```

src/alpaka/AlpakaCore/alpakaMemoryHelper.h
ghugo83: Yes I had done that at first, was passing host and device as parameters.
But then I found it very redundant on call site, as in a sense we already know that allocHostBuf / allocDeviceBuf will be allocating memory on host and device respectively. And since they are global object anyway, found it nicer to just call them inside the helper functions.

src/alpaka/AlpakaDataFormats/BeamSpotAlpaka.h
ghugo83: `createHostView()` creates a non-owning view.
One cannot passes the raw data pointer to `alpaka::memcpy()`.
Basically:
- the `alpakaHost/DeviceBuf `are owning (shared) pointers.
- with `alpaka::getPtrNative(myDeviceBuf)`, one gets a non-owning (raw) pointer.
- `alpakaHost/DeviceViews` are just wraps around raw pointers, allowing operations like `alpaka::memcpy()`, or also allowing to view a fraction of the owned data with a `alpakaSubview`.

src/alpaka/AlpakaCore/alpakaMemoryHelper.h
ghugo83: But indeed as you mentioned, some further work should be done inside the Framework, to just access evth with a ScopedContext or equivalent.

src/alpaka/AlpakaCore/alpaka/alpakaDevAcc.cc
ghugo83: This was actually an issue for me for some time, as in the examples, the host device were just passed around through 2 or 3 files.
In the implementation inside Kokkos, from what I understood, there is simply the use of a singleton pattern to access host/device. There was no such feature in Alpaka.
I eventually found it a nice workaround for now, without modifying the Acc classes inside Alpaka. But indeed, some more framework-related work will be needed to directly access these with a Context or equivalent.

src/alpaka/AlpakaCore/alpakaMemoryHelper.h
ghugo83: Ha yes thanks, that worked.

src/alpaka/AlpakaDataFormats/BeamSpotAlpaka.h
ghugo83: Ok cool.
I have fully finished the RecHits with Alpaka (evth runs smoothly and results are 100% ok now in validation), and considerably advanced the pixel tracks/vertices (hoping to finish by the end of the weekkkkkk), it does not seem to have been a major issue.
In general the namespace `KOKKOS_NAMESPACE` and` ALPAKA_ACCELERATOR_NAMESPACE` are perfectly matching (not here though indeed).

src/alpaka/AlpakaCore/alpaka/alpakaDevAcc.cc
makortel: > In the implementation inside Kokkos, from what I understood, there is simply the use of a singleton pattern to access host/device

Right, because Kokkos itself provides those constraints (at most one device backend can be enabled). For the goal of supporting multiple device backends in a single build that constraint causes quite some headache.

Makefile
makortel: That would be along
```Makefile
define CLEAN_template
clean_$(1):
	rm -fR lib/$(1) obj/$(1) test/$(1) $(1)
endef
$(foreach target,$(TARGETS_ALL),$(eval $(call CLEAN_template,$(target))))
```

src/alpaka/AlpakaCore/alpakaConfig.h
makortel: Ok, I don't follow though how `src/alpaka/AlpakaCore/alpaka/alpakaDevAcc.cc` would not work with the monolithic `alpakaConfig.h`.

src/alpaka/AlpakaDataFormats/BeamSpotAlpaka.h
```diff
@@ -0,0 +1,29 @@
...
+
+  class BeamSpotAlpaka {
+  public:
+    BeamSpotAlpaka() = default;
+
+    BeamSpotAlpaka(BeamSpotPOD const* data, Queue& queue) : data_d{cms::alpakatools::allocDeviceBuf<BeamSpotPOD>(1u)} {
+      auto data_h{cms::alpakatools::createHostView<const BeamSpotPOD>(data, 1u)};
+
+      alpaka::memcpy(queue, data_d, data_h, 1u);
```
makortel, line 18: I suppose the `data_h` does not need to be alive when the asynchronous operation launched by `alpaka::memcpy` completes?

src/alpaka/CondFormats/SiPixelFedCablingMapGPUWrapper.h
```diff
@@ -0,0 +1,31 @@
...
+    using CablingMapDeviceBuf = AlpakaDeviceBuf<SiPixelFedCablingMapGPU>;
+
+    explicit SiPixelFedCablingMapGPUWrapper(CablingMapDeviceBuf cablingMap, bool quality)
+        : cablingMapDevice_{std::move(cablingMap)}, hasQuality_{quality} {}
+    ~SiPixelFedCablingMapGPUWrapper() = default;
+
+    bool hasQuality() const { return hasQuality_; }
+
+    //const CablingMapDeviceBuf& cablingMap() const { return cablingMapDevice_; }
+    // TO DO: what about returning a view only?
```
makortel, line 21: Could you elaborate this TODO comment?

src/alpaka/bin/main.cc
```diff
@@ -123,7 +148,9 @@ int main(int argc, char** argv) {
             << numberOfThreads << " threads." << std::endl;
 
   // Initialize tasks scheduler (thread pool)
-  tbb::task_scheduler_init tsi(numberOfThreads);
+  // tbb::task_scheduler_init tsi(numberOfThreads);
+  // TO DO: Warning: this will not work, need to be included in header.
+  // See external/alpaka/include/alpaka/kernel/TaskKernelCpuTbbBlocks.hpp
```
makortel, line 153: I took a quick look of this file and it is not clear to me what the problem is.

makortel (commented):
Reviewed up to SiPixelGainCalibrationForHLTESProducer

Makefile
ghugo83: Ok I put your function there, thanks

src/alpaka/AlpakaCore/alpakaConfig.h
ghugo83: For each package:
- Source placed in alpaka subdirectory (`alpaka/AlpakaCore/alpaka`) is compiled with -DALPAKA_ACC_GPU_CUDA_ENABLED, etc, for each accelerator type.
- Source directly placed in the package (`alpaka/AlpakaCore/`) is not compiled for each backend (source common for all backends).

Here, `alpakaDevAcc.cc` needs to be placed in the alpaka subdirectly (because it is compiled for each backend).
Instead, alpakaDevHost.cc cannot be placed in the alpaka subdirectory: would not make sense, as it is common for all backends, and actually the linker complains about it. 
So I put `alpakaDevHost.cc` in `src/alpaka/AlpakaCore`.
In my opinion, this makes sense: all common code is in `src/alpaka/AlpakaCore`, and all accelerator-dependent code is in `src/alpaka/AlpakaCore/alpaka`.

Both `alpakaDevHost.cc` and `alpakaDevAcc.cc` need to include `src/alpaka/AlpakaCore/alpakaConfig.h`.
If `src/alpaka/AlpakaCore/alpakaConfig.h` stays as it is (one file), the compilation of `alpakaDevHost.cc` does not work, because there is no -DALPAKA_ACC_GPU_CUDA_ENABLED etc during compilation (because `alpakaDevHost.cc` is not placed in alpaka subdirectory). 

Hence the split of `src/alpaka/AlpakaCore/alpakaConfig.h` into common utilities  (`alpakaConfigCommon`) and accelerator-dependent utilities ( `alpakaConfigAcc`): `alpakaDevHost` only needs, and only includes, `alpakaConfigCommon`.

Before adding host/device, there was no such issue, because there was not source file in entire `src/alpaka/AlpakaCore`: the headers (including the acc-dependents parts) were directly included from `src/alpaka/my-plugin/alpaka` (where all the alpaka-dependent code is placed).

src/alpaka/AlpakaDataFormats/BeamSpotAlpaka.h
ghugo83: Yes I am not sure about this case. 
Usually `alpaka::wait(queue);` is needed at end of scope, because otherwise the Queue object / host & device buffers objects and their data would be destructed.

Here it is not obvious, because the `queue` is not destructed at end of scope, `data_d` owns the device data for entire`BeamSpotAlpaka` lifetime, and data is not owned by `data_h `anyway (it is just a view). 

So it likely works anyway, but I presume in client code, if the `BeamSpotAlpaka` object is destructed before a `alpaka::wait(queue);` we run into trouble.

Added the `alpaka::wait(queue);`  for safety. If anything, studying potential removals of  `alpaka::wait(queue);` will be completely straightforwards, will dealing with the segfaults turns out way more complicated.

src/alpaka/CondFormats/SiPixelFedCablingMapGPUWrapper.h
ghugo83: At first I had a version which was returning a DeviceBuf. I didnt like it, as the client code did not need to own the data, so I just changed it to 
`const SiPixelFedCablingMapGPU* cablingMap() const { return alpaka::getPtrNative(cablingMapDevice_); }`
Which return a non-owning, raw pointer.

i just forgot to remove the to do comment.

src/alpaka/bin/main.cc
ghugo83: I could not change the number of threads in the TBB pool by using `tbb::task_scheduler_init tsi(numberOfThreads);` in main (had no effect).
Instead, by adding a `tbb::task_scheduler_init tsi(numberOfThreads);` in external/alpaka/include/alpaka/kernel/TaskKernelCpuTbbBlocks.hpp , I could control the number of threads in the TBB pool.

makortel: Thanks for the explanation. This needs to be understood better (because in CMSSW we must have a `task_scheduler_init` , well `global_control` nodaways, done in `main()` and tasks from Alpaka should go into the same thread pool), but can be done after this PR.

src/alpaka/AlpakaCore/alpakaConfig.h
makortel: Thanks for the explanation

makortel:
Merging now even if the review is incomplete to facilitate further work. Review comments should eventually be summarized in a separate issue.

ghugo83:
>Merging now even if the review is incomplete to facilitate further work. Review comments should eventually be summarized in a separate issue.
Haaa ok.
Let me know for the next one, because I will squash the commits just before the merge, will make the history cleaner.

fwyzard:
If squashing into a single commit is OK for you, it can be done quasi-automatically when merging (i.e. `Squash and merge` instead of `Merge pull request`).
If you prefer to split it into 4 or 5 commits (for example, following your suggestion) you can do that already !

ghugo83:
Yes ok will just squash the commits now for the next PR, not a big deal if there are few extra commits from PR review afterwards.

PR #181, opened 2021-04-30
Entire pixel track reconstruction chain with Alpaka source

ghugo83:
**Entire pixel track reconstruction chain with Alpaka source compiles and runs for all SERIAL, TBB and CUDA backends.
Results as produced with --histogram (and --validation) are same as legacy (within numerical precision).**

**Includes the entire application, all kernels**, AlpakaCore functionalities, all AlpakaData formats, transfers to host utilities, validation utilities.

First-try, by-hand, estimates of perf on my local in one-event-at-a-time mode (but the advancement here was mostly already to have an operational source code ;p):
- **Kokkos Serial:** ~21 +-1 events/s
- **Alpaka Serial:** ~27 +-1 events/s (Important NB: need to desactivate grid atomics at https://github.com/alpaka-group/alpaka/blob/0.6.1/include/alpaka/acc/AccCpuSerial.hpp#L63 ).
- **Alpaka TBB:** ~19 +-1 events/s (!! To be investigated. A big drain is that the grid atomics are not desactivated at https://github.com/alpaka-group/alpaka/blob/0.6.1/include/alpaka/acc/AccCpuTbbBlocks.hpp#L60)
- **Alpaka CUDA:** ~73 +-1 events/s (tested with no asserts, no caching allocator, no async in the sense that the async queues are created locally and not passed around by framework Producers, no `edm::EDProducerExternalWork`).

NB: For clarity of review, could have a look per package in this PR files list, or use the following progressive diffs:
- **Clusterizer:** https://github.com/cms-patatrack/pixeltrack-standalone/pull/180 
- **RecHits:** https://github.com/ghugo83/pixeltrack-standalone/compare/dev_trackreco_merge...ghugo83:rechit?expand=1 
- **Tracks:** https://github.com/ghugo83/pixeltrack-standalone/compare/rechit_PR...ghugo83:triplets 
- **Vertices:** https://github.com/ghugo83/pixeltrack-standalone/compare/triplets_PR...ghugo83:vertices
- **Final cleaning / review comments inclusion**: https://github.com/ghugo83/pixeltrack-standalone/compare/vertices...ghugo83:full_trackreco_chain
To avoid branches conflicts and partial changes, I will just push all review comments to this present PR. Then squah all commits into 1 when the review is completed.

@makortel @felicepantaleo @fwyzard


ghugo83:
Outlook: Mostly mini-framework related work.
- Need to further debug performance (same for Kokkos).
- Parallelization is intra-event only for now. Debugging / framework work is needed to handle multiple concurrent events mode.
_Inter-events_ parallelization run smoothly with Alpaka with only low number of events, and always crashes with Kokkos.
- Need to add framework-level facilities for choice of device. 
- Synchronization-related framework work: Already added local Alpaka asynchronous queues in the code (equivalent to `cudaStream_t` for CUDA backend), so the work should really be mostly framework-oriented, by simply adding an equivalent to edm::EDProducerExternalWork / cms::cuda::ScopedContextAcquire compatible with Alpaka. Same for Kokkos.
- CUDA version uses a caching allocator based on CUB to improve performance, no equivalent in Alpaka for now.

fwyzard:
This doesn't seem to work for me:
```bash
$ CUDA_VISIBLE_DEVICES=0 ./alpaka --cuda --maxEvents 1
Processing 1 events, of which 1 concurrently, with 1 threads.

(sits there forever...)
```

Built with
  - gcc 8.3.1
  - cuda 11.2
  - alpaka 0.6.0

Running on
  - 2x AMD EPYC 7502
  - 1x NVIDIA Tesla T4

The code as of #180 does work with the same setup.
Any suggestion of what may be going wrong ?

fwyzard:
> Ok.
> 
>     * Where does it hang?

According to `cuda-gdb`, almos always in in
```
57                      __syncthreads();

#0  void alpaka_cuda_async::GPUCACell::find_ntuplets<alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> >(alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> const&, alpaka_cuda_async::TrackingRecHit2DSOAView const&, alpaka_cuda_async::GPUCACell*, cms::alpakatools::SimpleVector<cms::alpakatools::VecArray<unsigned short, 48> >&, cms::alpakatools::HistoContainer<unsigned int, 32768u, 163840u, 32u, unsigned short, 1u>&, cms::alpakatools::AtomicPairCounter&, trackQuality::Quality*, cms::alpakatools::VecArray<unsigned int, 6>&, unsigned int, bool) const ()
    at /data/user/fwyzard/pixeltrack-standalone/external/alpaka/include/alpaka/block/sync/BlockSyncUniformCudaHipBuiltIn.hpp:57 in _ZN6alpaka6traits16SyncBlockThreadsINS_30BlockSyncUniformCudaHipBuiltInEvE16syncBlockThreadsERKS2_ inlined from BlockSyncUniformCudaHipBuiltIn.hpp:48
#1  0x000000001841e580 in void alpaka_cuda_async::GPUCACell::find_ntuplets<alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> >(alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> const&, alpaka_cuda_async::TrackingRecHit2DSOAView const&, alpaka_cuda_async::GPUCACell*, cms::alpakatools::SimpleVector<cms::alpakatools::VecArray<unsigned short, 48> >&, cms::alpakatools::HistoContainer<unsigned int, 32768u, 163840u, 32u, unsigned short, 1u>&, cms::alpakatools::AtomicPairCounter&, trackQuality::Quality*, cms::alpakatools::VecArray<unsigned int, 6>&, unsigned int, bool) const () at /data/user/fwyzard/pixeltrack-standalone/src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h:356
#2  0x000000001841e580 in void alpaka_cuda_async::GPUCACell::find_ntuplets<alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> >(alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> const&, alpaka_cuda_async::TrackingRecHit2DSOAView const&, alpaka_cuda_async::GPUCACell*, cms::alpakatools::SimpleVector<cms::alpakatools::VecArray<unsigned short, 48> >&, cms::alpakatools::HistoContainer<unsigned int, 32768u, 163840u, 32u, unsigned short, 1u>&, cms::alpakatools::AtomicPairCounter&, trackQuality::Quality*, cms::alpakatools::VecArray<unsigned int, 6>&, unsigned int, bool) const () at /data/user/fwyzard/pixeltrack-standalone/src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h:356
#3  0x000000001841e580 in void alpaka_cuda_async::GPUCACell::find_ntuplets<alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> >(alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int> const&, alpaka_cuda_async::TrackingRecHit2DSOAView const&, alpaka_cuda_async::GPUCACell*, cms::alpakatools::SimpleVector<cms::alpakatools::VecArray<unsigned short, 48> >&, cms::alpakatools::HistoContainer<unsigned int, 32768u, 163840u, 32u, unsigned short, 1u>&, cms::alpakatools::AtomicPairCounter&, trackQuality::Quality*, cms::alpakatools::VecArray<unsigned int, 6>&, unsigned int, bool) const () at /data/user/fwyzard/pixeltrack-standalone/src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h:356
#4  0x0000000018464320 in void alpaka::uniform_cuda_hip::detail::uniformCudaHipKernel<alpaka::AccGpuCudaRt<std::integral_constant<unsigned long, 1ul>, unsigned int>, std::integral_constant<unsigned long, 1ul>, unsigned int, alpaka_cuda_async::kernel_find_ntuplets, alpaka_cuda_async::TrackingRecHit2DSOAView const*, alpaka_cuda_async::GPUCACell*, unsigned int*, cms::alpakatools::SimpleVector<cms::alpakatools::VecArray<unsigned short, 48> >*, cms::alpakatools::HistoContainer<unsigned int, 32768u, 163840u, 32u, unsigned short, 1u>*, cms::alpakatools::AtomicPairCounter*, trackQuality::Quality*, unsigned int>(alpaka::Vec<std::integral_constant<unsigned long, 1ul>, unsigned int>, alpaka_cuda_async::kernel_find_ntuplets, alpaka_cuda_async::TrackingRecHit2DSOAView const*, alpaka_cuda_async::GPUCACell*, unsigned int*, cms::alpakatools::SimpleVector<cms::alpakatools::VecArray<unsigned short, 48> >*, cms::alpakatools::HistoContainer<unsigned int, 32768u, 163840u, 32u, unsigned short, 1u>*, cms::alpakatools::AtomicPairCounter*, trackQuality::Quality*, unsigned int)
   <<<(5376,1,1),(64,1,1)>>> () at /data/user/fwyzard/pixeltrack-standalone/src/alpaka/plugin-PixelTriplets/alpaka/CAHitNtupletGeneratorKernelsImpl.h:328 in _ZNK3cms11alpakatools8VecArrayIjLi6EE5emptyEv inlined from VecArray.h:330
```

>     * What about the SERIAL and TBB backends, do they run smoothly?

Yes, the work fine.

src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h
```diff
@@ -0,0 +1,404 @@
...
+                                              (ptrAsInt)(&cellNeighbors[i]));  // if fails we cannot give "i" back...
+#endif
+        } else
+          return -1;
+      }
+      // TO DO: Alpaka equivalent does not exist?
+      //__threadfence();
+      // This is obviously not equivalent, but seems to be sufficient.
+      // To DO: address in a better way?
+      alpaka::syncBlockThreads(acc);
```
ghugo83, line 93: Here

src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h
```diff
@@ -0,0 +1,404 @@
...
+              acc, (ptrAsInt*)(&theTracks), zero, (ptrAsInt)(&cellTracks[i]));  // if fails we cannot give "i" back...
+#endif
+        } else
+          return -1;
+      }
+      // TO DO: Alpaka equivalent does not exist?
+      //__threadfence();
+      // This is obviously not equivalent, but seems to be sufficient.
+      // TO DO: address in a better way?
+      alpaka::syncBlockThreads(acc);
```
ghugo83, line 121: And there

ghugo83:
Hmm without looking at anything, as mentioned in https://github.com/cms-patatrack/pixeltrack-standalone/pull/181#issuecomment-830104220 , there is one fish in GPUCACell.h, which absolutely need to be addressed before anything:
finding a `__threadfence()` equivalent, present workaround is of course absolutely not ok.
In `src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h `, in `addTrack` and `addOuterNeighbor`
Not sure this has any link with the issue, but these are called in `GPUCACell::find_ntuplets` and would definitely need to be the first point addressed anyway I would say.

felicepantaleo:
> Hmm without looking at anything, as mentioned in [#181 (comment)](https://github.com/cms-patatrack/pixeltrack-standalone/pull/181#issuecomment-830104220) , there is one fish in `GPUCACell.h`, which absolutely needs to be addressed before anything:
> finding a `__threadfence()` equivalent, present workaround is of course absolutely not ok.
> In `src/alpaka/plugin-PixelTriplets/alpaka/GPUCACell.h `, in `addTrack` and `addOuterNeighbor`
> Not sure this has any link with the issue, but these are called in `GPUCACell::find_ntuplets` and would definitely need to be the first point addressed anyway I would say.

@psychocoderHPC , @j-stephan would you be able to advise on this `__threadfence()` issue?

fwyzard:
I tried the exact same configuration on a GTX 1080, and it works fine:
```
$ CUDA_VISIBLE_DEVICES=0 ./alpaka --cuda --maxEvents 1
Processing 1 events, of which 1 concurrently, with 1 threads.
Processed 1 events in 7.215500e-02 seconds, throughput 13.8591 events/s.
```

So I assume the problem may come from the change in thread behaviour inside warps between Pascal and Volta.

fwyzard:
By the way, with this now I get
```
$ CUDA_VISIBLE_DEVICES=0 ./alpaka --cuda --maxEvents 1000
Processed 1000 events in 6.210932e+00 seconds, throughput 161.006 events/s.
```
vs
```
$ CUDA_VISIBLE_DEVICES=0 ./cuda --maxEvents 1000
Found 1 devices
Processed 1000 events in 1.951040e+00 seconds, throughput 512.547 events/s.
```

ghugo83:
With succesful `-DCUDA_DISABLE_CACHING_ALLOCATOR` and `-DCUDA_DISABLE_ASYNC_ALLOCATOR `for the cuda case?

ghugo83:
>Without the caching allocator, cuda gives

Ok. Yep the effect of caching allocator is impressive :)

Now, what about `disable_async` for legacy?
The present work on device is asynchronous with host in legacy, while ~synchronous in alpaka, hence this needs to be looked at for a fair comparison. 
There are alpaka asynchronous queues, but defined locally (with the only goal that the code changes will be minimal when a framework implementation is added), no work passed around producers: hence presently, the work on device do not overlap with any meaningful work on cpu. 
At this stage, to be fully clean, one could simply change the type of Queue to synchronous (`alpaka::QueueCudaRtBlocking`) in `alpaka/AlpakaCore/alpakaConfigAcc.h` (which leads to nearly similar perf as `alpaka::QueueCudaRtNonBlocking` for now, for the reasons explained above).

j-stephan:
> @psychocoderHPC , @j-stephan would you be able to advise on this __threadfence() issue?

Thanks for notifying us of this PR!

I have linked this issue in the alpaka bug tracker: https://github.com/alpaka-group/alpaka/issues/14. As you can see we have had this on our TODO list for a while but so far had no user requests to actually implement this. Since this has changed we will target this for the alpaka 0.8 release (expected 01 November 2021). If you are feeling daring: it might show up sooner than that on the `develop` branch. I will notify you of all developments in this matter.

For now I advise you to use the native CUDA functionality (if possible / sufficient).

ghugo83:
@j-stephan Ok thanks a lot!

src/alpaka/AlpakaCore/alpakaWorkDivHelper.h
```diff
@@ -8,234 +8,225 @@ using namespace alpaka_common;
...
-        const uint32_t firstElementIdx = firstElementIdxLocal + elementIdxShift[dimIndex];  // Add the shift!
-        const uint32_t endElementIdxUncut = firstElementIdx + threadDimension;
-
-        firstElementIdxVec[dimIndex] = firstElementIdx;
-        endElementIdxUncutVec[dimIndex] = endElementIdxUncut;
-      }
+    template <typename TAcc>
+    ALPAKA_FN_ACC std::pair<Idx, Idx> element_index_range_in_block(const TAcc& acc,
+                                                                   const Idx elementIdxShift,
+                                                                   const Idx dimIndex = 0u) {
```
fwyzard, line 46: If I read the code correctly, this is the index of the dimension (i.e. the equivalent of `.x`, `.y`, or `.z` in CUDA).
In this case the type should probably not be `Idx` but `unsigned int` ?

src/alpaka/AlpakaCore/alpakaWorkDivHelper.h
```diff
@@ -8,234 +8,225 @@ using namespace alpaka_common;
...
-      auto&& [firstElementIdxLocalVec, endElementIdxLocalVec] = element_index_range_in_block(acc, elementIdxShift);
+      //static_assert(alpaka::Dim<TAcc>::value == Dim1::value,
+      //              "Accelerator and maxNumberOfElements need to have same dimension.");
+      auto [firstElementIdxLocal, endElementIdxLocal] = element_index_range_in_block(acc, elementIdxShift, dimIndex);
 
       // Truncate
-      for (typename T_Dim::value_type dimIndex(0); dimIndex < T_Dim::value; ++dimIndex) {
-        endElementIdxLocalVec[dimIndex] = std::min(endElementIdxLocalVec[dimIndex], maxNumberOfElements[dimIndex]);
-      }
+      endElementIdxLocal = std::min(endElementIdxLocal, maxNumberOfElements);
```
fwyzard, line 77: IIUC this makes sense only for 1D grids ?

makortel:
Rebased to take out #180 (including merge commits that included updates of #180 to here).

src/alpaka/AlpakaCore/alpakaWorkDivHelper.h
ghugo83: Yes.
I had done helpers targeting more dimensions at first, but then realized that the use cases of 2D grids for the pixel track reconstruction were very few, and that actually for these, I could not use the 2D helper functions as they were.
I had taken care that the `indexShift` / `maxNumberOfElements` could obviously be specified for each dimension, but there was the extra issue that one can do a grid stride for a given dimension, and a block stride for another dimension. Plus issues with the body of the lambda. Overall, this was making things more complicated then helpful on call site.

In legacy, the 2D grids are in `plugin-PixelTriplets/alpaka/gpuPixelDoubletsAlgos.h`, `plugin-PixelTriplets/alpaka/CAHitNtupletGeneratorKernelsImpl.h` and `plugin-PixelTriplets/alpaka/gpuFishbone.h`. I kept the use of 2D grids, just found it equivalent and much simpler to call 1D helpers for each dimension on call site (which is what the 2D helper would have been doing anyway).

src/alpaka/AlpakaCore/alpakaWorkDivHelper.h
ghugo83: Yes, correct. At first it was `alpaka::Dim<TAcc>`, which is not possible anymore as, like explained below, a simple 1D access is used also for 2D grids (kernels launched with `Acc2`), for each dimension. Using `Dim1` (`alpaka::DimInt<1u>`) would be very inelegant, because of the use for 2D grids. Hence I have just used `Idx.` 
But indeed, having simply `unsigned int` (or better, a `DimIdx` ?) is the way to go, will add a small correction (also accordingly on call site).

makortel:
If I read the code correctly, the workaround by @fwyzard https://github.com/cms-patatrack/pixeltrack-standalone/pull/181#issuecomment-836663856 is not yet included here. @ghugo83 Can you add it (either squashing or in a separate commit), or shall I?

(my test of the previous not-squashed-to-single-commit failed because of '/alpaka --cuda` gettting stuck)

ghugo83:
@makortel From what I saw he added a clean fix in a later PR, I will cherry-pick his commit tonight after work (along with changed mentioned at https://github.com/cms-patatrack/pixeltrack-standalone/pull/181#discussion_r634281692 )

fwyzard:
Yes, please take https://github.com/cms-patatrack/pixeltrack-standalone/pull/184/commits/fed3ba2d8fcfecaee0cfb9e5454a65b58feec581 and https://github.com/cms-patatrack/pixeltrack-standalone/pull/184/commits/ba30ba8865e0630edc5784c053814cfc8a6c3ebf rather than what I put in the comment.

src/alpaka/AlpakaCore/alpakaWorkDivHelper.h
ghugo83: Just replaced `Idx` for dim indices with `unsigned int` in the end, likely indices types and their use on call sites would need to be thought of more though.

ghugo83: To be consistent, do you think `Idx` should be used on call sites instead of `uin32_t` for element indices? (even if that's obviously the same right now).

makortel:
Doing the same as in #180, i.e. merging now even if the review is incomplete to facilitate further work. Review comments should eventually be summarized in a separate issue.

(I did check the Alpaka version runs, and that the results are consistent with the `cuda` version)

j-stephan:
Hi @ghugo83, @fwyzard, @felicepantaleo and @makortel,

we have merged a memory fence API into alpaka's `develop` branch last week. Feel free to give it a try, we're happy to learn about any inconveniences or errors in this new feature. You can find a usage example here:

https://github.com/alpaka-group/alpaka/blob/develop/test/unit/mem/fence/src/FenceTest.cpp

fwyzard:
@j-stephan thanks for the update !

ghugo83:
Hi @j-stephan 
Yes indeed, thanks a lot!! Very useful, very clear interface!!
Just one question: why the choice of also calling https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/mem/fence/MemFenceCpu.hpp#L36 for the `AccCpuSerial` backend (one thread) (through https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/acc/AccCpuSerial.hpp#L102 ), instead of just an _empty_ function?

fwyzard:
Good point... I think it would make sense to keep the function for the `memory_scope::Device` case (one may want to synchronise with *other* kernels) and make it empty in the `memory_scope::Block` case (since there are no other threads in the serial case).

j-stephan:
> Just one question: why the choice of also calling [https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/mem/fence/MemFenceCpu.hpp#L36](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/mem/fence/MemFenceCpu.hpp?rgh-link-date=2021-08-24T13%3A17%3A30Z#L36) for the `AccCpuSerial` backend (through [https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/acc/AccCpuSerial.hpp#L102](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/acc/AccCpuSerial.hpp?rgh-link-date=2021-08-24T13%3A17%3A30Z#L102) ), instead of just an _empty_ function?

Good point, I didn't really think about the special `AccCpuSerial` case when implementing this. @fwyzard's suggestion is probably what I'll be implementing in the fix.

ghugo83:
Ok, thanks

PR #184, opened 2021-05-17
[cuda 11.2] Improve CUDA compilation

fwyzard:
Update `nvcc` compilation flags:
  - compile for multiple CUDA architectures in parallel;
  - display the numeric code for CUDA errors;
  - add external-specific flags for `nvcc`;
  - pass the `--diag-suppress 20014` flag to `nvcc` when Eigen is used.

All these changes require CUDA 11.2 or higher.


makortel:
Seems that `--display-error-number` came in CUDA 11.2. I'm not sure if I'd be ready switch from CUDA 11.1. At minimum I'd need (much) more testing first.

fwyzard:
> Seems that `--display-error-number` came in CUDA 11.2.

Correct.

I'll see if I can add a version check.

makortel:
I have the Kokkos part done in https://github.com/cms-patatrack/pixeltrack-standalone/commit/3887e20c5d2e8eb3b29eb28b5fb56c2401d8c765. Do you want to cherry-pick it here?

fwyzard:
I've implemented a version check, but all of
  - `--display-error-number`
  - `--diag-suppress NNN`
  - `--threads N`

require CUDA 11.2 or later.

So I would suggest just keeping this PR open until we can move to 11.2 everywhere.

makortel:
> So I would suggest just keeping this PR open until we can move to 11.2 everywhere.

Is this only pending my testing to move to 11.2, or are there other known cases for 11.1? (I hope to move to 11.2 next week following #187, I only want to run bunch of performance tests first and they take time).

makortel:
The README should also be updated (in this PR or shortly after) to bump the minimum CUDA requirement to 11.2.

fwyzard:
> The README should also be updated (in this PR or shortly after) to bump the minimum CUDA requirement to 11.2.

Done (I think).

fwyzard:
> Is this only pending my testing to move to 11.2, or are there other known cases for 11.1? (I hope to move to 11.2 next week following #187, I only want to run bunch of performance tests first and they take time).

I don't know of any other showstopper - we have 11.2 in CMSSW, and all machines we manage at CERN have either 11.2 or 11.3.

makortel:
I'd start to be ready to proceed with this PR, but now there is a conflict (that looks easily fixable) in te `Makefile`. @fwyzard Do you want to fix it or shall I? #193 would also conflict with this PR. If you have preference for merge order between #193 and this PR, I can follow that.

fwyzard:
hi @makortel, if you merge #193 first, I can fix all conflicts in this one

makortel:
Ok, I'll start with #193. Maybe we're lucky and that actually removes the conflict here (since it restores the conflicting line)?

makortel:
> Maybe we're lucky and that actually removes the conflict here (since it restores the conflicting line)?

Wishful thinking, the conflict is still there.

PR #185, opened 2021-06-09
[alpaka] Use inter-block instead of global atomics

fwyzard:


fwyzard:
This PR makes the Alpaka code use the "blocks" hierarchy of atomic operations - that is, operations that are atomic for different blocks in the same grid, but give no guarantee about different grids or host-device synchronization.

With these changes, the throughput of `alpaka --serial` gets closer to the `cudacompat` without the need to "hack" away the synchronizations of the CPU serial backend.

`alpaka --cuda` is unaffected (Alpaka seems to use the same atomic operations for all hierarchy levels with the CUDA backend).

src/alpaka/AlpakaCore/AtomicPairCounter.h
```diff
@@ -39,7 +39,7 @@ namespace cms {
         c += incr;
 
         Atomic2 ret;
-        ret.ac = alpaka::atomicOp<alpaka::AtomicAdd>(acc, &counter.ac, c);
+        ret.ac = alpaka::atomicAdd(acc, &counter.ac, c, alpaka::hierarchy::Blocks{});
```
VinInn, line 42: There is no guarantee that here (and elsewhere below) the atomic is limited to a single block

VinInn: Cuda DO differentiate between block and grid wide atomics, the fact that TODAY Alpaka does not differentiate is a defect that I hope will be corrected in future. At that time this code will break. So this is a hack anyhow

VinInn: For even more detailed discussion please refer to the presentation and code of https://github.com/NVIDIA/libcudacxx in particular
https://github.com/NVIDIA/libcudacxx/blob/main/include/cuda/std/detail/__atomic_generated
that I have not digested all yet.
(I do not exclude that some of those details is what causes A100 crashes.)

src/alpaka/plugin-PixelTriplets/alpaka/gpuPixelDoubletsAlgos.h
```diff
@@ -248,9 +248,9 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
             if (doPtCut && ptcut(oi, idphi))
               continue;
 
-            auto ind = alpaka::atomicOp<alpaka::AtomicAdd>(acc, nCells, 1u);
+            auto ind = alpaka::atomicAdd(acc, nCells, 1u, alpaka::hierarchy::Blocks{});
```
VinInn, line 251: everything below should be block scope. They need to be checked one by one!

src/alpaka/plugin-PixelVertexFinding/alpaka/gpuClusterTracksByDensity.h
```diff
@@ -193,7 +193,7 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
       cms::alpakatools::for_each_element_in_block_strided(acc, nt, [&](uint32_t i) {
         if (iv[i] == int(i)) {
           if (nn[i] >= minT) {
-            auto old = alpaka::atomicOp<alpaka::AtomicInc>(acc, &foundClusters, 0xffffffff);
+            auto old = alpaka::atomicInc(acc, &foundClusters, 0xffffffff, alpaka::hierarchy::Blocks{});
```
VinInn, line 196: everything below should be block scope. They need to be checked one by one!

src/alpaka/AlpakaCore/AtomicPairCounter.h
fwyzard: Yes, CUDA has
  - `atomicAdd_system`¹ that is atomic across all GPUs and all CPUs
  - `atomicAdd` that is atomic across the whole GPU, so including other grids and kernels
  - `atomicAdd_block`¹ that is atomic only within a single block

¹ available only from Pascal / sm 6.0 and later

While Alpaka has
  - `alpaka::atomicAdd(..., alpaka::hierarchy::Grids{})` (default) that is atomic for operations between grids within a device
  - `alpaka::atomicAdd(..., alpaka::hierarchy::Blocks{})` that is atomic for operations between blocks within a grid
  - `alpaka::atomicAdd(..., alpaka::hierarchy::Threads{})` that is atomic for operations between threads within a block

So,
  - Alpaka does not have an equivalent for `atomicAdd_system`
  - `alpaka::atomicAdd(..., alpaka::hierarchy::Grids{})` should map to `atomicAdd`
  - `alpaka::atomicAdd(..., alpaka::hierarchy::Blocks{})` should map to `atomicAdd` (for lack of a more fine-grained operation)
  - `alpaka::atomicAdd(..., alpaka::hierarchy::Threads{})` should map to `atomicAdd_block` on SM >= 6.0, or `atomicAdd` on older GPUs

However, I don't think we have been using `atomicAdd_block` in our CUDA coda so far.

VinInn: Not yet as we checked that (when applied to a block shared variable) the code generated was the same.
(was in the do to list)
And, for the utilities we do not know if used in a block or grid-wide (We may have to add a template param if makes a difference).

what are operations among "grids" within a device? Is something possible in CUDA?

fwyzard: > Not yet as we checked that (when applied to a block shared variable) the code generated was the same.

Yes, I expect that on variables in shared memory using `atomicAdd` and `atomicAdd_block` should generate the same code - since shared memory is known to be block-wise anyway.

I assume it can make a difference only for global memory (maybe because of caching ?).

> And, for the utilities we do not know if used in a block or grid-wide (We may have to add a template param if makes a difference).

OK. For the moment I'd stick with the safest option, i.e. grid-wide, and maybe review it later if we have concrete uses cases ?

> what are operations among "grids" within a device? Is something possible in CUDA?

One can have multiple kernels (or multiple launches of the same kernel) operate on the same global memory.
I don't think it would be useful; for us, given how we use the GPUs, but it is possible.

felicepantaleo:
The implementation is safe and is the correct way to avoid the hacking of the lock in the serial workflows. Good for me.

PR #186, opened 2021-06-10
Update the version of Eigen to include the Patatrack changes

fwyzard:
This PR updated the version of Eigen to include the changes originally developed by the Patatrack team, please some fixes added by Eric.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -442,8 +442,10 @@ $(TBB_LIB): $(TBB_BASE) $(TBB_LIBDIR)
 external_eigen: $(EIGEN_BASE)
 
 $(EIGEN_BASE):
-	git clone -b master https://gitlab.com/libeigen/eigen.git $@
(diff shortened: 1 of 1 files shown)


makortel:
Could you describe the expected benefit compared to the previous version of Eigen?

fwyzard:
It should just silence *a lot* of warnings.

makortel:
Sounds good.

ericcano:
The previous version missed many `__device__` function attibutes (in the form or `EIGEN_DEVICE_FUNC` macros), which led to many warnings during compilation. 
But on top, there is a bug in `nvcc`, where it just emits warnings (instead of failing), but also fails to generate the kernel code.

ericcano:
We have a simple reproducer here: https://godbolt.org/z/K5ccocTWW. The second kernel (at the bottom of the PTX output) is just empty.

PR #193, opened 2021-06-24
[cuda] Keep PTX in the final binaries

fwyzard:
Keep the generated PTX in the final binaries.

This should have two advantages:
  - allow running on newer GPUs that were not explicitly targeted thanks to the JIT compilation
  - allow to extract and inspect the generated PTX

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,8 @@ export CUDA_TEST_CXXFLAGS := -DGPU_DEBUG
 export CUDA_LDFLAGS := -L$(CUDA_BASE)/lib64 -lcudart -lcudadevrt
 export CUDA_NVCC := $(CUDA_BASE)/bin/nvcc
 define CUFLAGS_template
-$(2)NVCC_FLAGS := $$(foreach ARCH,$(1),-gencode arch=compute_$$(ARCH),code=sm_$$(ARCH)) -Wno-deprecated-gpu-targets -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored --expt-relaxed-constexpr --expt-extended-lambda --generate-line-info --source-in-ptx --cudart=shared
-$(2)NVCC_COMMON := -std=c++17 -O3 $$($(2)NVCC_FLAGS) -ccbin $(CXX) --compiler-options '$(HOST_CXXFLAGS) $(USER_CXXFLAGS)' -g -lineinfo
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -44,8 +44,8 @@ export CUDA_TEST_CXXFLAGS := -DGPU_DEBUG
 export CUDA_LDFLAGS := -L$(CUDA_BASE)/lib64 -lcudart -lcudadevrt
 export CUDA_NVCC := $(CUDA_BASE)/bin/nvcc
 define CUFLAGS_template
-$(2)NVCC_FLAGS := $$(foreach ARCH,$(1),-gencode arch=compute_$$(ARCH),code=sm_$$(ARCH)) -Wno-deprecated-gpu-targets -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored --expt-relaxed-constexpr --expt-extended-lambda --generate-line-info --source-in-ptx --cudart=shared
-$(2)NVCC_COMMON := -std=c++17 -O3 $$($(2)NVCC_FLAGS) -ccbin $(CXX) --compiler-options '$(HOST_CXXFLAGS) $(USER_CXXFLAGS)' -g -lineinfo
+$(2)NVCC_FLAGS := $$(foreach ARCH,$(1),-gencode arch=compute_$$(ARCH),code=[sm_$$(ARCH),compute_$$(ARCH)]) -Wno-deprecated-gpu-targets -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored --expt-relaxed-constexpr --expt-extended-lambda --generate-line-info --source-in-ptx --cudart=shared
```
fwyzard, line 47: by the way, re-reading the rule, I'm guessing these
```
-Wno-deprecated-gpu-targets -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored --expt-relaxed-constexpr --expt-extended-lambda --generate-line-info --source-in-ptx --cudart=shared
```
should actually go to `NVCC_COMMON` ?

makortel: Functionally doesn't seem to make much difference (`NVCC_FLAGS` is used only in `NVCC_COMMON`, unless I missed something). For readability I think it would be better to keep these parameters in `NVCC_FLAGS` to keep the `NVCC_COMMON` short-enough to show the general structure (especially towards arguments for `--compiler-options`).

fwyzard: OK, no problem. In fact I simply misread the `foreach` expansion - keeping things as-is is just fine.

makortel (approved):

Issue #178, opened 2021-03-26
Analyzing problems with HIP port


markdewing:
Track investigations into the HIP port

Snapshot of the state is in #161

Tests with ROCm 4.1
Main 'hip' executable still fails with the same error

Now working:
* prefixScan_t
* eigenSoA_t

Still failing
* radixSort_t
* gpuVertexFinder_t

Broken differently - testEigenGPUNoFit returns an error rather than hangs
```
./testEigenGPUNoFit
TEST EIGENVALUES
TEST INVERSE 3x3
TEST INVERSE 4x4
TEST INVERSE 5x5
:0:rocdevice.cpp            :2325: 4804578096 us: Device::callbackQueue aborting with status: 0x1016
Aborted (core dumped)
```

New failure?
```
./testEigenGPU
...
testEigenGPU: /mnt/nvme/physics/hep/tracking/patatrack/pixeltrack-standalone/src/hip/test/test_common.h:24: bool isEqualFuzzy(C1, C2, double) [C1 = Eigen::Matrix<double, 2, 1, 0, 2, 1>, C2 = Eigen::Matrix<double, 2, 1, 0, 2, 1>]: Assertion `std::abs(a(i, j) - b(i, j)) < std::min(std::abs(a(i, j)), std::abs(b(i, j))) * epsilon' failed.
Aborted (core dumped)
```

markdewing:
Trying to get rocgdb to work to make debugging some of these failures a little easier.  Still not working.

Saga here: https://github.com/ROCm-Developer-Tools/ROCgdb/issues/6

markdewing:
`gpuVertexFinder_t` hangs in radix sort.
Looking into the failures in `radixSort_t`

markdewing:
One issue with radix sort is in the while loop using `ibs` as the control variable.  At the end of the loop, there is the code
```
if (threadIdx.x == 0)
  ibs -= sb;
__syncthreads();
```
Apparently the decrement of `ibs` does not get propagated to the other threads.  The loop repeats until the assert right before this code triggers (`assert(c[bin] >= 0)`)
If a printf statement is added to print out the value of `ibs`, the radix sort works (for some cases).   

This change also causes `gpuVertexFinder_t` finish its run and not hang.

The radix sort test still issues messages about values being out of order.

markdewing:
With rocm 4.2

Main hip executable:
```
[...]

PR #195, opened 2021-07-02
Fix shorten function and add HIP workarounds

markdewing:
Add the workarounds listed in #178 

Assign the `ibs` variable to another variable - fixes radixSort_t (`ibs2`) and gpuVertexFinder (`ibs3`)

Remove launch bounds - fixes radixSort_t where the kernel silently fails to run

Fix the shorten function to avoid overwriting other memory if the type size is less than 8 bytes.

### DIFF
--- a/src/hip/CUDACore/radixSort.h
+++ b/src/hip/CUDACore/radixSort.h
@@ -92,6 +92,8 @@ __device__ __forceinline__ void radixSortImpl(
   __shared__ int32_t c[sb], ct[sb], cu[sb];
 
   __shared__ int ibs;
+  __shared__ int ibs2;
+  __shared__ int ibs3;
   __shared__ int p;
 
   assert(size > 0);
@@ -148,6 +150,7 @@ __device__ __forceinline__ void radixSortImpl(
 
     // broadcast
     ibs = size - 1;
+    ibs3 = ibs;
     __syncthreads();
     while (__syncthreads_and(ibs > 0)) {
       int i = ibs - threadIdx.x;
@@ -177,8 +180,10 @@ __device__ __forceinline__ void radixSortImpl(
       __syncthreads();
       if (bin >= 0)
         assert(c[bin] >= 0);
-      if (threadIdx.x == 0)
+      if (threadIdx.x == 0) {
         ibs -= sb;
+        ibs2 = ibs;
+      }
       __syncthreads();
     }
 
@@ -260,7 +265,9 @@ namespace cms {
   namespace hip {
 
     template <typename T, int NS = sizeof(T)>
-    __global__ void __launch_bounds__(256, 4)
+    // The launch bounds seems to cause the kernel to silently fail to run (rocm 4.2)
+    //__global__ void __launch_bounds__(256, 4)
+    __global__ void
         radixSortMultiWrapper(T const* v, uint16_t* index, uint32_t const* offsets, uint16_t* workspace) {
       radixSortMulti<T, NS>(v, index, offsets, workspace);
     }
--- a/src/hip/CUDACore/radixSort.h
+++ b/src/hip/CUDACore/radixSort.h
@@ -92,8 +92,6 @@ __device__ __forceinline__ void radixSortImpl(
   __shared__ int32_t c[sb], ct[sb], cu[sb];
 
   __shared__ int ibs;
-  __shared__ int ibs2;
-  __shared__ int ibs3;
   __shared__ int p;
 
   assert(size > 0);
@@ -151,8 +149,6 @@ __device__ __forceinline__ void radixSortImpl(
     // broadcast
     ibs = size - 1;
 
-    // Workaround for hang in gpuVertexFinder_t.
-    ibs3 = ibs;
 
     __syncthreads();
     while (__syncthreads_and(ibs > 0)) {
@@ -185,8 +181,10 @@ __device__ __forceinline__ void radixSortImpl(
         assert(c[bin] >= 0);
       if (threadIdx.x == 0) {
         ibs -= sb;
-        // Workaround for problems in radixSort_t.
-        ibs2 = ibs;
+
+        // Fix for problems in radixSort_t.
+        // Without this, this c[bin] >=0 assert above will be triggered.
+        __threadfence();
       }
       __syncthreads();
     }
--- a/src/hip/CUDACore/radixSort.h
+++ b/src/hip/CUDACore/radixSort.h
@@ -150,7 +150,10 @@ __device__ __forceinline__ void radixSortImpl(
 
     // broadcast
     ibs = size - 1;
+
+    // Workaround for hang in gpuVertexFinder_t.
     ibs3 = ibs;
+
     __syncthreads();
     while (__syncthreads_and(ibs > 0)) {
       int i = ibs - threadIdx.x;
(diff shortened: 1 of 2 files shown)


src/hip/CUDACore/radixSort.h
```diff
@@ -148,6 +150,7 @@ __device__ __forceinline__ void radixSortImpl(
 
     // broadcast
     ibs = size - 1;
+    ibs3 = ibs;
```
fwyzard, line 153: could you add comments on the code about why the ibs2/ibs3 workaround is needed ?

src/hip/test/radixSort_t.cu
```diff
@@ -140,8 +140,9 @@ void go(bool useShared) {
         auto sh = sizeof(uint64_t) - NS;
         sh *= 8;
         auto shorten = [sh](T& t) {
-          auto k = (uint64_t*)(&t);
```
fwyzard, line 143: is this approach used only in the test ? or also in some of the application code ?

markdewing: The `shorten` function is only used in the radixSort_t test in the cuda* ports and the hip port.  The radixSort function is used in plugin-PixelVertexFinding/gpuSortByPt2.h, and it doesn't seem that any such bit-level processing of the numbers is needed there.

makortel: (apologies for long delay)

I'm trying to understand what this lambda is supposed to do
```cpp
        auto sh = sizeof(uint64_t) - NS;
        sh *= 8;
        auto shorten = [sh](T& t) {
          auto k = (uint64_t*)(&t);
          *k = (*k >> sh) << sh;
        };
```
It gets run for `T ` with `sizeof` equal to that of `int8_t, int16_t, int32_t, int64_t`, leading to `sh` values of 56, 48, 32. The code appears to zero the lowest `sh` number of bits, i.e. the portion of `T` of the `uint64_t` piece of memory. Adding a printout here (for `cuda` program) confirms that whenever `sh != 0`, both `k1` and `k2` are `0`.

@VinInn, could you help here with the intention of this lambda? (the code is exactly the same in CMSSW)

VinInn: to test sorting? (one way to speed up radix sort is to consider only the MSBs.  for instance in ptsort only the 16MSBs are used. 
most probably this lambda is used to verify that the result is the same (I need to check)

src/hip/CUDACore/radixSort.h
```diff
@@ -260,7 +269,9 @@ namespace cms {
   namespace hip {
 
     template <typename T, int NS = sizeof(T)>
-    __global__ void __launch_bounds__(256, 4)
+    // The launch bounds seems to cause the kernel to silently fail to run (rocm 4.2)
```
VinInn, line 272: known issue?
what hippify tells you about?

src/hip/CUDACore/radixSort.h
```diff
@@ -177,8 +183,11 @@ __device__ __forceinline__ void radixSortImpl(
       __syncthreads();
       if (bin >= 0)
         assert(c[bin] >= 0);
-      if (threadIdx.x == 0)
+      if (threadIdx.x == 0) {
         ibs -= sb;
+        // Workaround for problems in radixSort_t.
+        ibs2 = ibs;
```
VinInn, line 189: which problem?

src/hip/CUDACore/radixSort.h
VinInn: where ibs3 is used?

src/hip/CUDACore/radixSort.h
VinInn: where ibs2 is used?

VinInn: I suspect what is needed is a threadfence

VinInn: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#memory-fence-functions
RocM memory model maybe even weaker than CUDA

VinInn:
from HIP doc
```
The __syncthreads() built-in function is supported in HIP. The __syncthreads_count(int), __syncthreads_and(int) and __syncthreads_or(int) functions are under development.
```
not sure what exaclly means

src/hip/test/radixSort_t.cu
makortel: > most probably this lambda is used to verify that the result is the same (I need to check)

Right, but setting the compared-to values to zero for all types `T` shorter than `uint64_t` doesn't sound very useful for that.

VinInn: ok. most probably was written for 64bit values and never fix it...

src/hip/CUDACore/radixSort.h
markdewing: Using a threadfence instead of the temporary variable works.

src/hip/test/radixSort_t.cu
fwyzard: Here is a version that keep GCC 11 happy and seems to work as intended for arbitrary types:
```c++
auto shorten = [](T& t) {
    // byte representation of t
    char* bytes = reinterpret_cast<char*>(&t);
    // bytes to zero out
    const int zeroes = static_cast<int>(sizeof(T)) - NS;
    // zero out the least significant bytes (assuming a little endian architecture)
    static_assert(__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__, "This test assumes a little-endian architecture");
    for (int i = 0; i < zeroes; ++i) {
        bytes[i] = 0x00;
    }
};
```

fwyzard: And here is a version that is more similar to the original, and avoids the loop, but adds a lot of boilerplate to keep the compiler happy about "aliasing rules":
```c++
template <int N>
struct uintN;

template <>
struct uintN<8> {
    using type = uint8_t;
};

template <>
struct uintN<16> {
    using type = uint16_t;
};

template <>
struct uintN<32> {
    using type = uint32_t;
};

template <>
struct uintN<64> {
    using type = uint64_t;
};

template <int N>
using uintN_t = typename uintN<N>::type;

template <typename T>
using uintT_t = uintN_t<sizeof(T) * 8>;


template <typename T>
union safe_uint_cast {
    T t;
    uintT_t<T> u;
};

template <int N, typename T, typename SFINAE = std::enable_if_t<N <= sizeof(T)>>
void shorten(T& t) {
    const int shift = 8 * (sizeof(T) - N);
    safe_uint_cast<T> c;
    c.t = t;
    c.u = c.u >> shift << shift;
    t = c.t;
}
```

fwyzard: PR'ed in #209.

makortel:
@markdewing Could you try if the version of `shorten()` merged in #209 would work for HIP too?

markdewing:
@makortel The version of shorten from #209 works on HIP.  

The remaining changes needed to make the radixSort_t test pass on HIP are
* remove `__launch_bounds__` from the function declaration near the end of radixSort.h
* add `__threadfence();` around line 182 of radixSort.h

makortel:
Thanks @markdewing. Could you change this PR (or close this one and open a new one) to contain

> The remaining changes needed to make the radixSort_t test pass on HIP are
> 
> * remove `__launch_bounds__` from the function declaration near the end of radixSort.h
> * add `__threadfence();` around line 182 of radixSort.h

and then I'll merge?

markdewing:
I might open a new PR with the simplified changes.

PR #196, opened 2021-07-19
[alpaka] Remove CUDA from hard dependencies

felicepantaleo:
This PR removes CUDA from the hard dependencies of the alpaka build target, as in principle it is not a requirement. 

This is a first step towards the addition of the HIP backend of alpaka.


felicepantaleo:
Compiles and run on a machine with NVIDIA GPU, and on another machine without an NVIDIA GPU.

src/alpaka/Makefile
```diff
@@ -4,6 +4,10 @@ TARGET := $(BASE_DIR)/$(TARGET_NAME)
 include Makefile.deps
 EXTERNAL_DEPENDS := $(alpaka_EXTERNAL_DEPENDS)
 
+ifneq ($(wildcard $(CUDA_BASE)),)
+EXTERNAL_DEPENDS += CUDA
+endif
```
makortel, line 9: Please move this to `Makefile.deps`, otherwise the main `Makefile` won't be aware of it (although with system dependencies it makes little difference in practice).

You could also use `ifdef CUDA_BASE` for the check (that is used already elsewhere). The main `Makefile` leaves `CUDA_BASE` undefined if it detects CUDA doesn't exist.

(I'll look the rest later today, they need a bit more thought)

src/alpakatest/bin/main.cc
```diff
@@ -6,11 +6,12 @@
 #include <filesystem>
 #include <string>
 #include <vector>
+#include "AlpakaCore/alpakaConfig.h"
```
ghugo83, line 9: @felicepantaleo Ha just one thing: this `alpakatest/bin/main.cc` is actually an old version of `alpaka/bin/main.cc` (the one used for the pixeltrack reconstruction toolchain), where among other aspects I had already addressed this issue of the inclusion of `cuda_runtime.h`.
Maybe that latest version of `main.cc` (the one from `src/alpaka`, https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/alpaka/bin/main.cc) could be propagated here, to uniform things, and updated when needed?
The only diff between these 2 files should actually be the `edmodules` / `esmodules`.

If I remember properly, the inclusion of `cuda_runtime.h` is already handled inside `alpaka` anyway, so in a sense, we do not need to re-expose this treatment here? Basically, the choice of the CUDA backend will already trigger the `cuda_runtime.h` inclusion anyway (handled inside `alpaka` library).

ghugo83: Ha found it back:
- `DALPAKA_ACC_GPU_CUDA_ENABLED` is set here:
https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/alpaka/Makefile#L169 
- `alpaka.hpp` (included at https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/alpaka/AlpakaCore/alpakaConfigCommon.h#L4 or similar for the test side) includes `Cuda.hpp`:
https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/alpaka.hpp#L65
- `Cuda.hpp` only includes `cuda_runtime.h` (among others) inside `#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED`: https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/core/Cuda.hpp#L12

src/alpakatest/bin/main.cc
```diff
@@ -6,11 +6,12 @@
...
 
 #include <tbb/task_scheduler_init.h>
-
+#ifdef alpaka_cuda
```
ghugo83, line 12: Here ;)

src/alpakatest/bin/main.cc
ghugo83: This also means that for HIP, one might only need to add an extra compilation treatement with a -D ALPAKA_ACC_GPU_HIP_ENABLED, and then the hip_runtime.h will be automatically included as desired (Hip.hpp in alpaka.

src/alpakatest/bin/main.cc
```diff
@@ -69,9 +67,6 @@ int main(int argc, char** argv) {
       datadir = *i;
     } else if (*i == "--transfer") {
       transfer = true;
-    } else if (*i == "--validation") {
-      transfer = true;
-      validation = true;
```
makortel, line 74: Please remove the `validation` from here
https://github.com/cms-patatrack/pixeltrack-standalone/blob/ae72abc417649b0b32eaed854de16980ac454d08/src/alpakatest/bin/main.cc#L121-L122
(it fails to compile otherwise)

It would be good to remove it also from
https://github.com/cms-patatrack/pixeltrack-standalone/blob/ae72abc417649b0b32eaed854de16980ac454d08/src/alpakatest/bin/main.cc#L20
and https://github.com/cms-patatrack/pixeltrack-standalone/blob/ae72abc417649b0b32eaed854de16980ac454d08/src/alpakatest/bin/main.cc#L30

felicepantaleo:
Thanks, all comments have been addressed.
Tested with and without CUDA, with and without NVIDIA GPU.

src/alpaka/Makefile
```diff
@@ -45,13 +45,14 @@ $(1)_SERIAL_DEP := $$($(1)_SERIAL_OBJ:$.o=$.d)
 # TBB backend
 $(1)_TBB_OBJ := $$(patsubst $(SRC_DIR)%,$(OBJ_DIR)%,$$($(1)_PORTABLE_SRC:%=%.tbb.o))
 $(1)_TBB_DEP := $$($(1)_TBB_OBJ:$.o=$.d)
-# CUDA backend
+# # CUDA backend
```
fwyzard, line 48:
```suggestion
# CUDA backend
```

src/alpaka/Makefile
```diff
@@ -225,3 +239,5 @@ $(OBJ_DIR)/$(TARGET_NAME)/test/alpaka/%.cudadlink.o: $(OBJ_DIR)/$(TARGET_NAME)/t
 $(TEST_DIR)/$(TARGET_NAME)/%.cuda: $(OBJ_DIR)/$(TARGET_NAME)/test/alpaka/%.cc.cuda.o $(OBJ_DIR)/$(TARGET_NAME)/test/alpaka/%.cc.cuda.cudadlink.o | $(LIBS)
 	@[ -d $(@D) ] || mkdir -p $(@D)
 	$(CXX) $^ $(LDFLAGS) $(MY_LDFLAGS) -o $@ -L$(LIB_DIR)/$(TARGET_NAME) $(patsubst %,-l%,$(LIBNAMES)) $(foreach dep,$(EXTERNAL_DEPENDS),$($(dep)_LDFLAGS))
+endif
+
```
fwyzard, line 243:
```suggestion
```

src/alpakatest/bin/main.cc
```diff
@@ -88,47 +83,48 @@ int main(int argc, char** argv) {
...
+    
     addModules("alpaka_serial_sync::", Backend::SERIAL);
     addModules("alpaka_tbb_async::", Backend::TBB);
     addModules("alpaka_cuda_async::", Backend::CUDA);
     esmodules = {"IntESProducer"};
     if (transfer) {
       // add modules for transfer
     }
-  }
+  }
```
fwyzard, line 120:
```suggestion
  }
```

src/alpakatest/Makefile
```diff
@@ -226,3 +236,4 @@ $(OBJ_DIR)/$(TARGET_NAME)/test/alpaka/%.cudadlink.o: $(OBJ_DIR)/$(TARGET_NAME)/t
 $(TEST_DIR)/$(TARGET_NAME)/%.cuda: $(OBJ_DIR)/$(TARGET_NAME)/test/alpaka/%.cc.cuda.o $(OBJ_DIR)/$(TARGET_NAME)/test/alpaka/%.cc.cuda.cudadlink.o | $(LIBS)
 	@[ -d $(@D) ] || mkdir -p $(@D)
 	$(CXX) $^ $(LDFLAGS) $(MY_LDFLAGS) -o $@ -L$(LIB_DIR)/$(TARGET_NAME) $(patsubst %,-l%,$(LIBNAMES)) $(foreach dep,$(EXTERNAL_DEPENDS),$($(dep)_LDFLAGS))
+endif
```
fwyzard, line 239:
```suggestion
endif
```

PR #197, opened 2021-07-23
Various improvements to Kokkos version

makortel:
The first three commits are to bring the behavior of the Kokkos version closer to `cudacompat` (and `cuda`). The last commit is for cleaner code (that may become important for future architectures).

### DIFF
--- a/src/kokkos/KokkosCore/atomic.h
+++ b/src/kokkos/KokkosCore/atomic.h
@@ -1,0 +1,112 @@
+#ifndef KokkosCore_atomic_h
+#define KokkosCore_atomic_h
+
+#include <Kokkos_Core.hpp>
+
+// Abstractions to optionally disable atomicity of atomics when using serial backend only
+// TODO: It would be better to do this only depending on the "active
+// backend", regardless of how the Kokkos runtime library has been
+// configured. I can see two options
+// - Template over the execution space. But execution space objects are not available in the device code.
+// - Enclose everything in KOKKOS_NAMESPACE namespace for the "poor-man" templating we're doing with that
+// To be looked into after performance impact is demonstrated with Serial-only build
+namespace cms::kokkos {
+  template <typename T>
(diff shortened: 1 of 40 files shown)


src/kokkos/KokkosCore/atomic.h
```diff
@@ -0,0 +1,102 @@
+#ifndef KokkosCore_atomic_h
+#define KokkosCore_atomic_h
+
+#include <Kokkos_Core.hpp>
+
+// Abstractions to disable atomicity of atomics for serial backend
+namespace cms::kokkos {
+  template <typename T>
+  KOKKOS_FORCEINLINE_FUNCTION T atomic_fetch_add(T* dest, const T val) {
```
makortel, line 9: Note to self that this is fragile at best, and these functions should be (e.g.) templated on the execution space to guarantee different symbols. Probably the `FORCEINLINE` has saved for trouble in the tests so far.

fwyzard:
I guess these are needed because Kokkos does not provide no-op atomics for
the serial case.
Is it because there is no distinction of per-block, per-grid and per-device
atomics ?

makortel:
> I guess these are needed because Kokkos does not provide no-op atomics for the serial case. 

Correct. Avoiding real atomics for the serial case was suggested by Kokkos developers as a step towards closing the `cudacompat` performance (real impact still to be quantified).

> Is it because there is no distinction of per-block, per-grid and per-device atomics ?

Good question. The API currently is, AFAICT, such that there is no distinction between different levels of atomicity guarantees, but we should find out if they have any plans for such.

src/kokkos/KokkosCore/atomic.h
makortel: For now went with a simpler solution that the bypassing of atomics is an option available when only the Serial backend is enabled. This should be sufficient to demonstrate the impact on performance, after which some better solution can be figured out.

PR #199, opened 2021-07-28
Print a stack trace on CUDA errors

fwyzard:
Extend #189 to the other variants.


makortel:
Looks good by eye. The [dependence table in README](https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/README.md#introduction) should also be eventually updated.

fwyzard:
> The [dependence table in README](https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/README.md#introduction) should also be eventually updated.

Done.

makortel:
Thanks!

Issue #188, opened 2021-06-13
Illegal memory access in the CUDA implementation with multiple concurrent processes


fwyzard:
When running multiple concurrent processes without MPS, one often runs into this error:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  
.../pixeltrack-standalone/src/cuda/CUDACore/CachingDeviceAllocator.h, line 617:
cudaCheck(error = cudaEventRecord(search_key.ready_event, search_key.associated_stream));
cudaErrorIllegalAddress: an illegal memory access was encountered
```
inside `notcub::DeviceFree(int device, void *d_ptr)`.

Anecdotal evidence suggests that this happens a lot more on Ampere vs Volta/Turing, and on larger (e.g. V100/A100) vs smaller (e.g. T4/A10) GPUs.

fwyzard:
This has been around for a while now - I'm currently investigating it starting from 4c34e1788c80b12e083e0f81a6889600aca239df .

fwyzard:
Adding a couple of checks earlier in the function:
```diff
diff --git a/src/cuda/CUDACore/CachingDeviceAllocator.h b/src/cuda/CUDACore/CachingDeviceAllocator.h
index 50c1ebd..44b1fae 100644
--- a/src/cuda/CUDACore/CachingDeviceAllocator.h
+++ b/src/cuda/CUDACore/CachingDeviceAllocator.h
@@ -556,6 +556,7 @@ namespace notcub {
     cudaError_t DeviceFree(int device, void *d_ptr) {
       int entrypoint_device = INVALID_DEVICE_ORDINAL;
       cudaError_t error = cudaSuccess;
+      cudaCheck(error = cudaGetLastError());  // line 559
       // CMS: use RAII instead of (un)locking explicitly
       std::unique_lock<std::mutex> mutex_locker(mutex, std::defer_lock);
 
@@ -612,6 +613,7 @@ namespace notcub {
       }
 
       if (recached) {
+        cudaCheck(error = cudaGetLastError());  // line 616
         // Insert the ready event in the associated stream (must have current device set properly)
         // CMS: throw exception on error
         cudaCheck(error = cudaEventRecord(search_key.ready_event, search_key.associated_stream));
```

shows that - at least in some cases - the error is a pre-existing condition; the call to `DeviceFree` is simply the first place that checks for errors:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  
[...]

PR #202, opened 2021-08-02
Fix uploading EventSetup conditions from multiple CUDA streams

fwyzard:
When multiple CUDA streams are trying to initialise the same EventSetup object, the first one to do
so starts the asynchronous operations, and the others are supposed to wait for it to finish.
However, code for recording the CUDA event was missing, so the other streams would find the default-
constructed event, which is always "valid".

Adding the missing call to record the event fixes the problem.


fwyzard:
Fixes #188

fwyzard:
The same fix seems to be working fine also in CMSSW (see e.g. https://github.com/cms-sw/cmssw/pull/34725).

PR #208, opened 2021-08-18
[all] Update Eigen and extend CUDA support

fwyzard:
Fixes compilation problems with CUDA 11.3 and later.

See [libeigen/eigen!586](https://gitlab.com/libeigen/eigen/-/merge_requests/586) for more details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -472,10 +472,10 @@ $(TBB_LIB): $(TBB_BASE) $(TBB_LIBDIR)
 external_eigen: $(EIGEN_BASE)
 
 $(EIGEN_BASE):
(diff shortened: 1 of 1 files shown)


fwyzard:
This PR is for testing and for helping anybody that runs into issues with CUDA 11.3 or 11.4.

However, I would rather wait for the changes to be merged upstream (or at least in CMS) instead of pointing to my branch.

fwyzard:
The same changes have been picked up by the CMS externals, so it should be safe to merge this.
I've update the Makefile to use the CMS branch instead of mine.

PR #209, opened 2021-08-18
[all] Fix radix sort test

fwyzard:
Fix the compilation errors in test/radixSort_t.cu:
  - take into account the actual size of the type being sorted, instead of always assuming a 64-bit type;
  - use an union instead of a cast to keep the compiler happy about the aliasing rules.


fwyzard:
See https://github.com/cms-sw/cmssw/pull/34929 .

I have not tested the HIP version, though.

PR #211, opened 2021-08-27
[cudadev] Macro based SoA

ericcano:
Converted the SiPixelROCsStatusAndMapping to macro generated SoA.

This implements the SoA macros discussed in https://github.com/cms-patatrack/cmssw/issues/272.

### DIFF
--- a/src/cudadev/CUDACore/device_unique_ptr.h
+++ b/src/cudadev/CUDACore/device_unique_ptr.h
@@ -49,8 +49,8 @@ namespace cms {
 
     template <typename T>
     typename device::impl::make_device_unique_selector<T>::non_array make_device_unique(cudaStream_t stream) {
-      static_assert(std::is_trivially_constructible<T>::value,
-                    "Allocating with non-trivial constructor on the device memory is not supported");
+      static_assert(std::is_trivially_copyable<T>::value,
+                    "Allocating with non-trivial copy on the device memory is not supported");
       void *mem = allocate_device(sizeof(T), stream);
       return typename device::impl::make_device_unique_selector<T>::non_array{reinterpret_cast<T *>(mem),
                                                                               device::impl::DeviceDeleter{stream}};
@@ -60,8 +60,8 @@ namespace cms {
     typename device::impl::make_device_unique_selector<T>::unbounded_array make_device_unique(size_t n,
                                                                                               cudaStream_t stream) {
       using element_type = typename std::remove_extent<T>::type;
-      static_assert(std::is_trivially_constructible<element_type>::value,
-                    "Allocating with non-trivial constructor on the device memory is not supported");
+      static_assert(std::is_trivially_copyable<element_type>::value,
+                    "Allocating with non-trivial copy on the device memory is not supported");
       void *mem = allocate_device(n * sizeof(element_type), stream);
       return typename device::impl::make_device_unique_selector<T>::unbounded_array{
           reinterpret_cast<element_type *>(mem), device::impl::DeviceDeleter{stream}};
--- a/src/cudadev/CUDACore/host_unique_ptr.h
+++ b/src/cudadev/CUDACore/host_unique_ptr.h
@@ -39,8 +39,8 @@ namespace cms {
     // Allocate pinned host memory
     template <typename T>
     typename host::impl::make_host_unique_selector<T>::non_array make_host_unique(cudaStream_t stream) {
-      static_assert(std::is_trivially_constructible<T>::value,
-                    "Allocating with non-trivial constructor on the pinned host memory is not supported");
+      static_assert(std::is_trivially_copyable<T>::value,
+                    "Allocating with non-trivial copy on the pinned host memory is not supported");
       void *mem = allocate_host(sizeof(T), stream);
       return typename host::impl::make_host_unique_selector<T>::non_array{reinterpret_cast<T *>(mem)};
     }
@@ -48,8 +48,8 @@ namespace cms {
     template <typename T>
     typename host::impl::make_host_unique_selector<T>::unbounded_array make_host_unique(size_t n, cudaStream_t stream) {
       using element_type = typename std::remove_extent<T>::type;
-      static_assert(std::is_trivially_constructible<element_type>::value,
-                    "Allocating with non-trivial constructor on the pinned host memory is not supported");
+      static_assert(std::is_trivially_copyable<element_type>::value,
+                    "Allocating with non-trivial copy on the pinned host memory is not supported");
       void *mem = allocate_host(n * sizeof(element_type), stream);
       return typename host::impl::make_host_unique_selector<T>::unbounded_array{reinterpret_cast<element_type *>(mem)};
     }
--- a/src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
+++ b/src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
@@ -4,12 +4,22 @@
 #include "CUDACore/device_unique_ptr.h"
 #include "CUDACore/host_unique_ptr.h"
 #include "CUDACore/cudaCompat.h"
+#include "DataFormats/SoAStore.h"
 
 #include <cuda_runtime.h>
 
 class SiPixelClustersCUDA {
 public:
-  SiPixelClustersCUDA() = default;
+  generate_SoA_store(DeviceStore,
+    SoA_column(uint32_t, moduleStart),  // index of the first pixel of each module
+    SoA_column(uint32_t, clusInModule), // number of clusters found in each module
+    SoA_column(uint32_t, moduleId),     // module id of each module
+
+    // originally from rechits
+    SoA_column(uint32_t, clusModuleStart) // index of the first cluster of each module
+  );
+
+  explicit SiPixelClustersCUDA();
   explicit SiPixelClustersCUDA(size_t maxModules, cudaStream_t stream);
   ~SiPixelClustersCUDA() = default;
 
@@ -22,41 +32,21 @@ public:
 
   uint32_t nClusters() const { return nClusters_h; }
 
-  uint32_t *moduleStart() { return moduleStart_d.get(); }
-  uint32_t *clusInModule() { return clusInModule_d.get(); }
-  uint32_t *moduleId() { return moduleId_d.get(); }
-  uint32_t *clusModuleStart() { return clusModuleStart_d.get(); }
+  uint32_t *moduleStart() { return deviceStore_.moduleStart(); }
+  uint32_t *clusInModule() { return deviceStore_.clusInModule(); }
+  uint32_t *moduleId() { return deviceStore_.moduleId(); }
+  uint32_t *clusModuleStart() { return deviceStore_.clusModuleStart(); }
 
-  uint32_t const *moduleStart() const { return moduleStart_d.get(); }
-  uint32_t const *clusInModule() const { return clusInModule_d.get(); }
-  uint32_t const *moduleId() const { return moduleId_d.get(); }
-  uint32_t const *clusModuleStart() const { return clusModuleStart_d.get(); }
+  uint32_t const *moduleStart() const { return deviceStore_.moduleStart(); }
+  uint32_t const *clusInModule() const { return deviceStore_.clusInModule(); }
+  uint32_t const *moduleId() const { return deviceStore_.moduleId(); }
+  uint32_t const *clusModuleStart() const { return deviceStore_.clusModuleStart(); }
 
-  class DeviceConstView {
-  public:
-    __device__ __forceinline__ uint32_t moduleStart(int i) const { return __ldg(moduleStart_ + i); }
-    __device__ __forceinline__ uint32_t clusInModule(int i) const { return __ldg(clusInModule_ + i); }
-    __device__ __forceinline__ uint32_t moduleId(int i) const { return __ldg(moduleId_ + i); }
-    __device__ __forceinline__ uint32_t clusModuleStart(int i) const { return __ldg(clusModuleStart_ + i); }
-
-    uint32_t const *moduleStart_;
-    uint32_t const *clusInModule_;
-    uint32_t const *moduleId_;
-    uint32_t const *clusModuleStart_;
-  };
-
-  DeviceConstView *view() const { return view_d.get(); }
+  const DeviceStore store() const { return deviceStore_; }
 
 private:
-  cms::cuda::device::unique_ptr<uint32_t[]> moduleStart_d;   // index of the first pixel of each module
-  cms::cuda::device::unique_ptr<uint32_t[]> clusInModule_d;  // number of clusters found in each module
-  cms::cuda::device::unique_ptr<uint32_t[]> moduleId_d;      // module id of each module
-
-  // originally from rechits
-  cms::cuda::device::unique_ptr<uint32_t[]> clusModuleStart_d;  // index of the first cluster of each module
-
-  cms::cuda::device::unique_ptr<DeviceConstView> view_d;  // "me" pointer
-
+  cms::cuda::device::unique_ptr<std::byte[]> data_d;         // Single SoA storage
+  DeviceStore deviceStore_;
   uint32_t nClusters_h = 0;
 };
 
--- a/src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc
+++ b/src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc
@@ -5,44 +5,54 @@
 #include "CUDACore/host_unique_ptr.h"
 
 SiPixelDigisCUDA::SiPixelDigisCUDA(size_t maxFedWords, cudaStream_t stream)
-    : xx_d(cms::cuda::make_device_unique<uint16_t[]>(maxFedWords, stream)),
-      yy_d(cms::cuda::make_device_unique<uint16_t[]>(maxFedWords, stream)),
-      adc_d(cms::cuda::make_device_unique<uint16_t[]>(maxFedWords, stream)),
-      moduleInd_d(cms::cuda::make_device_unique<uint16_t[]>(maxFedWords, stream)),
-      clus_d(cms::cuda::make_device_unique<int32_t[]>(maxFedWords, stream)),
-      view_d(cms::cuda::make_device_unique<DeviceConstView>(stream)),
-      pdigi_d(cms::cuda::make_device_unique<uint32_t[]>(maxFedWords, stream)),
-      rawIdArr_d(cms::cuda::make_device_unique<uint32_t[]>(maxFedWords, stream)) {
-  auto view = cms::cuda::make_host_unique<DeviceConstView>(stream);
-  view->xx_ = xx_d.get();
-  view->yy_ = yy_d.get();
-  view->adc_ = adc_d.get();
-  view->moduleInd_ = moduleInd_d.get();
-  view->clus_ = clus_d.get();
+    : data_d(cms::cuda::make_device_unique<std::byte[]>(
+        DeviceOnlyStore::computeDataSize(maxFedWords) + 
+        HostDeviceStore::computeDataSize(maxFedWords),
+      stream)),
+      deviceOnlyStore_d(data_d.get(), maxFedWords),
+      hostDeviceStore_d(deviceOnlyStore_d.soaMetadata().nextByte(), maxFedWords),
+      deviceFullView_(deviceOnlyStore_d, hostDeviceStore_d),
+      devicePixelView_(deviceFullView_)
+{}
 
-  cms::cuda::copyAsync(view_d, view, stream);
+SiPixelDigisCUDA::SiPixelDigisCUDA()
+    : data_d(),deviceOnlyStore_d(), hostDeviceStore_d(), deviceFullView_(), devicePixelView_()
+{}
+
+SiPixelDigisCUDA::HostStoreAndBuffer::HostStoreAndBuffer()
+     : data_h(), hostStore_(nullptr, 0)
+{}
+
+SiPixelDigisCUDA::HostStoreAndBuffer::HostStoreAndBuffer(size_t maxFedWords, cudaStream_t stream)
+     : data_h(cms::cuda::make_host_unique<std::byte[]>(SiPixelDigisCUDA::HostDeviceStore::computeDataSize(maxFedWords), stream)),
+       hostStore_(data_h.get(), maxFedWords)
+{}
+
+void SiPixelDigisCUDA::HostStoreAndBuffer::reset() {
+  hostStore_.~HostDeviceStore();
+  new(&hostStore_) HostDeviceStore(nullptr, 0);
+  data_h.reset();
 }
 
 cms::cuda::host::unique_ptr<uint16_t[]> SiPixelDigisCUDA::adcToHostAsync(cudaStream_t stream) const {
   auto ret = cms::cuda::make_host_unique<uint16_t[]>(nDigis(), stream);
-  cms::cuda::copyAsync(ret, adc_d, nDigis(), stream);
+  // TODO: this is downgraded from cms::cuda::copyAsync as we copy data from within a block but not the full block.
+  cudaCheck(cudaMemcpyAsync(ret.get(), deviceFullView_.adc(), nDigis() * sizeof(decltype(ret[0])), cudaMemcpyDeviceToHost, stream));
   return ret;
 }
 
-cms::cuda::host::unique_ptr<int32_t[]> SiPixelDigisCUDA::clusToHostAsync(cudaStream_t stream) const {
-  auto ret = cms::cuda::make_host_unique<int32_t[]>(nDigis(), stream);
-  cms::cuda::copyAsync(ret, clus_d, nDigis(), stream);
-  return ret;
-}
-
-cms::cuda::host::unique_ptr<uint32_t[]> SiPixelDigisCUDA::pdigiToHostAsync(cudaStream_t stream) const {
-  auto ret = cms::cuda::make_host_unique<uint32_t[]>(nDigis(), stream);
-  cms::cuda::copyAsync(ret, pdigi_d, nDigis(), stream);
-  return ret;
-}
-
-cms::cuda::host::unique_ptr<uint32_t[]> SiPixelDigisCUDA::rawIdArrToHostAsync(cudaStream_t stream) const {
-  auto ret = cms::cuda::make_host_unique<uint32_t[]>(nDigis(), stream);
-  cms::cuda::copyAsync(ret, rawIdArr_d, nDigis(), stream);
+SiPixelDigisCUDA::HostStoreAndBuffer SiPixelDigisCUDA::dataToHostAsync(cudaStream_t stream) const {
+  // Allocate the needed space only and build the compact data in place in host memory (from the larger device memory).
+  // Due to the compaction with the 2D copy, we need to know the precise geometry, and hence operate on the store (as opposed
+  // to the view, which is unaware of the column pitches.
+  HostStoreAndBuffer ret(nDigis(), stream);
+  cudaCheck(cudaMemcpyAsync(ret.hostStore_.adc(), hostDeviceStore_d.adc(), nDigis_h * sizeof(decltype(*deviceFullView_.adc())),
+          cudaMemcpyDeviceToHost, stream));
+  // Copy the other columns, realigning the data in shorter arrays. clus is the first but all 3 columns (clus, pdigis, rawIdArr) have
+  // the same geometry.
+  cudaCheck(cudaMemcpy2DAsync(ret.hostStore_.clus(), ret.hostStore_.soaMetadata().clusPitch(),
+          hostDeviceStore_d.clus(), hostDeviceStore_d.soaMetadata().clusPitch(),
+          3 /* rows */,
+          nDigis() * sizeof(decltype (*ret.hostStore_.clus())), cudaMemcpyDeviceToHost, stream));
   return ret;
 }
--- a/src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
+++ b/src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
@@ -6,10 +6,62 @@
 #include "CUDACore/cudaCompat.h"
 #include "CUDACore/device_unique_ptr.h"
 #include "CUDACore/host_unique_ptr.h"
+#include "DataFormats/SoAStore.h"
+#include "DataFormats/SoAView.h"
 
 class SiPixelDigisCUDA {
 public:
-  SiPixelDigisCUDA() = default;
+  generate_SoA_store(DeviceOnlyStore,
+    /* These are consumed by downstream device code                                                   */
+    SoA_column(uint16_t, xx),         /* local coordinates of each pixel                              */
+    SoA_column(uint16_t, yy),         /*                                                              */
+    SoA_column(uint16_t, moduleInd)   /* module id of each pixel                                      */
+  );
+  
+  generate_SoA_store(HostDeviceStore,
+    /* These are also transferred to host (see HostDataView) */
+    SoA_column(uint16_t, adc),        /* ADC of each pixel                                            */
+    SoA_column(int32_t, clus),        /* cluster id of each pixel                                     */
+    /* These are for CPU output; should we (eventually) place them to a                               */
+    /* separate product?                                                                              */
+    SoA_column(uint32_t, pdigi),     /* packed digi (row, col, adc) of each pixel                     */
+    SoA_column(uint32_t, rawIdArr)   /* DetId of each pixel                                           */
+  );
+  
+  generate_SoA_view(DeviceFullView,
+    SoA_view_store_list(
+      SoA_view_store(DeviceOnlyStore, deviceOnly),
+      SoA_view_store(HostDeviceStore, hostDevice)
+    ),
+    SoA_view_value_list(
+      SoA_view_value(deviceOnly, xx, xx),    /* local coordinates of each pixel                       */
+      SoA_view_value(deviceOnly, yy, yy),    /*                                                       */
+      SoA_view_value(deviceOnly, moduleInd, moduleInd),  /* module id of each pixel                   */
+      SoA_view_value(hostDevice, adc, adc),  /* ADC of each pixel                                     */
+      SoA_view_value(hostDevice, clus, clus),/* cluster id of each pixel                              */
+      SoA_view_value(hostDevice, pdigi, pdigi), /* packed digi (row, col, adc) of each pixel          */
+      SoA_view_value(hostDevice, rawIdArr, rawIdArr)  /* DetId of each pixel                          */
+      /* TODO: simple, no rename interface */
+    )    
+  );
+
+  /* Device pixel view: this is a second generation view (view from view) */
+  generate_SoA_const_view(DevicePixelView,
+    /* We get out data from the DeviceFullStore */
+    SoA_view_store_list(
+      SoA_view_store(DeviceFullView, deviceFullView)
+    ),
+    /* These are consumed by downstream device code                                                   */
+    SoA_view_value_list(
+      SoA_view_value(deviceFullView, xx, xx),    /* local coordinates of each pixel                   */
+      SoA_view_value(deviceFullView, yy, yy),    /*                                                   */
+      SoA_view_value(deviceFullView, moduleInd, moduleInd),  /* module id of each pixel               */
+      SoA_view_value(deviceFullView, adc, adc),  /* ADC of each pixel                                 */
+      SoA_view_value(deviceFullView, clus, clus) /* cluster id of each pixel                          */
+    )
+  );
+
+  explicit SiPixelDigisCUDA();
   explicit SiPixelDigisCUDA(size_t maxFedWords, cudaStream_t stream);
   ~SiPixelDigisCUDA() = default;
 
@@ -26,58 +78,47 @@ public:
   uint32_t nModules() const { return nModules_h; }
   uint32_t nDigis() const { return nDigis_h; }
 
-  uint16_t *xx() { return xx_d.get(); }
-  uint16_t *yy() { return yy_d.get(); }
-  uint16_t *adc() { return adc_d.get(); }
-  uint16_t *moduleInd() { return moduleInd_d.get(); }
-  int32_t *clus() { return clus_d.get(); }
-  uint32_t *pdigi() { return pdigi_d.get(); }
-  uint32_t *rawIdArr() { return rawIdArr_d.get(); }
+  uint16_t *xx() { return deviceFullView_.xx(); }
+  uint16_t *yy() { return deviceFullView_.yy(); }
+  uint16_t *adc() { return deviceFullView_.adc(); }
+  uint16_t *moduleInd() { return deviceFullView_.moduleInd(); }
+  int32_t *clus() { return deviceFullView_.clus(); }
+  uint32_t *pdigi() { return deviceFullView_.pdigi(); }
+  uint32_t *rawIdArr() { return deviceFullView_.rawIdArr(); }
 
-  uint16_t const *xx() const { return xx_d.get(); }
-  uint16_t const *yy() const { return yy_d.get(); }
-  uint16_t const *adc() const { return adc_d.get(); }
-  uint16_t const *moduleInd() const { return moduleInd_d.get(); }
-  int32_t const *clus() const { return clus_d.get(); }
-  uint32_t const *pdigi() const { return pdigi_d.get(); }
-  uint32_t const *rawIdArr() const { return rawIdArr_d.get(); }
+  uint16_t const *xx() const { return deviceFullView_.xx(); }
+  uint16_t const *yy() const { return deviceFullView_.yy(); }
+  uint16_t const *adc() const { return deviceFullView_.adc(); }
+  uint16_t const *moduleInd() const { return deviceFullView_.moduleInd(); }
+  int32_t const *clus() const { return deviceFullView_.clus(); }
+  uint32_t const *pdigi() const { return deviceFullView_.pdigi(); }
+  uint32_t const *rawIdArr() const { return deviceFullView_.rawIdArr(); }
 
-  cms::cuda::host::unique_ptr<uint16_t[]> adcToHostAsync(cudaStream_t stream) const;
-  cms::cuda::host::unique_ptr<int32_t[]> clusToHostAsync(cudaStream_t stream) const;
-  cms::cuda::host::unique_ptr<uint32_t[]> pdigiToHostAsync(cudaStream_t stream) const;
-  cms::cuda::host::unique_ptr<uint32_t[]> rawIdArrToHostAsync(cudaStream_t stream) const;
-
-  class DeviceConstView {
+  class HostStoreAndBuffer {
+    friend SiPixelDigisCUDA;
   public:
-    __device__ __forceinline__ uint16_t xx(int i) const { return __ldg(xx_ + i); }
-    __device__ __forceinline__ uint16_t yy(int i) const { return __ldg(yy_ + i); }
-    __device__ __forceinline__ uint16_t adc(int i) const { return __ldg(adc_ + i); }
-    __device__ __forceinline__ uint16_t moduleInd(int i) const { return __ldg(moduleInd_ + i); }
-    __device__ __forceinline__ int32_t clus(int i) const { return __ldg(clus_ + i); }
-
-    uint16_t const *xx_;
-    uint16_t const *yy_;
-    uint16_t const *adc_;
-    uint16_t const *moduleInd_;
-    int32_t const *clus_;
+    HostStoreAndBuffer();
+    const SiPixelDigisCUDA::HostDeviceStore store() { return hostStore_; }
+    void reset();
+  private:
+    HostStoreAndBuffer(size_t maxFedWords, cudaStream_t stream);
+    cms::cuda::host::unique_ptr<std::byte[]> data_h;
+    HostDeviceStore hostStore_;
   };
+  HostStoreAndBuffer dataToHostAsync(cudaStream_t stream) const;
 
-  const DeviceConstView *view() const { return view_d.get(); }
+   // Special copy for validation
+   cms::cuda::host::unique_ptr<uint16_t[]> adcToHostAsync(cudaStream_t stream) const;
+
+  const DevicePixelView& pixelView() const { return devicePixelView_; }
 
 private:
   // These are consumed by downstream device code
-  cms::cuda::device::unique_ptr<uint16_t[]> xx_d;         // local coordinates of each pixel
-  cms::cuda::device::unique_ptr<uint16_t[]> yy_d;         //
-  cms::cuda::device::unique_ptr<uint16_t[]> adc_d;        // ADC of each pixel
-  cms::cuda::device::unique_ptr<uint16_t[]> moduleInd_d;  // module id of each pixel
-  cms::cuda::device::unique_ptr<int32_t[]> clus_d;        // cluster id of each pixel
-  cms::cuda::device::unique_ptr<DeviceConstView> view_d;  // "me" pointer
-
-  // These are for CPU output; should we (eventually) place them to a
-  // separate product?
-  cms::cuda::device::unique_ptr<uint32_t[]> pdigi_d;     // packed digi (row, col, adc) of each pixel
-  cms::cuda::device::unique_ptr<uint32_t[]> rawIdArr_d;  // DetId of each pixel
-
+  cms::cuda::device::unique_ptr<std::byte[]> data_d;      // Single SoA storage
+  DeviceOnlyStore deviceOnlyStore_d;
+  HostDeviceStore hostDeviceStore_d;
+  DeviceFullView deviceFullView_;
+  DevicePixelView devicePixelView_;
   uint32_t nModules_h = 0;
   uint32_t nDigis_h = 0;
 };
--- a/src/cudadev/CUDADataFormats/TrackingRecHit2DHostSOAView.h
+++ b/src/cudadev/CUDADataFormats/TrackingRecHit2DHostSOAView.h
@@ -1,0 +1,26 @@
+
+#ifndef CUDADataFormats_TrackingRecHit_interface_TrackingRecHit2DSOAHostView_h
+#define CUDADataFormats_TrackingRecHit_interface_TrackingRecHit2DSOAHostView_h
+
+#include "CUDADataFormats/TrackingRecHit2DSOAView.h"
+#include "CUDACore/host_unique_ptr.h"
+
+template <typename Traits>
+class TrackingRecHit2DHeterogeneous; 
+
+struct TrackingRecHit2DHostSOAView {
+  template <typename Traits>
+  friend class TrackingRecHit2DHeterogeneous;
+public:
+  TrackingRecHit2DHostSOAView();
+  void reset();
+  __device__ __forceinline__ const auto operator[](size_t i) const  { return hitsStore_[i]; }
+  __device__ __forceinline__ size_t size() { return hitsStore_.soaMetadata().size(); }
+private:
+  TrackingRecHit2DHostSOAView(size_t size, cudaStream_t stream);
+  cms::cuda::host::unique_ptr<std::byte[]> hits_h;
+  TrackingRecHit2DSOAStore::HitsStore hitsStore_;
+};
+
+
+#endif // ndef CUDADataFormats_TrackingRecHit_interface_TrackingRecHit2DSOAHostView_h
--- a/src/cudadev/CUDADataFormats/TrackingRecHit2DSOAView.h
+++ b/src/cudadev/CUDADataFormats/TrackingRecHit2DSOAView.h
@@ -7,12 +7,14 @@
 #include "CUDACore/HistoContainer.h"
 #include "CUDACore/cudaCompat.h"
 #include "Geometry/phase1PixelTopology.h"
+#include "DataFormats/SoAStore.h"
+#include "DataFormats/SoAView.h"
 
 namespace pixelCPEforGPU {
   struct ParamsOnGPU;
 }
 
-class TrackingRecHit2DSOAView {
+class TrackingRecHit2DSOAStore {
 public:
   using hindex_type = uint32_t;  // if above is <=2^32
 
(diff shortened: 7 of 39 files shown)


src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.cc
```diff
@@ -13,43 +13,40 @@
...
 
 SiPixelROCsStatusAndMappingWrapper::SiPixelROCsStatusAndMappingWrapper(SiPixelROCsStatusAndMapping const& cablingMap,
                                                                std::vector<unsigned char> modToUnp)
     : modToUnpDefault(modToUnp.size()), hasQuality_(true) {
-  cudaCheck(cudaMallocHost(&cablingMapHost, sizeof(SiPixelROCsStatusAndMapping)));
-  std::memcpy(cablingMapHost, &cablingMap, sizeof(SiPixelROCsStatusAndMapping));
-
+  // TODO: check if cudaStreamDefault is appropriate
+  auto cablingMapMetadata = cablingMap.soaMetadata();
+  cablingMapHostBuffer = cms::cuda::make_host_unique<std::byte[]>(cablingMapMetadata.byteSize(), cudaStreamDefault);
```
makortel, line 23: Not really in principle, but it probably won't matter in practice. I'm going to rework the EventSetup side in this fall to allow the use of these smart pointers more naturally here.

src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.cc
```diff
@@ -13,43 +13,40 @@
...
+  );
   return data.cablingMapDevice;
 }
 
 const unsigned char* SiPixelROCsStatusAndMappingWrapper::getModToUnpAllAsync(cudaStream_t cudaStream) const {
   const auto& data =
       modToUnp_.dataForCurrentDeviceAsync(cudaStream, [this](ModulesToUnpack& data, cudaStream_t stream) {
-        cudaCheck(cudaMalloc((void**)&data.modToUnpDefault, pixelgpudetails::MAX_SIZE_BYTE_BOOL));
-        cudaCheck(cudaMemcpyAsync(data.modToUnpDefault,
+        data.modToUnpDefault = cms::cuda::make_device_unique<unsigned char []>(pixelgpudetails::MAX_SIZE_BYTE_BOOL, stream);
```
makortel, line 44: Also here the use of `stream` for memory allocations is not, in principle, appropriate, but it probably doesn't matter in practice as the ES product outlives the event loop and all GPU work.

src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.h
```diff
@@ -28,17 +28,21 @@ class SiPixelROCsStatusAndMappingWrapper {
...
-  SiPixelROCsStatusAndMapping *cablingMapHost = nullptr;  // pointer to struct in CPU
+  cms::cuda::host::unique_ptr<std::byte[]> cablingMapHostBuffer;  // host pined memory for cabling map.
 
   struct GPUData {
-    ~GPUData();
-    SiPixelROCsStatusAndMapping *cablingMapDevice = nullptr;  // pointer to struct in GPU
+    void allocate(size_t size, cudaStream_t stream) {
+      cablingMapDeviceBuffer = cms::cuda::make_device_unique<std::byte[]>(
+              SiPixelROCsStatusAndMapping::computeDataSize(size), stream);
+      new(&cablingMapDevice) SiPixelROCsStatusAndMapping(cablingMapDeviceBuffer.get(), size);
```
makortel, line 37: Should the destructor of earlier `calingMapDevice` object be called first?

src/cudadev/Makefile
```diff
@@ -47,7 +47,8 @@ $(1)_CUSRC := $$(wildcard $(TARGET_DIR)/plugin-$(1)/*.cu)
 $(1)_OBJ := $$(patsubst $(SRC_DIR)%,$(OBJ_DIR)%,$$($(1)_SRC:%=%.o))
 $(1)_CUOBJ := $$(patsubst $(SRC_DIR)%,$(OBJ_DIR)%,$$($(1)_CUSRC:%=%.o))
 $(1)_DEP := $$($(1)_OBJ:$.o=$.d)
-ALL_DEPENDS += $$($(1)_DEP)
+$(1)_CUDEP := $$($(1)_CUOBJ:$.o=$.d)
+ALL_DEPENDS += $$($(1)_DEP) $$($(1)_CUDEP)
```
makortel, line 51: Good catch! Could you separate the changes in the `Makefile` into a separate PR that could be integrated more quickly? This fix should be propagated to the `Makefile`s of the other programs as well.

src/cudadev/plugin-SiPixelClusterizer/SiPixelRawToClusterGPUKernel.cu
```diff
@@ -334,7 +329,7 @@ namespace pixelgpudetails {
   }
 
   // Kernel to perform Raw to Digi conversion
-  __global__ void RawToDigi_kernel(const SiPixelROCsStatusAndMapping *cablingMap,
+  __global__ void RawToDigi_kernel(const SiPixelROCsStatusAndMapping cablingMap,
```
makortel, line 332: What would happen if the `const` would be missing along
```suggestion
  __global__ void RawToDigi_kernel(SiPixelROCsStatusAndMapping cablingMap,
```
? Would the function still get only read access to the memory, or full read-write access?

makortel (commented):
Thanks, very interesting! Here are my first comments, based on how the SoA is used. I'll need another round to look for the implementation.

src/cudadev/plugin-SiPixelClusterizer/SiPixelROCsStatusAndMappingWrapperESProducer.cc
```diff
@@ -30,8 +30,10 @@ void SiPixelROCsStatusAndMappingWrapperESProducer::produce(edm::EventSetup& even
   {
     std::ifstream in(data_ / "cablingMap.bin", std::ios::binary);
     in.exceptions(std::ifstream::badbit | std::ifstream::failbit | std::ifstream::eofbit);
-    SiPixelROCsStatusAndMapping obj;
-    in.read(reinterpret_cast<char*>(&obj), sizeof(SiPixelROCsStatusAndMapping));
+    // We use default alignment
+    auto objBuffer = std::make_unique<std::byte[]>(SiPixelROCsStatusAndMapping::computeDataSize(pixelgpudetails::MAX_SIZE));
+    SiPixelROCsStatusAndMapping obj(objBuffer.get(), pixelgpudetails::MAX_SIZE);
```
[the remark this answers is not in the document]
makortel: On a further thought, in the memory ownership model we've mainly discussing about there is a distinction between an owner object and a non-owning "view". I realize the SoA here looks a lot like such non-owning view, in which case leaving the memory management outside of the SoA is completely understandable. But I wonder if it would be possible to craft a generic owner class in this case?

src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.cc
ericcano: OK, so leaving is as-is, thanks.

src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.cc
ericcano: Also leaving like this, then.

src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.h
ericcano: Not in this case, as the SoA is just a non-owning view with nothing to destroy, but I agree it's a good practice before an in-place constructor, and in case the class evolves. I will add it.

src/cudadev/Makefile
ericcano: Now in https://github.com/cms-patatrack/pixeltrack-standalone/pull/217

src/cudadev/plugin-SiPixelClusterizer/SiPixelROCsStatusAndMappingWrapperESProducer.cc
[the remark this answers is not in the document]
ericcano: I think this is an important place to look. Currently, while running `cudadev` with those parameters: `./cudadev --maxEvents 10000 --numberOfThreads 16 --numberOfStreams 24`, we can see in NSight that CPU threads spend most of their time waiting for the mutex of the caching memory allocator. The GPU and copy engine are not saturated. 

A single memory handler per producer allocating a single chuck of memory and cutting it into one or multiple views should reduce the number of allocations by a factor of a few units (up to 8 here: https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc#L8-L15 ). I am planning to look into this now as there is a good performance gain opportunity here.

Eventually, this memory manager could also handle the various copies in host, pinned, host and devices memory, and the transfers.

src/cudadev/plugin-SiPixelClusterizer/SiPixelRawToClusterGPUKernel.cu
ericcano: The function would get read-write access to the memory. I actually see an improvement here where we can add `__restrict__` references  in the const use case and get the data cached through the non-coherent cache. I will look into this.

src/cudadev/plugin-SiPixelClusterizer/SiPixelROCsStatusAndMappingWrapperESProducer.cc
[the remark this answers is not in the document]
makortel: > Currently, while running `cudadev` with those parameters: `./cudadev --maxEvents 10000 --numberOfThreads 16 --numberOfStreams 24`, we can see in NSight that CPU threads spend most of their time waiting for the mutex of the caching memory allocator.

Is the bottleneck really the mutex in the caching allocator? The last time I looked into that (which admittedly was nearly 2 years ago and in CMSSW) the CPU-side bottleneck was the mutex in the CUDA itself. Things have certainly evolved since then, but I'd like to see the evidence.

What GPU are you testing on? In my tests on V100 `cudadev` saturates around `--numberOfThreads 10`, so at least there going much beyond will just result in more lock contention. Luckily(?) this setup also reaches 92-98 % GPU utilization as reported by `nvidia-smi` (whatever that really means).

src/cudadev/plugin-SiPixelClusterizer/SiPixelRawToClusterGPUKernel.cu
makortel: I'm more worried that it is too easy (e.g. by mistake) to get a read-write access into something that should be treated as `const`. This is crucial for all consumers of Event or EventSetup data products, the constness of the data must be preserved (which is why there is `DeviceConstView` for the consumers e.g. here 
https://github.com/cms-patatrack/pixeltrack-standalone/blob/bb3fff74dbf8b34c6bd8ea7f8336e71faebce1fa/src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h#L50
)

src/cudadev/plugin-SiPixelClusterizer/SiPixelROCsStatusAndMappingWrapperESProducer.cc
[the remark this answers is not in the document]
ericcano: I had a look with NSight. Here is a snapshot:

image

This is with a T4 (with 2 Intel(R) Xeon(R) Gold 6338 for a total of 64 cores/128 threads). NSight saw 23 threads in total, so CPU is not used at the maximum. The very bottom blue/purple bars show the usage of the GPU processing and copy engine, respectively. We are clearly not saturating the GPU (~50%, judging by eye).

It's easy to differentiate the CUDA locking from the memory manager locking: CUDA uses a pthread_rwlock_wrlock and NSight shows the CUDA call under the system call (see the red `cudaMem...` on the 4th thread on the left), while the memory manager locking uses pthread_mutex_lock. 

All can be checked with a tooltip stack trace:

image

I tried to minimize the locking in the memory manager by deferring potentially expensive calls our of the critical section, but this did not help. One of such expensive calls can be visualized here:
image

We can see all other threads waiting for the mutex in this snapshot.

I added deferred calls here: https://github.com/ericcano/pixeltrack-standalone/commit/40abde015f3b51aebb71b6296724c3e294cd8519 that's a risk free change, but it did not help much.

I also moved the cudaEventRecord out of the critical section, but this requires to change to a 2 pass strategy, which actually reduces performance (probably adding to the contention as each pass requires a lock): https://github.com/ericcano/pixeltrack-standalone/commit/a3fdba0e873bced08e3e8bd37d9bbca6de1ee8de

So, in all, reducing the number of allocations by gathering SoA columns in a single block should improve the performance (looking into this now).

src/cudadev/plugin-SiPixelClusterizer/SiPixelRawToClusterGPUKernel.cu
ericcano: I see. Const correctness should achieve this, but I will look into having an explicitly const version of the SoA structures.

src/cudadev/CUDACore/CachingDeviceAllocator.h
```diff
@@ -445,17 +455,17 @@ namespace notcub {
...
             // on the current device
 
-            // Free device memory and destroy stream event.
-            // CMS: silently ignore errors and pass them to the caller
-            if ((error = cudaFree(block_itr->d_ptr)))
-              break;
-            if ((error = cudaEventDestroy(block_itr->ready_event)))
-              break;
+            // CMS: Free device memory and destroy stream event later.
+            toFree.push_back(BlockAndEvent{block_itr->d_ptr, block_itr->ready_event});
```
makortel, line 468: Note that this happens only if `cudaMalloc()` fails because of no free memory, i.e. it should not happen normally.

src/cudadev/CUDACore/CachingDeviceAllocator.h
```diff
@@ -689,12 +714,8 @@ namespace notcub {
...
         }
 
-        // Free device memory
-        // CMS: silently ignore errors and pass them to the caller
-        if ((error = cudaFree(begin->d_ptr)))
-          break;
-        if ((error = cudaEventDestroy(begin->ready_event)))
-          break;
+        // CMS: Actually free device memory later
+        toFree.push_back(BlockAndEvent{begin->d_ptr, begin->ready_event});
```
makortel, line 718: Note that the `FreeAllCached()` is called only at the end of the job.

src/cudadev/plugin-SiPixelClusterizer/SiPixelROCsStatusAndMappingWrapperESProducer.cc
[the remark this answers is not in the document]
makortel: This is very interesting. Have you tried what happens with smaller number of CPU threads, say between 5 and 10? I'd expect T4 to saturate with at most the same amount of CPU threads as V100, so I'd expect 23 to suffer heavily from lock contention.

Nevertheless I agree it is clear that the caching allocator must be improved. Given also the discussion in #216, I opened an issue to discuss that further https://github.com/cms-patatrack/pixeltrack-standalone/issues/218.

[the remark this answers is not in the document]
ericcano: I moved the relevant code out to another branch: https://github.com/ericcano/pixeltrack-standalone/tree/memoryManagement

ericcano:
The latest rewrite of the branch fixes the "one has have to know what he/she's doing" approach where views were overlaid on one another to provide multiple sets of columns depending on the use pattern and transfer needs.

The new approach uses a 3 layers approach with:
- Buffer: a buffer of std::byte, allocated on host or device
- Store: a collection of contiguous, aligned columns of data of same length
- View: a collection of pointers to columns from one or several stores, adjacent or not.

At typical example can be see in https://github.com/cms-patatrack/pixeltrack-standalone/pull/211/commits/3e9b8fbb3094854d96f4887d4d2c9b81d5171614#diff-a84742dff2728673b16912f2e5297015c8c9d47ab27a616fc757d450f2d01808, where a store is used on both the host and the device, and the data gets transferred between the 2, but in addition, extra columns are present on the device side only, and located in a second store. Both stores use the same buffer to save memory allocation, and a view merges both together in a logical view to provide unified access.

Utility function allow easy computation of store sizes (`computeDataSize(size_t length)`) and locating stores position in shared buffers (`nextByte()`).

4 examples where migrated to this scheme (1 condition structure and 3 data structures).

Views do not support scalar (non-column) elements yet, but those are not so common (to be added though).

Support for Eigen columns in views could be added too.

Finally, views could be templated to add cache optimization flavors (non-coherent cache, streamed reads and writes).

src/cudadev/CUDACore/device_unique_ptr.h
```diff
@@ -49,8 +49,8 @@ namespace cms {
 
     template <typename T>
     typename device::impl::make_device_unique_selector<T>::non_array make_device_unique(cudaStream_t stream) {
-      static_assert(std::is_trivially_constructible<T>::value,
-                    "Allocating with non-trivial constructor on the device memory is not supported");
+      static_assert(std::is_trivially_copyable<T>::value,
+                    "Allocating with non-trivial copy on the device memory is not supported");
```
makortel, line 53: Maybe it is time to give up on "trivially constructible" requirement, it was somewhat annoying anyway.

I'm somewhat concerned on clarity that this function doesn't initialize the memory (with "trivially constructible" that's somewhat expected, although technically different from "value initialization" of `std::make_unique()`), but that's not a real issue.

With this change I would consider removing `make_device_unique_uninitialized()` (that really should have had the "trivially copyable" requirement).

src/cudadev/CUDACore/host_unique_ptr.h
```diff
@@ -39,17 +39,17 @@ namespace cms {
     // Allocate pinned host memory
     template <typename T>
     typename host::impl::make_host_unique_selector<T>::non_array make_host_unique(cudaStream_t stream) {
-      static_assert(std::is_trivially_constructible<T>::value,
-                    "Allocating with non-trivial constructor on the pinned host memory is not supported");
+      static_assert(std::is_trivially_copyable<T>::value,
+                    "Allocating with non-trivial copy on the pinned host memory is not supported");
```
makortel, line 43: This raises the question if `make_host_unique()` should value-initialize like `std::make_unique()`, Or if there should be a separate `make` function that does also the value initialization.

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
```diff
@@ -22,41 +32,21 @@ class SiPixelClustersCUDA {
...
-    uint32_t const *moduleId_;
-    uint32_t const *clusModuleStart_;
-  };
+  uint32_t const *moduleStart() const { return deviceStore_.moduleStart(); }
+  uint32_t const *clusInModule() const { return deviceStore_.clusInModule(); }
+  uint32_t const *moduleId() const { return deviceStore_.moduleId(); }
+  uint32_t const *clusModuleStart() const { return deviceStore_.clusModuleStart(); }
 
-  DeviceConstView *view() const { return view_d.get(); }
+  const DeviceStore store() const { return deviceStore_; }
```
makortel, line 45: Does `DeviceStore` give read-write access to the data? Assuming yes, in CMSSW this function would not be allowed (and thus I'd like to avoid it here too) because of giving mutable access to Event data product from a `const` function.

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc
```diff
@@ -5,44 +5,54 @@
...
+{}
+
+SiPixelDigisCUDA::HostStoreAndBuffer::HostStoreAndBuffer(size_t maxFedWords, cudaStream_t stream)
+     : data_h(cms::cuda::make_host_unique<std::byte[]>(SiPixelDigisCUDA::HostDeviceStore::computeDataSize(maxFedWords), stream)),
+       hostStore_(data_h.get(), maxFedWords)
+{}
+
+void SiPixelDigisCUDA::HostStoreAndBuffer::reset() {
+  hostStore_.~HostDeviceStore();
+  new(&hostStore_) HostDeviceStore(nullptr, 0);
```
makortel, line 33: Would 
```suggestion
  hostStore_ = HostDeviceStore(nullptr, 0);
```
work?

makortel (commented):
Got to around `SiPixelDigisCUDA.cc`.

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
```diff
@@ -26,60 +78,49 @@ class SiPixelDigisCUDA {
...
-    __device__ __forceinline__ uint16_t moduleInd(int i) const { return __ldg(moduleInd_ + i); }
-    __device__ __forceinline__ int32_t clus(int i) const { return __ldg(clus_ + i); }
-
-    uint16_t const *xx_;
-    uint16_t const *yy_;
-    uint16_t const *adc_;
-    uint16_t const *moduleInd_;
-    int32_t const *clus_;
+    HostStoreAndBuffer();
+    const SiPixelDigisCUDA::HostDeviceStore store() { return hostStore_; }
```
makortel, line 101: I'd suggest to remove the `const` because it doesn't really enforce anything here
```suggestion
    SiPixelDigisCUDA::HostDeviceStore store() { return hostStore_; }
```

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
```diff
@@ -6,10 +6,62 @@
...
+      SoA_view_value(hostDevice, adc, adc),  /* ADC of each pixel                                     */
+      SoA_view_value(hostDevice, clus, clus),/* cluster id of each pixel                              */
+      SoA_view_value(hostDevice, pdigi, pdigi), /* packed digi (row, col, adc) of each pixel          */
+      SoA_view_value(hostDevice, rawIdArr, rawIdArr)  /* DetId of each pixel                          */
+      /* TODO: simple, no rename interface */
+    )    
+  );
+
+  /* Device pixel view: this is a second generation view (view from view) */
+  generate_SoA_const_view(DevicePixelView,
```
makortel, line 49: Does the `DevicePixelView` have `const`-only interface? Including `Const` somewhere in the class name would make that clear in places where the class name is used.

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc
```diff
@@ -5,44 +5,54 @@
...
-  view->yy_ = yy_d.get();
-  view->adc_ = adc_d.get();
-  view->moduleInd_ = moduleInd_d.get();
-  view->clus_ = clus_d.get();
-
-  cms::cuda::copyAsync(view_d, view, stream);
-}
+    : data_d(cms::cuda::make_device_unique<std::byte[]>(
+        DeviceOnlyStore::computeDataSize(maxFedWords) + 
+        HostDeviceStore::computeDataSize(maxFedWords),
```
makortel, line 10: Is the return value of `computeDataSize()` esseitnally rounded up to the next integer multiple of alignment bytes?

src/cudadev/CUDADataFormats/TrackingRecHit2DHostSOAView.h
```diff
@@ -0,0 +1,26 @@
...
+  template <typename Traits>
+  friend class TrackingRecHit2DHeterogeneous;
+public:
+  TrackingRecHit2DHostSOAView();
+  void reset();
+  __device__ __forceinline__ const auto operator[](size_t i) const  { return hitsStore_[i]; }
+  __device__ __forceinline__ size_t size() { return hitsStore_.soaMetadata().size(); }
+private:
+  TrackingRecHit2DHostSOAView(size_t size, cudaStream_t stream);
+  cms::cuda::host::unique_ptr<std::byte[]> hits_h;
```
makortel, line 21: Given the earlier use of `View` to denote types that do not own the memory buffer, I find the nomenclature here confusing.

src/cudadev/CUDADataFormats/TrackingRecHit2DSOAView.h
```diff
@@ -22,40 +24,84 @@ class TrackingRecHit2DSOAView {
...
+      SoA_view_value(supportObjectsStore, iphi, iphi),
+      SoA_view_value(supportObjectsStore, detectorIndex, detectorIndex)
+    )
+  );
+  
+  // Shortcut operator saving the explicit calls to view in usage.
+  __device__ __forceinline__ HitsAndSupportView::element operator[] (size_t index) { 
+    return m_hitsAndSupportView[index]; 
+  }
+  __device__ __forceinline__ const HitsAndSupportView::const_element operator[] (size_t index) const {
```
makortel, line 101:
```suggestion
  __device__ __forceinline__ HitsAndSupportView::const_element operator[] (size_t index) const { 
```
?

makortel (commented):
Got up to `TrackingRecHit2DSOAView.h`.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+  struct SoAMetadata {                                                                                                              \
+    friend CLASS;                                                                                                                   \
+    SOA_HOST_DEVICE_INLINE size_t size() const { return parent_.nElements_; }                                                       \
+    SOA_HOST_DEVICE_INLINE size_t byteSize() const { return parent_.byteSize_; }                                                    \
+    SOA_HOST_DEVICE_INLINE size_t byteAlignment() const { return parent_.byteAlignment_; }                                          \
+    SOA_HOST_DEVICE_INLINE std::byte* data() const { return parent_.mem_; }                                                         \
+    SOA_HOST_DEVICE_INLINE std::byte* nextByte() const { return parent_.mem_ + parent_.byteSize_; }                                 \
+    SOA_HOST_DEVICE_INLINE CLASS cloneToNewAddress(std::byte* addr) {                                                               \
```
makortel, line 461: This function looks like it could be `const`
```suggestion
    SOA_HOST_DEVICE_INLINE CLASS cloneToNewAddress(std::byte* addr) const {                                                               \
```

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+    _ITERATE_ON_ALL(_DEFINE_METADATA_MEMBERS, ~, __VA_ARGS__)                                                                       \
+                                                                                                                                    \
+  private:                                                                                                                          \
+    SOA_HOST_DEVICE_INLINE SoAMetadata(const CLASS& parent): parent_(parent) {}                                                     \
+    const CLASS& parent_;                                                                                                           \
+  };                                                                                                                                \
+  friend SoAMetadata;                                                                                                               \
+  SOA_HOST_DEVICE_INLINE const SoAMetadata soaMetadata() const { return  SoAMetadata(*this); }                                      \
```
makortel, line 471: Confusing `const`
```suggestion
  SOA_HOST_DEVICE_INLINE SoAMetadata soaMetadata() const { return  SoAMetadata(*this); }                                      \
```

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+    if(mem_ + byteSize_ != curMem)                                                                                                  \
+      throw std::out_of_range("In " #CLASS "::" #CLASS ": unexpected end pointer.");                                                \
+  }                                                                                                                                 \
+                                                                                                                                    \
+  /* Constructor relying on user provided storage */                                                                                \
+  SOA_DEVICE_ONLY CLASS(bool devConstructor, std::byte* mem, size_t nElements, size_t byteAlignment = defaultAlignment):            \
+      mem_(mem), nElements_(nElements), byteAlignment_(byteAlignment) {                                                             \
+    auto curMem = mem_;                                                                                                             \
```
makortel, line 492: Is `curMem` used somewhere?

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+    rangeCheck(index);                                                                                                              \
+    return element(index,                                                                                                           \
+        _ITERATE_ON_ALL_COMMA(_DECLARE_ELEMENT_CONSTR_CALL, ~, __VA_ARGS__) );                                                      \
+  }                                                                                                                                 \
+                                                                                                                                    \
+  /* AoS-like accessor (const) */                                                                                                   \
+  SOA_HOST_DEVICE_INLINE                                                                                                            \
+  const const_element operator[](size_t index) const {                                                                              \
```
makortel, line 536: Confusing `const`
```suggestion
  const_element operator[](size_t index) const {                                                                              \
```

makortel (commented):
Got to `SoAStore.h`. I might have easily misunderstood something.

src/cudadev/CUDACore/device_unique_ptr.h
ericcano: `make_device_unique_uninitialized()` is not used (neither in any variant of `pixeltrack standalone, nor in CMSSW (I checked `CMSSW_12_1_0_pre5`).

Both changes are fine in all variations of pixeltrack standalone: https://github.com/ericcano/pixeltrack-standalone/tree/device_unique_ptr_relax (should I PR this?)

src/cudadev/CUDACore/host_unique_ptr.h
ericcano: In the current SoA examples, we are unaffected as we do a 2 stage allocation where `make_(host|device)_unique()` are used to provide a bare buffer which will typically be filled by a ES/ED producer. Actually, this is the most common use case in general.

We can work this around with placement new, but I'm not sure we have so many use cases. It also breaks some symmetry with the device memory (Although this is less of an issue as we tend to a model where CPU is just another device).

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
ericcano: The `DeviceStore` is mutable, but it is `const` correct. There is an `operator[] (size_t) const` returning a different accessor type (`const_element`). Likewise, the full column pointer accessors (`[const] uint32_t * DeviceStore::x() [const]`) exist in 2 variants and are const-correct.

We can also make a stripped-down version of the store but the compiler should achieve the same with `const`correctness. The framework could even enforce returning const variants with `std::is_const`.

Should we still define a stripped down `DeviceConstStore` ? (I did if for views, but it's redundant to `const` views).

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc
ericcano: Indeed, it does. Thanks.

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
ericcano: Unless I miss something, it does. Data access functions have const/non-const variants. I made a macro expansion of this class here: https://gist.github.com/ericcano/1d10e0902829408a0d8815e99567102c

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
ericcano: Sure. I did it,

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.cc
ericcano: A bit more. It computes the sizes of each column, which have to be individually aligned and sums the sizes. On example can be found here: https://gist.github.com/ericcano/1d10e0902829408a0d8815e99567102c#file-cc-L19.

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
makortel: I'm afraid `DeviceStore` being `const`-correct is not sufficient, because the following is possible
```cpp
SiPixelClustersCUDA const& clusters = ctx.get(iEvent, ...);
DeviceStore mutableStore = clusters.store();
// 'mutableStore' "leaks" mutable access to the memory buffer owned by `clusters`
```

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
makortel: I can think of it preventing e.g. the following
```cpp
SiPixelDigisCUDA const& digis = ctx.get(iEvent, ...);
auto hostData = digis.dataToHostAsync(ctx.stream());
hostData.store().adc(0) = 42;
```
but it allows
```cpp
SiPixelDigisCUDA const& digis = ctx.get(iEvent, ...);
auto hostData = digis.dataToHostAsync(ctx.stream());
SiPixelDigisCUDA::HostDeviceStore store = hostData.store();
store.adc(0) = 42;
```

On one hand in this case the function itself is not `const`, so returning non-`const` would be fine.

ericcano: Thanks, I tried to turn it around, but indeed, we have no way to prevent this while retaining the trivial copyability, which is not only nice but also probably needed to pass view/store as arguments to a kernel (they get copied then).
So we do have to create a stripped down, const-only flavor of the store. Having a one-way copy constructor (non-const -> const) might simplify the coding and should not be too much hassle as it's few lines in the macros (and this would apply also to view).

src/cudadev/DataFormats/SoAView.h
```diff
@@ -0,0 +1,375 @@
...
+  BOOST_PP_EXPAND(_DECLARE_VIEW_STORE_TYPE_ALIAS_IMPL TYPE_NAME)
+
+/**
+ * Member types aliasing for referencing by name
+ */
+#define _DECLARE_VIEW_MEMBER_TYPE_ALIAS_IMPL(STORE_NAME, STORE_MEMBER, LOCAL_NAME, DATA)                                            \
+  typedef BOOST_PP_CAT(TypeOf_, STORE_NAME) :: SoAMetadata:: BOOST_PP_CAT(TypeOf_, STORE_MEMBER) BOOST_PP_CAT(TypeOf_, LOCAL_NAME); \
+  static const SoAColumnType BOOST_PP_CAT(ColumnTypeOf_, LOCAL_NAME) =                                                              \
+    BOOST_PP_CAT(TypeOf_, STORE_NAME) :: SoAMetadata:: BOOST_PP_CAT(ColumnTypeOf_, STORE_MEMBER);                                   \
+  DATA BOOST_PP_CAT(TypeOf_, LOCAL_NAME) * BOOST_PP_CAT(addressOf_, LOCAL_NAME) () const  {                                         \
```
makortel, line 65: Does this return a pointer-to-non-`const` from a `const` function? (I'm not really following the macro code)

src/cudadev/DataFormats/SoAView.h
```diff
@@ -0,0 +1,375 @@
...
+    /* Alias member types to name-derived identifyer to allow simpler definitions */                                                \
+    _ITERATE_ON_ALL(_DECLARE_VIEW_MEMBER_TYPE_ALIAS, BOOST_PP_EMPTY(), VALUE_LIST)                                                  \
+  private:                                                                                                                          \
+    SOA_HOST_DEVICE_INLINE SoAMetadata(const CLASS& parent): parent_(parent) {}                                                     \
+    const CLASS& parent_;                                                                                                           \
+  };                                                                                                                                \
+  friend SoAMetadata  ;                                                                                                             \
+  SOA_HOST_DEVICE_INLINE const SoAMetadata soaMetadata() const { return  SoAMetadata(*this); }                                      \
```
makortel, line 246: Confusing `const`
```suggestion
  SOA_HOST_DEVICE_INLINE SoAMetadata soaMetadata() const { return  SoAMetadata(*this); }                                      \
```

src/cudadev/DataFormats/SoAView.h
```diff
@@ -0,0 +1,375 @@
...
+    ):                                                                                                                              \
+      _ITERATE_ON_ALL_COMMA(_DECLARE_VIEW_ELEM_MEMBER_INIT, index, VALUE_LIST)                                                      \
+       {}                                                                                                                           \
+    SOA_HOST_DEVICE_INLINE                                                                                                          \
+    element& operator=(const element& other) {                                                                                      \
+      _ITERATE_ON_ALL(_DECLARE_VIEW_ELEMENT_VALUE_COPY, ~, VALUE_LIST)                                                              \
+      return *this;                                                                                                                 \
+    }                                                                                                                               \
```
makortel, line 283: Got curious, why is the explicit assignment operator needed?

src/cudadev/DataFormats/SoAView.h
```diff
@@ -0,0 +1,375 @@
...
+  element operator[](size_t index) {                                                                                                \
+    return element(index,                                                                                                           \
+        _ITERATE_ON_ALL_COMMA(_DECLARE_VIEW_ELEMENT_CONSTR_CALL, ~, VALUE_LIST) );                                                  \
+  }                                                                                                                                 \
+                                                                                                                                    \
+  /* AoS-like accessor (const) */                                                                                                   \
+  SOA_HOST_DEVICE_INLINE                                                                                                            \
+  const const_element operator[](size_t index) const {                                                                              \
```
makortel, line 296: Confusing `const`
```suggestion
  const_element operator[](size_t index) const {                                                                              \
```

src/cudadev/DataFormats/SoAView.h
```diff
@@ -0,0 +1,375 @@
...
+    /* Alias member types to name-derived identifyer to allow simpler definitions */                                                \
+    _ITERATE_ON_ALL(_DECLARE_VIEW_MEMBER_TYPE_ALIAS, const, VALUE_LIST)                                                             \
+  private:                                                                                                                          \
+    SOA_HOST_DEVICE_INLINE SoAMetadata(const CLASS& parent): parent_(parent) {}                                                     \
+    const CLASS& parent_;                                                                                                           \
+  };                                                                                                                                \
+  friend SoAMetadata  ;                                                                                                             \
+  SOA_HOST_DEVICE_INLINE const SoAMetadata soaMetadata() const { return  SoAMetadata(*this); }                                      \
```
makortel, line 333: Confusing `const`
```suggestion
  SOA_HOST_DEVICE_INLINE SoAMetadata soaMetadata() const { return  SoAMetadata(*this); }                                      \
```

src/cudadev/DataFormats/SoAView.h
```diff
@@ -0,0 +1,375 @@
...
+                                                                                                                                    \
+  private:                                                                                                                          \
+    _ITERATE_ON_ALL(_DECLARE_VIEW_CONST_ELEMENT_VALUE_MEMBER, ~, VALUE_LIST)                                                        \
+  };                                                                                                                                \
+                                                                                                                                    \
+  /* AoS-like accessor (const) */                                                                                                   \
+  SOA_HOST_DEVICE_INLINE                                                                                                            \
+  const const_element operator[](size_t index) const {                                                                              \
```
makortel, line 360: Confusing `const`
```suggestion
  const_element operator[](size_t index) const {                                                                              \
```

src/cudadev/plugin-SiPixelRecHits/gpuPixelRecHits.h
```diff
@@ -30,14 +30,14 @@ namespace gpuPixelRecHits {
 
     auto& hits = *phits;
 
-    auto const digis = *pdigis;  // the copy is intentional!
-    auto const& clusters = *pclusters;
+    auto const digis = pdigis;  // the copy is intentional!
+    auto const& clusters = pclusters;
```
makortel, line 34: Is either of these really needed anymore when the function gets the `pdigis` and `pclusters` by value?

makortel (commented):
Got to the end.

makortel:
I've been thinking about the names. I find the name `Store` confusing too since it does not own the storage, and therefore acts more like a View. Throwing out some ideas
* `SoAStore` -> `SoAView`, `SoAView` -> `SoACombView`
   * The `SoAView` acts like a combination of columns from other "Views"
 * `SoAStore` -> `SoALayout`, keep `SoAView`

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
makortel: > So we do need the const-only views, and products should only provide those.

Let me just add that for a product class it is ok for a non-`const` function to return a mutable view (or pointer-to-mutable), as long as the product itself is not copyable (like here), or a copy constructor/assignment makes a deep copy.

src/cudadev/CondFormats/SiPixelROCsStatusAndMapping.h
```diff
@@ -11,15 +13,15 @@ namespace pixelgpudetails {
...
+generate_SoA_store(SiPixelROCsStatusAndMapping,
+  SoA_column(unsigned int, fed),
+  SoA_column(unsigned int, link),
+  SoA_column(unsigned int, roc),
+  SoA_column(unsigned int, rawId),
+  SoA_column(unsigned int, rocInDet),
+  SoA_column(unsigned int, moduleId),
+  SoA_column(unsigned char, badRocs),
+  SoA_scalar(unsigned int, size)
+);
```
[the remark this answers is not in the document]
makortel: Even then the data could be read in from a binary file with the current struct definition (that could be even made internal to the `SiPixelROCsStatusAndMappingWrapperESProducer`), and then copied to a data structure that is more efficient for reads. But as long as the bit representation stays the same, I suppose it will continue to work.

src/cudadev/plugin-SiPixelClusterizer/SiPixelRawToClusterGPUKernel.cu
ericcano: This feature is implemented in branch https://github.com/ericcano/pixeltrack-standalone/tree/macroBasedSoA-PR211-folowup. I will introduce it in a new PR and close this one.

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
ericcano: To be complete, we have both const and non-const views. Which one is provided to the user of the product will be up to the product implementation.

src/cudadev/CUDADataFormats/SiPixelDigisCUDA.h
ericcano: In the end, the store (renamed to layout in the follow up PR) is exclusively taking care of the columns layout. The access is then done via views (with const and mutable flavors).

src/cudadev/CUDADataFormats/TrackingRecHit2DSOAView.h
ericcano: Yes, that was quite redundant. I fixed it.

src/cudadev/CondFormats/SiPixelROCsStatusAndMapping.h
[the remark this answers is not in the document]
ericcano: I reverted this change, and we now have a view directly defined with a column by column constructor (this feature was added during the hackathon).

src/cudadev/CondFormats/SiPixelROCsStatusAndMappingWrapper.cc
```diff
@@ -13,44 +13,40 @@
...
-    cudaStream_t cudaStream) const {
-  const auto& data = gpuData_.dataForCurrentDeviceAsync(cudaStream, [this](GPUData& data, cudaStream_t stream) {
-    // allocate
-    cudaCheck(cudaMalloc(&data.cablingMapDevice, sizeof(SiPixelROCsStatusAndMapping)));
-
-    // transfer
-    cudaCheck(cudaMemcpyAsync(
-        data.cablingMapDevice, this->cablingMapHost, sizeof(SiPixelROCsStatusAndMapping), cudaMemcpyDefault, stream));
-  });
+const SiPixelROCsStatusAndMapping & SiPixelROCsStatusAndMappingWrapper::getGPUProductAsync(cudaStream_t cudaStream) const {
```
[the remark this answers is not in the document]
ericcano: Scalar support ins views is done the new branch (it is now the only way to access them). We currently miss the Eigen access altogether.

src/cudadev/DataFormats/SoACommon.h
```diff
@@ -0,0 +1,182 @@
...
+#define LOAD_INCOHERENT(A) *(A)
+#define LOAD_STREAMED(A)   *(A)
+#define STORE_STREAMED(A, V) *(A) = (V)
+#endif
+
+// compile-time sized SoA
+
+// Helper template managing the value within it column
+template<typename T>
+class SoAValue {
```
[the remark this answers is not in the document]
ericcano: Done in https://github.com/ericcano/pixeltrack-standalone/commit/8bcecbea062a03c4b20dc38c8cb103138dbbfd93 (namespace is `cms::soa`).

src/cudadev/DataFormats/SoACommon.h
```diff
@@ -0,0 +1,182 @@
...
+#if defined(__CUDACC__) && defined(__CUDA_ARCH__)
+// Read a pointer content via read-only (non coherent) cache.
+#define LOAD_INCOHERENT(A) __ldg(A)
+#define LOAD_STREAMED(A)   __ldcs(A)
+#define STORE_STREAMED(A, V)  __stcs(A, V)
+#else
+#define LOAD_INCOHERENT(A) *(A)
+#define LOAD_STREAMED(A)   *(A)
+#define STORE_STREAMED(A, V) *(A) = (V)
+#endif
```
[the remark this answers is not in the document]
ericcano: The intent is to use them in the future, by adding a template parameter to the SoA viewsto easily switch from one cache style to the other (and hence easily benchmark the effect).

src/cudadev/DataFormats/SoACommon.h
```diff
@@ -0,0 +1,182 @@
...
+
+enum class SoAColumnType {
+  scalar = _VALUE_TYPE_SCALAR,
+  column = _VALUE_TYPE_COLUMN,
+  eigen = _VALUE_TYPE_EIGEN_COLUMN
+};
+
+#define SoA_scalar(TYPE, NAME) (_VALUE_TYPE_SCALAR, TYPE, NAME)
+#define SoA_column(TYPE, NAME) (_VALUE_TYPE_COLUMN, TYPE, NAME)
+#define SoA_eigenColumn(TYPE, NAME) (_VALUE_TYPE_EIGEN_COLUMN, TYPE, NAME)
```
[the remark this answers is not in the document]
ericcano: This is done in the new branch (https://github.com/ericcano/pixeltrack-standalone/commit/e3fdaea96da7de578271bcf79a17edb938d9dda1)

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+   */                                                                                                                               \
+  struct SoAMetadata {                                                                                                              \
+    friend CLASS;                                                                                                                   \
+    SOA_HOST_DEVICE_INLINE size_t size() const { return parent_.nElements_; }                                                       \
+    SOA_HOST_DEVICE_INLINE size_t byteSize() const { return parent_.byteSize_; }                                                    \
+    SOA_HOST_DEVICE_INLINE size_t byteAlignment() const { return parent_.byteAlignment_; }                                          \
+    SOA_HOST_DEVICE_INLINE std::byte* data() const { return parent_.mem_; }                                                         \
+    SOA_HOST_DEVICE_INLINE std::byte* nextByte() const { return parent_.mem_ + parent_.byteSize_; }                                 \
```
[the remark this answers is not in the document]
ericcano: I think we will now be safe. `data()` only exists in the context of the layout (formerly store) that is typically not exposed to the user anymore. The only use case I had for `data()` so far is to provide `memcpy` & friends with the buffer address. The views, which are exposed to the user, have no such notion of `data()` as the columns can be in various locations. I'll close this here, but this might need further discussion in the follow-up  PR.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+ * Example:
+ * generate_SoA_store(SoA,
+ *   // predefined static scalars
+ *   // size_t size;
+ *   // size_t alignment;
+ *
+ *   // columns: one value per element
+ *   SoA_FundamentalTypeColumn(double, x),
+ *   SoA_FundamentalTypeColumn(double, y),
+ *   SoA_FundamentalTypeColumn(double, z),
```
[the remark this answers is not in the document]
ericcano: No, this is a leftover, and was removed.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+ *   SoA_eigenColumn(Eigen::Vector3d, a),
+ *   SoA_eigenColumn(Eigen::Vector3d, b),
+ *   SoA_eigenColumn(Eigen::Vector3d, r),
+ *   SoA_column(uint16_t, colour),
+ *   SoA_column(int32_t, value),
+ *   SoA_column(double *, py),
```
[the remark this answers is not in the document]
ericcano: Yes. It works, but whether we have any use for this is a different question. This was just to push the limits and see if is worked.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+  ,                                                                                                                                 \
+    /* Column */                                                                                                                    \
+    size_t BOOST_PP_CAT(NAME, Pitch()) const {                                                                                      \
+      return (((parent_.nElements_ * sizeof(CPP_TYPE) - 1) / parent_.byteAlignment_) + 1) * parent_.byteAlignment_;                 \
+    }                                                                                                                               \
+    typedef CPP_TYPE BOOST_PP_CAT(TypeOf_, NAME);                                                                                   \
+    static const SoAColumnType BOOST_PP_CAT(ColumnTypeOf_, NAME) = SoAColumnType::column;                                           \
+    CPP_TYPE * BOOST_PP_CAT(addressOf_, NAME)() const { return parent_.BOOST_PP_CAT(NAME, _); }                                     \
```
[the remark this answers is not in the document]
ericcano: Compiler let this go though, but that's not coherent. I fixed the SoAMetadata const coherence in https://github.com/ericcano/pixeltrack-standalone/commit/988e8db615b61466740f19c436f6c56a298c4ad7

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+    /* Eigen column */                                                                                                              \
+    size_t BOOST_PP_CAT(NAME, Pitch()) const {                                                                                      \
+      return (((parent_.nElements_ * sizeof(CPP_TYPE::Scalar) - 1) / parent_.byteAlignment_) + 1) * parent_.byteAlignment_          \
+          * CPP_TYPE::RowsAtCompileTime * CPP_TYPE::ColsAtCompileTime;                                                              \
+    }                                                                                                                               \
+    typedef CPP_TYPE BOOST_PP_CAT(TypeOf_, NAME);                                                                                   \
+    static const SoAColumnType BOOST_PP_CAT(ColumnTypeOf_, NAME) = SoAColumnType::eigen;                                            \
+    CPP_TYPE::Scalar * BOOST_PP_CAT(addressOf_, NAME)() const { return parent_.BOOST_PP_CAT(NAME, _); }                             \
```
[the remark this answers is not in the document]
ericcano: Also covered in https://github.com/ericcano/pixeltrack-standalone/commit/988e8db615b61466740f19c436f6c56a298c4ad7.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+ *   SoA_eigenColumn(Eigen::Vector3d, r),
+ *   SoA_column(uint16_t, colour),
+ *   SoA_column(int32_t, value),
+ *   SoA_column(double *, py),
+ *   SoA_FundamentalTypeColumn(uint32_t, count),
+ *   SoA_FundamentalTypeColumn(uint32_t, anotherCount),
+ *
+ *   // scalars: one value for the whole structure
+ *   SoA_scalar(const char *, description),
+ *   SoA_scalar(uint32_t, someNumber)
```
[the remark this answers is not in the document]
ericcano: A general explanation of the SoA usage was added in https://github.com/ericcano/pixeltrack-standalone/commit/d86dd8a28336ec8c7ecd6a3288a34d71b4d22908.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+    CPP_TYPE const & NAME() { return soa_. NAME (); }                                                                               \
+  ,                                                                                                                                 \
+    /* Column */                                                                                                                    \
+    CPP_TYPE const & NAME() { return * (soa_. NAME () + index_); }                                                                  \
+  ,                                                                                                                                 \
+    /* Eigen column */                                                                                                              \
+    /* Ugly hack with a helper template to avoid having commas inside the macro parameter */                                        \
+    EigenConstMapMaker<CPP_TYPE>::Type const NAME() {                                                                               \
```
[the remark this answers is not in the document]
ericcano: This is a wrapper to `Eigen::Map<const CPP_TYPE, 0, Eigen::InnerStride<Eigen::Dynamic>>` , `CPP_TYPE` being for example `Eigen::Vector3d`.

This boils down to (in this case):
```C++
Eigen::Map<const Eigen::Vector3d, 0, Eigen::InnerStride<Eigen::Dynamic>> (
  soa_. NAME () + index_, 
  Eigen::Vector3d::RowsAtCompileTime, 
  Eigen::Vector3d::ColsAtCompileTime, 
  Eigen::InnerStride<Eigen::Dynamic>(soa_.NAMEStride())
);
```
The stride indicates our column length to Eigen, and we pick the right offset for the first address to access the desired vector in the SoA array of Eigen vectors. This is just a wrapping of code already developed by @VinInn.

The harcoded `0` is the alignement we can promise to Eigen (currently we don't, but we can probably go to `sizeof(Eigen::Vector3d::Scalar)`. I will replace it with `Eigen::AlignmentType::Unaligned` for the moment.

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+  _ACCUMULATE_SOA_ELEMENT_IMPL TYPE_NAME
+
+/**
+ * Value accessor of the const_element subclass.
+ */
+#define _DECLARE_SOA_CONST_ELEMENT_ACCESSOR_IMPL(VALUE_TYPE, CPP_TYPE, NAME)                                                        \
+  SOA_HOST_DEVICE_INLINE                                                                                                            \
+  _SWITCH_ON_TYPE(VALUE_TYPE,                                                                                                       \
+    /* Scalar */                                                                                                                    \
+    CPP_TYPE const & NAME() { return soa_. NAME (); }                                                                               \
```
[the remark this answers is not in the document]
ericcano: This is now changed following the creations of const views (and const elements).

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+  _SWITCH_ON_TYPE(VALUE_TYPE,                                                                                                       \
+    /* Scalar */                                                                                                                    \
+    BOOST_PP_EMPTY()                                                                                                                \
+  ,                                                                                                                                 \
+    /* Column */                                                              \
+    SOA_HOST_DEVICE_INLINE CPP_TYPE NAME() const { return BOOST_PP_CAT(NAME, _)(); }                                                \
+  ,                                                                                                                                 \
+    /* Eigen column */                                                                                                              \
+    SOA_HOST_DEVICE_INLINE const SoAEigenValue<CPP_TYPE> NAME() const { return BOOST_PP_CAT(NAME, _); }                             \
```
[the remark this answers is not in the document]
ericcano: I added this (creation of a const variant) to the planned modifications in `SoA.md`. (To be done in the new branch/followup PR).

src/cudadev/DataFormats/SoAStore.h
```diff
@@ -0,0 +1,569 @@
...
+   * up to compute capability 8.X.                                                                                                  \
+   */                                                                                                                               \
+  constexpr static size_t defaultAlignment = 128;                                                                                   \
+                                                                                                                                    \
+  /* dump the SoA internal structure */                                                                                             \
+  SOA_HOST_ONLY                                                                                                                     \
+  static void dump(size_t nElements, size_t byteAlignment = defaultAlignment) {                                                     \
+    std::cout << #CLASS "(" << nElements << ", " << byteAlignment << "): " << std::endl;                                            \
```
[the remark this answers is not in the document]
ericcano: Yes. I also plan this for the follow-up PR.

src/cudadev/DataFormats/SoAStore.h
ericcano: Fixed in b50d5a9.

src/cudadev/DataFormats/SoAStore.h
ericcano: `SoAMetadata` gives access to column addresses through the `addressOf_x()` member functions. Those are present in both views and layouts and are used internally to initialize view column addresses. The const version of a SoAView only delivers pointers to const data. This is still simple const correctness that can be bypassed by copy, but only applies to layouts and mutable views, for which const can already be bypassed by copy.

I completed the previous incomplete const correctness in the this area in the followup version.

The const views, on the opposite, are fully safe as they do not rely on const correctness.

src/cudadev/DataFormats/SoAStore.h
ericcano: ... and redundant. Fixed in a8bf1ff.

src/cudadev/DataFormats/SoAView.h
ericcano: This macro makes the const/non-const selectable via the `DATA` parameter. It is either called with `BOOST_PP_EMPTY()` [in the mutable variant of SoAMetadata](https://github.com/ericcano/pixeltrack-standalone/blob/macroBasedSoA-PR211-followup/src/cudadev/DataFormats/SoAView.h#L322) or `const` [for the const variant](https://github.com/ericcano/pixeltrack-standalone/blob/macroBasedSoA-PR211-followup/src/cudadev/DataFormats/SoAView.h#L432).

src/cudadev/DataFormats/SoAView.h
ericcano: Fixed in a8bf1ff.

src/cudadev/DataFormats/SoAView.h
ericcano: This is the same use case as https://github.com/cms-patatrack/pixeltrack-standalone/pull/211#discussion_r768066785.

src/cudadev/DataFormats/SoAView.h
ericcano: This will allow constructs in the form `soa[i] = soa[j]`, copying all members in one go. This expands to:
```C++
    __attribute__(( host )) __attribute__(( device )) __inline__ __attribute__((always_inline)) element& operator=(const element& other) {
      x ( ) = other . x ( ) ;
      y ( ) = other . y ( ) ;
      z ( ) = other . z ( ) ;
[...]
      anotherCount ( ) = other . anotherCount ( ) ;
      description ( ) = other . description ( ) ;
      someNumber ( ) = other . someNumber ( ) ;
      return *this;
    }
```
This is not used so far. And thinking of it, this is going to fail with scalars (now fixed in 7b24d99, adding a constexpr condition).

src/cudadev/DataFormats/SoAView.h
ericcano: This is the same use case as https://github.com/cms-patatrack/pixeltrack-standalone/pull/211#discussion_r768066785.

src/cudadev/DataFormats/SoAView.h
ericcano: This redudancy was removed for views (and gone for layouts, as they do not have accessors).

src/cudadev/DataFormats/SoAView.h
makortel: What would the compiler-generated assignment operator do then? (I'm assuming it would be different)

ericcano: In the initial scenario (shown above) we have hand made copy using `T& SoAValue<T>::operator&()` and `T SoAValue<T>::operator() const`. Letting the compiler call `SoAValue<T>::operator=<T2>(const T2&)` member by member would also have achieved the same result. 

This is not true anymore as we now that have the `if constexpr` clause.

src/cudadev/plugin-SiPixelRecHits/gpuPixelRecHits.h
ericcano: My understanding of line 33 is that we want to force a copy of the SoA structure from global memory to registers. Line 34 would require the same, were the `&` not be there. 

I tried without the copy/reference, original and with full copy (no `&` for the second SoA) and the kernel performances are indistinguishable (respectively):
```
                    4.28%  655.45ms     10000  65.545us  47.744us  446.01us  gpuPixelRecHits::getHits()
                    4.28%  652.34ms     10000  65.233us  46.912us  310.78us  gpuPixelRecHits::getHits()
                    4.30%  655.12ms     10000  65.512us  47.904us  374.08us  gpuPixelRecHits::getHits()
```
I removed the copies in 3986550.

ericcano:
With all the comments reviewed and the todo list recorded in the last section of [SoA.md](https://github.com/ericcano/pixeltrack-standalone/blob/macroBasedSoA-PR211-followup/src/cudadev/DataFormats/SoA.md) I will close this PR to preserve the branch and create a new one with the new https://github.com/ericcano/pixeltrack-standalone/tree/macroBasedSoA-PR211-followup branch.

ericcano:
This branch will now be followed up in #287 .

PR #216, opened 2021-08-31
[cudadev] Generalize caching allocator

makortel:
This PR abstracts the commonalities between `CachingDeviceAllocator` and `CachingHostAllocator` into a class template, with two traits classes providing the device/host-specific functionality. This abstraction should help with
- maintenance and evolution of the caching allocator (less copy-pasted code)
   * I have some ideas for reducing the number of CUDA events as part of https://github.com/cms-sw/cmssw/issues/30266
- porting the caching allocator to other APIs (device interaction is abstracted into the traits class)

I'd expect no performance difference, but that is still to be measured. After the review here I'm planning to propagate the changes to CMSSW.

### DIFF
--- a/src/cudadev/CUDACore/GenericCachingAllocator.h
+++ b/src/cudadev/CUDACore/GenericCachingAllocator.h
@@ -15,87 +15,6 @@
 
 // Inspired by cub::CachingDeviceAllocator
 
-/*
-struct CUDAHostTraits {
-  using DeviceType = int;
-  using QueueType = cudaStream_t;
-  using EventType = cudaEvent_t;
-  struct Dummy {};
-
-  constexpr DeviceType kInvalidDevice = -1;
-
-  static DeviceType currentDevice() {
-    return cms::cuda::currentDevice();
-  }
-
-  static Dummy setDevice(DeviceType device) {
-    return {};
-  }
-
-  static bool canReuseInDevice(DeviceType a, DeviceType b) {
-    // Pinned host memory can be reused in any device, but in case of
-    // changes the event must be re-created
-    return true;
-  }
-
-  static bool canReuseInQueue(QueueType a, QueueType b) {
-    // For pinned host memory a freed block without completed event
-    // can not be re-used even for operations in the same queue
-    return false;
-  }
-
-  static bool eventWorkHasCompleted(EventType e) {
-    return cms::cuda::eventWorkHasCompleted(e);
-  }
-
-  static EventType createEvent() {
-    EventType e;
-    cudaCheck(cudaEventCreateWithFlags(&e, cudaEventDisableTiming));
-    return e;
-  }
-
-  static void destroyEvent(EventType e) {
-    cudaCheck(cudaEventDestroy(e));
-  }
-
-  static EventType recreateEvent(EventType e, DeviceType prev, DeviceType next) {
-    cudaCheck(cudaSetDevice(prev));
-    destroyEvent(e);
-    cudaCheck(cudaSetDevice(next));
-    return createEvent();
-  }
-
-  static EventType recordEvent(EventType e, QueueType queue) {
-    cudaCheck(cudaEventRecord(e, queue));
-  }
-
-  static std::ostream& printDevice(std::ostream& os, DeviceType dev) {
-    os << "Host";
-    return os;
-  }
-
-  static void* allocate(size_t bytes) {
-    void* ptr;
-    cudaCheck(cudaHostAlloc(&ptr, bytes, cudaHostAllocDefault));
-    return ptr;
-  }
-
-  static void* tryAllocate(size_t bytes) {
-    void* ptr;
-    auto error = cudaHostAlloc(&ptr, bytes, cudaHostAllocDefault);
-    if (error == cudaErrorMemoryAllocation) {
-      return nullptr;
-    }
-    cudaCheck(error);
-    return ptr;
-  }
-
-  static void free(void* ptr) {
-    cudaCheck(cudaFreeHost(ptr));
-  }
-}
-*/
-
 namespace allocator {
   inline unsigned int intPow(unsigned int base, unsigned int exp) {
     unsigned int ret = 1;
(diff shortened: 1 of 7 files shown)


src/cudadev/CUDACore/GenericCachingAllocator.h
```diff
@@ -0,0 +1,345 @@
...
+        debug_(debug) {}
+  ~GenericCachingAllocator() { freeAllCached(); }
+
+  // Cache allocation status (for monitoring purposes)
+  DeviceCachedBytes cacheStatus() const {
+    std::scoped_lock lock(mutex_);
+    return cachedBytes_;
+  }
+
+  // Allocate given number of bytes on the current device associated to queue
```
fwyzard, line 79: if I read the code correctly,  the allocation is on the device identified by `device` , not on the current device associated to the queue ?

src/cudadev/CUDACore/GenericCachingAllocator.h
```diff
@@ -0,0 +1,345 @@
...
+    while (not cachedBlocks_.empty()) {
+      auto iBlock = cachedBlocks_.begin();
+      [[maybe_unused]] auto scopedSetDevice = Traits::setDevice(iBlock->device);
+      Traits::free(iBlock->ptr);
+      Traits::destroyEvent(iBlock->readyEvent);
+      cachedBytes_[iBlock->device].free -= iBlock->bytes;
+
+      if (debug_) {
+        std::cout << "\t";
+        Traits::printDevice(std::cout, iBlock->device)
```
fwyzard, line 314: This would feel more natural if it could be used like
```c++
std::cout << Traits::printDevice(iBlock->device) << ...
```

A simple implementation could simply return an `std::string`, while a more optimised implementation could return an object that can print its content.

src/cudadev/CUDACore/GenericCachingAllocator.h
```diff
@@ -0,0 +1,345 @@
...
+            << " live blocks (" << cachedBytes_[iBlock->device].live << " bytes) outstanding." << std::endl;
+      }
+
+      cachedBlocks_.erase(iBlock);
+    }
+  }
+
+  using Compare = typename std::add_pointer<bool(BlockDescriptor const&, BlockDescriptor const&)>::type;
+  using CachedBlocks = std::multiset<BlockDescriptor, Compare>;  // ordered by size
+  using BusyBlocks = std::multiset<BlockDescriptor, Compare>;    // ordered by ptr
```
fwyzard, line 326: Can this be a `set` instead of a `multiset` ?
If the key is the value of the pointer, we should never have two equivalent values.

src/cudadev/CUDACore/GenericCachingAllocator.h
```diff
@@ -0,0 +1,345 @@
...
+            << " available blocks cached (" << cachedBytes_[iBlock->device].free << " bytes), " << liveBlocks_.size()
+            << " live blocks (" << cachedBytes_[iBlock->device].live << " bytes) outstanding." << std::endl;
+      }
+
+      cachedBlocks_.erase(iBlock);
+    }
+  }
+
+  using Compare = typename std::add_pointer<bool(BlockDescriptor const&, BlockDescriptor const&)>::type;
+  using CachedBlocks = std::multiset<BlockDescriptor, Compare>;  // ordered by size
```
fwyzard, line 325: Thinking out loud - maybe it could be more efficient to use a `vector<vector<BlockDescriptor>>` ?
The outer `vector` would be indexed by the bin, so it could be constructed at the same time as the `GenericCachingAllocator`, and would never need to be resized.
Checking if a bin is empty would be fast (O(1)), while growing a bin would be O(log(N)) - but should quickly stop happening as we reuse the same buffers across the events.

src/cudadev/CUDACore/getCachingDeviceAllocator.h
```diff
@@ -50,15 +53,72 @@ namespace cms::cuda::allocator {
...
+    static EventType createEvent() {
+      EventType e;
+      cudaCheck(cudaEventCreateWithFlags(&e, cudaEventDisableTiming));
+      return e;
+    }
+
+    static void destroyEvent(EventType e) { cudaCheck(cudaEventDestroy(e)); }
+
+    static EventType recreateEvent(EventType e, DeviceType prev, DeviceType next) {
+      throw std::runtime_error("CUDADeviceTraits::recreateEvent() should never be called");
```
fwyzard, line 82: This should never happen because `canReuseInDevice()` returns false for different devices ?

src/cudadev/CUDACore/GenericCachingAllocator.h
makortel: Thanks, I'll fix the comment.

src/cudadev/CUDACore/GenericCachingAllocator.h
makortel: It certainly should be, but in a future PR. As we discussed privately, this PR will only reorganize the code (to make further changes easier) while keeping the old logic.

src/cudadev/CUDACore/getCachingDeviceAllocator.h
makortel: Correct. I think the logic is a bit convoluted even for the host allocator. I'm going to rewrite these parts ain order to reduce the number of CUDA events in flight in the allocator as part of https://github.com/cms-sw/cmssw/issues/30266 (essentially the allocator would be given only one `SharedEventPtr` for all allocations of an EDModule).

src/cudadev/CUDACore/GenericCachingAllocator.h
makortel: See https://github.com/cms-patatrack/pixeltrack-standalone/pull/216#discussion_r700359201.

src/cudadev/CUDACore/GenericCachingAllocator.h
makortel: Thanks, done.

src/cudadev/CUDACore/GenericCachingAllocator.h
makortel: I added this point to a new issue https://github.com/cms-patatrack/pixeltrack-standalone/issues/218

makortel:
No impact on performance, as expected.

PR #219, opened 2021-09-09
[alpaka] change assert with ALPAKA_ASSERT_OFFLOAD

antoniopetre:
Some Alpaka kernels have the occupancy lower than the Native CUDA implementation (for example ```countFromVector``` kernel from ```HistoContainer```). 

I found out that the problem was related to the ```assert``` use in Alpaka. In Native CUDA, the assert is disabled by default, becoming enabled only if the macro ```GPU_DEBUG``` is set (this is not the case in Alpaka). I changed the assert from the Alpaka implementation with the macro ```ALPAKA_ASSERT_OFFLOAD``` from the Alpaka library (compatible from Alpaka 0.6.1). This macro enables assert only if ```ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST``` is set. This change increased the occupancy for some kernels in Alpaka.


makortel:
Given that `cudacompat` has `assert()` enabled, should we include `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST` by default for serial and TBB backends?

@fwyzard

PR #221, opened 2021-09-13
Fix uploading the EventSetup conditions to multiple CUDA devices

makortel:
From https://github.com/cms-sw/cmssw/pull/34948 (and https://github.com/cms-sw/cmssw/pull/34968)

I did not try to compile the `hip` version.

FYI @fwyzard


makortel:
This should fix `cuda` and `cudauvm` to work on multiple GPUs. `cudadev` needs more work (in a subsequent PR) because I overlooked this aspect in #216 (for the host allocator).

PR #222, opened 2021-09-14
[cudadev] Fix generic caching allocator for multiple device

makortel:
This PR fixes an oversight of me in #216 towards multiple devices by making a distinction between "the device where the CUDA event is" and "the device where memory is allocated on". For device allocator those are the same, but for host allocator there is a difference.

This PR also removes the device number argument from the `allocate()` function in the hope of simplification as we always allocate on the current device, but leaves the argument for the `free()` because the device allocator uses that in the comparison function for the live memory blocks (but I'll continue on that in the PR comments). This change is not strictly necessary, it mainly avoids some redundant calls to `cudaSetDevice()`.


src/cudadev/CUDACore/getCachingDeviceAllocator.h
```diff
@@ -64,10 +64,24 @@ namespace cms::cuda::allocator {
...
     static bool canReuseInDevice(DeviceType a, DeviceType b) { return a == b; }
 
     static bool canReuseInQueue(QueueType a, QueueType b) { return a == b; }
 
+    template <typename C>
+    static bool deviceCompare(DeviceType a_dev, DeviceType b_dev, C&& compare) {
+      if (a_dev == b_dev) {
+        return compare();
+      }
+      return a_dev < b_dev;
```
makortel, line 82: If I understood the "Unified Virtual Addressing" correctly, using the device id explicitly to disambiguate device pointers from different devices should not be necessary on devices that support it (and also the device id can be obtained from the address). The use of `cudaMemcpyDefault` seems to be part of UVA as well, so in practice we're already requiring it. If we can rely on it here, the allocator API could be further simplified (assuming the overhead of `cudaPointerGetAttributes()` is negligible).

makortel:
> > This PR also removes the device number argument from the `allocate()` function in the hope of simplification as we always allocate on the current device
> 
> On a further thought I'm going to back this change out to keep the fix simpler (and merge it soon). Interface changes are better to be left with separate PRs.

On a further thought (after actually trying it out), I'm going to keep it :) The tension comes from the old `CachingHostAllocator` using the "current device" for the CUDA event to create on. Just using "current device" leads to simpler code (and we're anyway relying on the current device being set everywhere the `make_device_unique` or `make_host_unique` are called).

src/cudadev/CUDACore/getCachingDeviceAllocator.h
fwyzard: > The use of `cudaMemcpyDefault` seems to be part of UVA as well, so in practice we're already requiring it.

According to [3.2.8. Unified Virtual Address Space](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#unified-virtual-address-space) UVA requires 64 bit compilation and SM >= 2.0, so I'd say we run very little risks of ending up on any hardware that does not support it.

src/cudadev/CUDACore/GenericCachingAllocator.h
```diff
@@ -163,41 +172,36 @@ class GenericCachingAllocator {
...
         }
       }
-
-      if (recache) {
-        Traits::recordEvent(searchKey.readyEvent, searchKey.associatedQueue);
-      }
     }
 
     if (not recache) {
+      auto scopedSetDevice = Traits::setDevice(searchKey.device);
```
fwyzard, line 196: I think this is not needed, as `cudaFree`/`cudaFreeHost` and `cudaEventDestroy` can be called even if the current device does not match the one associated to the memory block and event.

I've tested by forcing the call to set the wrong device, and everything works fine.

src/cudadev/CUDACore/getCachingDeviceAllocator.h
fwyzard: In fact, I think unified addressing is built-in also in HIP and oneAPI.
Can we just consider it a requirement for any technology used with CMSSW ?

src/cudadev/CUDACore/GenericCachingAllocator.h
```diff
@@ -76,19 +76,24 @@ class GenericCachingAllocator {
...
+  void* allocate(size_t bytes, QueueType queue) {
     if (bytes > maxBinBytes_) {
       throw std::runtime_error("Requested allocation size " + std::to_string(bytes) +
                                " bytes is too large for the caching allocator with maximum bin " +
                                std::to_string(maxBinBytes_) +
                                " bytes. You might want to increase the maximum bin size");
     }
 
+    // device where to record the event on
+    const DeviceType deviceEvent = Traits::currentDevice();
```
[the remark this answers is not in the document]
makortel: Thanks, I agree.

PR #224, opened 2021-09-14
[cudadev][RFC] Next version of CUDA EDModule API

makortel:
This PR provides a prototype of https://github.com/cms-sw/cmssw/issues/30266#issuecomment-909526957.

In addition of interface changes, it fixes an issue with EDProducers that produce multiple CUDA products (I don't think we have any such EDProducers yet though). Currently the `cms::cuda::Product<T>` wrapper stores the same CUDA stream for each of such products, and since those `cms::cuda::SharedStreamPtr` are stored separately, tow (or more) of consumers of those products can submit their work into the same CUDA stream, whereas the intention was that only one consumer of any of the products of an EDProducer could use the same CUDA stream. This PR solves this by adding additional layer of indirection, i.e. the `cms::cuda::Product<T>` wrappers hold a `shared_ptr` into a helper object that contains the `SharedStreamPtr` and the logic of deciding which one of the customers can share the CUDA stream.


makortel:

PR #225, opened 2021-09-23
[kokkos] Use "device framework" in Kokkos version

makortel:
Introduces caching of the Execution Space objects in addition to what is in `kokkostest` (in principle should be "backported" there). The Execution Space objects are "heavy", and should be treated similarly to CUDA streams. With the caching the performance drop on a "one concurrent event" case is about 1-2 %.

With the Kokkos 3.4.1 only the CUDA backend works (as in runs and passes `--validation`) with multiple concurrent events and CPU threads. With Kokkos `develop` branch (i.e. their next version) also Serial backend works technically, but the performance is poor because of a "global lock". The performance of concurrent calls to Serial backend is expected to be improved in future versions of Kokkos. The performance of Kokkos-CUDA version is also bad, but because of continuous memory allocations and frees. The next step to improve performance is to use the caching allocator.

The Threads backend explicitly does not work from multiple calling CPU threads so I didn't even try, and the (Kokkos-)HIP version is still in progress in general so I didn't try that either.

### DIFF
--- a/src/kokkos/KokkosCore/ProductBase.h
+++ b/src/kokkos/KokkosCore/ProductBase.h
@@ -1,0 +1,128 @@
+#ifndef KokkosCore_ProductBase_h
+#define KokkosCore_ProductBase_h
+
+#include <exception>
+#include <memory>
+#include <string>
+
+#include <Kokkos_Core.hpp>
+
+#ifdef KOKKOS_ENABLE_CUDA
+#include "CUDACore/EventCache.h"
+#include "CUDACore/SharedEventPtr.h"
+#include "CUDACore/SharedStreamPtr.h"
+#include "CUDACore/StreamCache.h"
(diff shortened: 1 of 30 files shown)


src/kokkos/KokkosCore/ProductBase.h
```diff
@@ -0,0 +1,128 @@
...
+  namespace kokkos {
+    namespace impl {
+      class ExecSpaceSpecificBase {
+      public:
+        ExecSpaceSpecificBase() = default;
+        virtual ~ExecSpaceSpecificBase();
+      };
+
+      template <typename ExecSpace>
+      class ExecSpaceSpecific : public ExecSpaceSpecificBase {
```
makortel, line 29: These names are awful, but were good enough for prototyping.

PR #226, opened 2021-10-05
Fix namespace in hip port

markdewing:
Code introduced in 9f46b5695a7055577078f63702992a40b15056d8 added a call to ScopedSetDevice that should be in the cms::hip namespace in the hip port.


makortel:
Thanks!

PR #227, opened 2021-10-07
[cudacompat] Remove remaining CUDA dependencies

makortel:
I got annoyed-enough by `cudacompat` program still allocating pinned host memory and generally depending on CUDA, so this PR removes all that. A possible implication for the future is that we may need/want to create a separate program (or rename this one and dig the previous version of `cudacompat` as a base) for a full "portable via cudacompat" implementation.


fwyzard:
:+1: for dropping the dependency on CUDA

fwyzard:
I agree it might be better to change the name... just `compat` ? or `cpu-serial` ? or just `serial` ?

PR #229, opened 2021-10-07
[serial] Fork cudacompat and remove all CUDA dependencies.

makortel:
I got annoyed-enough by cudacompat program still allocating pinned host memory and generally depending on CUDA, so this PR forks `cudacompat` into `serial` program and removes all CUDA dependencies (replacing #227). This approach allows `cudacompat` to evolve independently if necessary.


makortel:
(Technical updates, fix the message of the first commit, and order two commits to get one of them compile as part of debugging. No functional difference)

makortel:
I'm anyway thinking to merge this while continuing to investigate the performance effect. The decision to keep `cudacompat` version paid off pretty quickly :)

PR #230, opened 2021-10-08
Fix AtomicPairCounter test for Kokkos/HIP

markdewing:
When running Kokkos with the HIP backend there is an assertion failure in `assert (c.m < n);` in the first loop ("update") in `src/kokos/test/kokkos/AtomicPairCounter_t.cc`.

The failure seems to be because the `counters` member is not initialized in the default constructor.   This was pretty clear from printing out values of `c.m` before the counter update.

The straight HIP port calls `hipMemset` to initialize the object to zero after allocation, so the problem does not occur there.

### DIFF
--- a/src/kokkos/KokkosCore/AtomicPairCounter.h
+++ b/src/kokkos/KokkosCore/AtomicPairCounter.h
@@ -11,7 +11,7 @@ namespace cms {
     public:
       using c_type = unsigned long long int;
 
-      KOKKOS_INLINE_FUNCTION AtomicPairCounter() {}
+      KOKKOS_INLINE_FUNCTION AtomicPairCounter() { counter.ac = 0; }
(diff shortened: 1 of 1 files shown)


makortel:
In the direct CUDA/HIP versions the constructors of data structures on device memory are trivial on purpose (so we can somewhat safely just omit calling them).

How about instead using `Kokkos::deep_copy()` to set the memory to zero, or if that doesn't work because `AtomicPairCounter` is a class, setting the `counter.ac` on 
https://github.com/cms-patatrack/pixeltrack-standalone/blob/0733b90b63675661375fc28bf16d02c7a746a41c/src/kokkos/test/kokkos/AtomicPairCounter_t.cc#L13
and then `deep_copy()`ing that to `dc_d` before the loop?

markdewing:
The deep copy does work.  If it's important to maintain the trivial constructor to keep it similar to CUDA, I can change the code to use that.

PR #234, opened 2021-10-13
Add --runForMinutes to all programs, and several updates to scanning script

### LINKED ISSUE #232: `run-scan.py` prints the wrong average
It looks like `run-scan.py` prints the wrong average when running with repeated measurements:
```
$ ./run-scan.py --minThreads 16 --maxThreads 24 --eventsPerStream 1000 --repeat 4 --warmup ./cudadev
21-10-12 09:31:17 Warming up
21-10-12 09:31:28 Processed 16000 events in 9.758109e+00 seconds, throughput 1639.66 events/s.

21-10-12 09:31:28 Number of streams 16 threads 16 events 16000
21-10-12 09:31:40 Processed 16000 events in 9.794204e+00 seconds, throughput 1633.62 events/s.
21-10-12 09:31:51 Processed 16000 events in 9.430950e+00 seconds, throughput 1696.54 events/s.
21-10-12 09:32:03 Processed 16000 events in 9.997050e+00 seconds, throughput 1600.47 events/s.
21-10-12 09:32:16 Processed 16000 events in 1.160675e+01 seconds, throughput 1378.51 events/s.
21-10-12 09:32:16 Number of streams 16 threads 16, average throughput 1378.510000

...
```

`1378.51` is the throughput of the last measurement, not the average.

makortel:
This PR includes several technical changes on running performance measurements

* Adds `--runForMinutes` argument for all programs. The argument is mutually exclusive with `--maxEvents`, and with it the program repeats the 1k events for at least the given integer number of minutes. The set of 1k events is always processed in full. In my tests so far the cost of the added synchronization is negligible.
* On `run-scan.py`
   * The **script no longer honors `CUDA_VISIBLE_DEVICES`**. Instead, it has a new argument `--cudaDevices` to specify the devices to use. The device indices correspond to those in `nvidia-smi` (whic does not, in general. hold for `CUDA_VISIBLE_DEVICES`)
       * This change was motivated by the monitoring of GPU status (see below), and the fact that the device indices in `CUDA_VISIBLE_DEVICES` are not necessarily consistent with those in `nvidia-smi`
       * The script prints all devices it sees (with the consistent indexing)
       * Information (name and driver version) are stored in the JSON file for those devices that were used in the measurement
   * Support for the --runForMinutes` option
   * Aggregate various arguments into groups to make the help message more clear
   * Add options to monitor the process host RSS and CUDA utilization, temperature, power, clock, and process memory usage
       * The host RSS is read from `/proc/`, and the GPU status with `nvidia-smi`
       * The `--monitorSeconds` argument specifies the sampling interval in seconds
       * In case the job uses multiple GPUs, the status (including the memory usage) is recorded for each GPU separately
   * Fix `--repeat` (resolves #232)
* A new script `run-many.py` to run a single meaurement that is a combination of the throughput of many processes. These processes can be different programs, use different numbers of threads/streams, be pinned to different NUMA nodes or CPU cores, and use different CUDA devices.
   * The script supports some of the arguments of `run-scan.py`, e.g. the monitoring.

### DIFF
--- a/run-many.py
+++ b/run-many.py
@@ -1,0 +1,308 @@
+#!/usr/bin/env python3
+
+import os
+import json
+import time
+import socket
+import argparse
+import importlib
+import subprocess
+import collections
+
+scan = importlib.import_module("run-scan")
+
+RunningProgram = collections.namedtuple("RunningProgram", ["program", "index", "handle"])
+
+class Program:
+    def __init__(self, description):
+        s = description.split(":")
+        self._program = s[0]
+        self._options = ["threads","streams","numa","cores","cudaDevices"]
+        valid = set(self._options)
+        for o in s[1:]:
+            (name, value) = o.split("=")
+            if name not in valid:
+                raise Exception("Unsupported option '{}'".format(name))
+            setattr(self, "_"+name, value)
+            valid.remove(name)
+        for o in valid:
+            setattr(self, "_"+o, None)
+        if self._threads is None:
+            self._threads = 1
+        if self._streams is None:
+            self._streams = self._threads
+        self._cudaDevices = self._cudaDevices.split(",") if self._cudaDevices is not None else []
+
+    def program(self):
+        return self._program
+
+    def programShort(self):
+        return os.path.basename(self._program)
+
+    def cudaDevices(self):
+        return self._cudaDevices
+
+    def makeCommandMessage(self, processUntil):
+        command = [self._program] + processUntil + ["--numberOfThreads", str(self._threads), "--numberOfStreams", str(self._streams)]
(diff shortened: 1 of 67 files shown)


fwyzard:
>     * The **script no longer honors `CUDA_VISIBLE_DEVICES`**. Instead, it has a new argument `--cudaDevices` to specify the devices to use. The device indices correspond to those in `nvidia-smi` (whic does not, in general. hold for `CUDA_VISIBLE_DEVICES`)

OK, looking at the changes I see this is done setting `CUDA_DEVICE_ORDER=PCI_BUS_ID`, and setting ``CUDA_VISIBLE_DEVICES` to the list of devices in PCI order -- which should be consistent with `nvidia-smi`. Right ?

PR #236, opened 2021-10-14
[alpaka] Add ScopedContext implementation

fwyzard:


src/alpaka/AlpakaCore/SharedEventPtr.h
```diff
@@ -0,0 +1,23 @@
...
+    // edm::ReusableObjectHolder the pointed-to type is more interesting
+    // to avoid extra layer of indirection
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    using Queue = alpaka::QueueCudaRtNonBlocking;
+#elif defined ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED
+    using Queue = alpaka::QueueCpuNonBlocking;
+#else
+    using Queue = alpaka::QueueCpuBlocking;
+#endif
+    using SharedEventPtr = std::shared_ptr<std::remove_pointer_t<alpaka::Event<Queue>>>;
```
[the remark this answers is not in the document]
fwyzard: True, thanks for catching this.

src/alpaka/AlpakaDataFormats/SiPixelDigiErrorsAlpaka.h
```diff
@@ -22,7 +22,6 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
       ALPAKA_ASSERT_OFFLOAD(perror_h->capacity() == static_cast<int>(maxFedWords));
 
       // TO DO: nothing really async in here for now... Pass the queue in constructor argument instead, and don't wait anymore!
```
fwyzard, line 24: remove the comment

src/alpaka/AlpakaDataFormats/SiPixelDigiErrorsAlpaka.h
```diff
@@ -22,7 +22,6 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
...
-      Queue queue(device);
       alpaka::memcpy(queue, error_d, error_h, 1u);
       alpaka::wait(queue);
```
fwyzard, line 26: remove the `wait` ?

src/alpaka/AlpakaDataFormats/TrackingRecHit2DAlpaka.h
```diff
@@ -72,7 +73,6 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
 #undef SET
 
       // SoA view on device:
-      Queue queue(device);
       auto view_h{cms::alpakatools::createHostView<TrackingRecHit2DSOAView>(&view, 1u)};
       alpaka::memcpy(queue, m_view, view_h, 1u);
       alpaka::wait(queue);
```
fwyzard, line 78: remove the `wait` ?

src/alpaka/plugin-SiPixelRecHits/alpaka/PixelRecHits.cc
```diff
@@ -41,9 +42,6 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
 #ifdef GPU_DEBUG
       std::cout << "launching getHits kernel for " << blocks << " blocks" << std::endl;
 #endif
-
-      Queue queue(device);
-
```
fwyzard, line 46: remove the `wait()` at the end of the function ?

src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelFedCablingMapESProducer.cc
```diff
@@ -33,15 +33,15 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
...
+    alpaka::prepareForAsyncCopy(cablingMap_d);
     alpaka::memcpy(queue, cablingMap_d, cablingMap_h, 1u);
     eventSetup.put(std::make_unique<SiPixelFedCablingMapGPUWrapper>(std::move(cablingMap_d), true));
 
     auto modToUnp_h{cms::alpakatools::createHostView<unsigned char>(modToUnpDefault.data(), modToUnpDefSize)};
     auto modToUnp_d{cms::alpakatools::allocDeviceBuf<unsigned char>(modToUnpDefSize)};
+    alpaka::prepareForAsyncCopy(modToUnp_d);
     alpaka::memcpy(queue, modToUnp_d, modToUnp_h, modToUnpDefSize);
 
-    alpaka::wait(queue);
```
fwyzard, line 43: this should not be removed !

src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelGainCalibrationForHLTESProducer.cc
```diff
@@ -48,19 +48,20 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
...
     auto rangeAndCols_d{cms::alpakatools::allocDeviceBuf<SiPixelGainForHLTonGPU::RangeAndCols>(2000u)};
+    alpaka::prepareForAsyncCopy(rangeAndCols_d);
     alpaka::memcpy(queue, rangeAndCols_d, rangeAndCols_h, 2000u);
 
     auto fields_h{cms::alpakatools::createHostView<SiPixelGainForHLTonGPU::Fields>(&gain.fields_, 1u)};
     auto fields_d{cms::alpakatools::allocDeviceBuf<SiPixelGainForHLTonGPU::Fields>(1u)};
+    alpaka::prepareForAsyncCopy(fields_d);
     alpaka::memcpy(queue, fields_d, fields_h, 1u);
 
-    alpaka::wait(queue);
```
fwyzard, line 62: this should not be removed !

src/alpaka/plugin-SiPixelRecHits/alpaka/PixelCPEFastESProducer.cc
```diff
@@ -60,10 +64,9 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
     params.m_averageGeometry = alpaka::getPtrNative(averageGeometry_d);
     auto params_h{cms::alpakatools::createHostView<pixelCPEforGPU::ParamsOnGPU>(&params, 1u)};
     auto params_d{cms::alpakatools::allocDeviceBuf<pixelCPEforGPU::ParamsOnGPU>(1u)};
+    alpaka::prepareForAsyncCopy(params_d);
     alpaka::memcpy(queue, params_d, params_h, 1u);
 
-    alpaka::wait(queue);
```
fwyzard, line 65: this should not be removed !

src/alpaka/AlpakaDataFormats/SiPixelDigiErrorsAlpaka.h
ghugo83: Once the `Scoped context` is implemented, the _local_ queue should be removed, and then it should be fine to remove the wait as well?

src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelFedCablingMapESProducer.cc
ghugo83: Ha yes true, it is not precisely that the `queue` is destructed, but just that the `queue` destructor _returns_.
https://github.com/alpaka-group/alpaka/blob/0.7.0/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp#L80
With no `wait`, the kernels start with no guarantee that the asynchronous mem copies are completed properly.

>Maybe we can suggest to the Alpaka developers that a Queue should wait for the work queued onto it on destruction ?
Maybe, but maybe in a sense we do not want a _local_ queue to be destructed, and wait _here_ anyway? 
I was thinking all these local queues (streams) should be removed once the `ScopedContext` is implemented: here the only thing we want is simply to have the data copies and the kernels that need the data on the _same_ stream?

Like in the ScopedContext PR, also maybe adding the equivalent of `SiPixelFedCablingMapGPUWrapper::getGPUProductAsync` (https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/cuda/CondFormats/SiPixelFedCablingMapGPUWrapper.cc#L27 ), and using it here: https://github.com/cms-patatrack/pixeltrack-standalone/pull/236/files#diff-04a70eb4f8dece3a473ca1100b50cd03da79257a1d6434d83607de8c106b23d2R73 
?

src/alpaka/AlpakaDataFormats/TrackingRecHit2DAlpaka.h
```diff
@@ -73,6 +73,11 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
       // SoA view on device:
       auto view_h{cms::alpakatools::createHostView<TrackingRecHit2DSOAView>(&view, 1u)};
       alpaka::memcpy(queue, m_view, view_h, 1u);
+#ifdef ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED
```
waredjeb, line 76: For some reason, if `alpaka::wait(queue)` is inside this `#ifdef`, it seems like it has no effect, and the problem remains

makortel:
On Cori I get a failure with `./alpaka --cuda --histogram`
```

----------
Caught std::runtime_error
SimpleAtomicHisto::fill(-nan): i -2147483648 min 0.000000 max 4000.000000 nbins 100
```

fwyzard:
I see the same problem locally, we just never tried `--histogram` so far.

fwyzard:
However, I get the same also with the current master .

src/alpaka/AlpakaCore/alpaka/ScopedContext.cc
```diff
@@ -0,0 +1,78 @@
...
+
+  ////////////////////
+
+  ScopedContextProduce::~ScopedContextProduce() {
+    // Intentionally not checking the return value to avoid throwing
+    // exceptions. If this call would fail, we should get failures
+    // elsewhere as well.
+    //cudaEventRecord(event_.get(), stream());
+    //alpaka::enqueue(stream(), getEvent(::ALPAKA_ACCELERATOR_NAMESPACE::Device).get());
+    //TODO
```
makortel, line 71: Was there a particular reason to leave this `enqueue(queue, event)` call out?

src/alpaka/AlpakaCore/alpaka/ScopedContext.cc
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    void ScopedContextHolderHelper::enqueueCallback(int device, ::ALPAKA_ACCELERATOR_NAMESPACE::Queue& stream) {
+      alpaka::enqueue(stream, [this, device]() {
+        auto data = new CallbackData{waitingTaskHolder_, device};
+        std::unique_ptr<CallbackData> guard{reinterpret_cast<CallbackData*>(data)};
+        edm::WaitingTaskWithArenaHolder& waitingTaskHolder = guard->holder;
+        int device2 = guard->device;
+        waitingTaskHolder.doneWaiting(nullptr);
+      });
```
makortel, line 44: How does the `alpaka::enqueue()` propagate errors from asynchronous processing? Is it like `cudaHostFunc()` that doesn't call the function in case of errors, an exception pops up elsewhere, or something else?

src/alpaka/AlpakaCore/alpaka/ScopedContext.cc
makortel: To answer myself, it is not needed yet because the `Product<T>` is not used.

Is there actually an `alpaka::wait()` missing from
https://github.com/cms-patatrack/pixeltrack-standalone/blob/30ce137658e71cede4c0fa952c7d1eca2b8aa3da/src/alpaka/plugin-SiPixelRecHits/alpaka/PixelRecHits.cc#L71-L81
?

fwyzard:
Ah, I see. I don't think we considered this (together with the fact that
the synchronisation on access to the objects is not implemented yet), which
actually may be why the TBB implementation fails.

PR #238, opened 2021-10-14
[alpaka] Increase vertex count tolerance to 2 to avoid spurious test failures

makortel:


fwyzard:
@makortel I've tried reverting this and I didn't run into any problems:
```bash
while true; do ./alpaka --tbb --numberOfThreads 16 --validation; done
```
did not complain after running over 20 times...

Is it possible that our recent work on the alpaka port (and on alpaka itself) has solved the reproducibility problems ?
In what environment did you encounter it ?

makortel:
> Is it possible that our recent work on the alpaka port (and on alpaka itself) has solved the reproducibility problems ?

I think this is certainly possible.

> In what environment did you encounter it ?

On Cori GPU nodes. Let me run a stress test there too with this PR reverted, and if that works, we'll actually revert the PR. How would that sound?

fwyzard:
> Let me run a stress test there too with this PR reverted, and if that works, we'll actually revert the PR. How would that sound?

Sounds good !

makortel:
Seems indeed to work now.

PR #239, opened 2021-10-15
Remove runtime recursion from find_ntuplets()

makortel:
From cms-sw/cmssw#35473 and cms-sw/cmssw#35542 for all other versions than `hip` (that was done in https://github.com/cms-patatrack/pixeltrack-standalone/pull/233).


VinInn:
I consider  "partial specializations of functions are not allowed" a defect in Alpaka.
The use of "if constexpr" instead of template specialization to terminate recursion is a more general coding pattern far more reaching that the case in hand

PR #220, opened 2021-09-09
[alpaka] Refactor prefixScan implementation

antoniopetre:
The ```prefixScan``` algorithm is implemented in Alpaka using two kernels, while a single kernel is used for Native CUDA.

I refactored the ```prefixScan``` implementation in order to use a single kernel (similar with the Native CUDA implementation).

PR #241, opened 2021-10-15
[alpaka] Make use of cms::alpakatools::ALPAKA_ACCELERATOR_NAMESPACE

fwyzard:
Move the accelerator-specific symbols from `cms::alpakatools` to `cms::alpakatools::ALPAKA_ACCELERATOR_NAMESPACE`.


fwyzard:
This PR includes #220 and #236, and is mostly to evaluate the feasibility of this approach.

fwyzard:
This PR **does not** fix the crash in `sortByPt2(...)` nor the crash in the TBB backend.

PR #248, opened 2021-10-19
[alpaka] Caching allocators for host and device

abhinavramesh8:
A caching memory allocator each for host and device have been implemented for Alpaka, similar to the CUDA version. Host and device unique pointers have been provided for managing memory allocations. These pointers use the caching allocators by default, but this behavior can be disabled at compile time. The appropriate changes have been made to the existing codebase in order to make use of these unique pointers.

### DIFF
--- a/src/alpaka/AlpakaDataFormats/SiPixelDigiErrorsAlpaka.h
+++ b/src/alpaka/AlpakaDataFormats/SiPixelDigiErrorsAlpaka.h
@@ -1,10 +1,11 @@
-#ifndef CUDADataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
-#define CUDADataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
+#ifndef AlpakaDataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
+#define AlpakaDataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
 
 #include "DataFormats/PixelErrors.h"
 #include "AlpakaCore/SimpleVector.h"
 
-#include "AlpakaCore/alpakaCommon.h"
+#include "AlpakaCore/device_unique_ptr.h"
+#include "AlpakaCore/host_unique_ptr.h"
 
 namespace ALPAKA_ACCELERATOR_NAMESPACE {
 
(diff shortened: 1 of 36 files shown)


src/alpaka/AlpakaDataFormats/SiPixelDigiErrorsAlpaka.h
```diff
@@ -1,29 +1,34 @@
-#ifndef CUDADataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
-#define CUDADataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
+#ifndef AlpakaDataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
+#define AlpakaDataFormats_SiPixelDigi_interface_SiPixelDigiErrorsCUDA_h
```
waredjeb, line 2:
```suggestion
#ifndef AlpakaDataFormats_SiPixelDigi_interface_SiPixelDigiErrorsAlpaka_h
#define AlpakaDataFormats_SiPixelDigi_interface_SiPixelDigiErrorsAlpaka_h
```

makortel:
In light of #260 can this PR be closed then?

fwyzard:
> I would imagine caching allocations for non-pinned host memory to only add overhead

While we likely won't gain in performance from the caching allocator for CPU memory, we do need the stream-ordered behaviour for the "device" memory operations when using a non-blocking queue for a CPU backend (_e.g._ TBB).

makortel:
> > I would imagine caching allocations for non-pinned host memory to only add overhead
> 
> While we likely won't gain in performance from the caching allocator for CPU memory, we do need the stream-ordered behaviour for the "device" memory operations when using a non-blocking queue for a CPU backend (_e.g._ TBB).

Right, but would a non-blocking queue for a CPU backend be useful for production use case? I'd naively expect non-blocking queue to mostly add overhead, even in the cases where intra-algorithm parallelization would otherwise be useful.

Written that, for testing purposes I agree it can be useful, and then a caching allocator (or other mechanism to keep the temporary memory alive) would be needed.

fwyzard:
> Right, but would a non-blocking queue for a CPU backend be useful for production use case? I'd naively expect non-blocking queue to mostly add overhead, even in the cases where intra-algorithm parallelization would otherwise be useful.

No, indeed - from a performance point of view, we do not want to use a non-blocking queue in production... especially with out `acquire()`/`produce()` mechanism.

But so far the non-blocking TBB backend has been very useful for finding synchronisation bugs :-)

> Written that, for testing purposes I agree it can be useful, and then a caching allocator (or other mechanism to keep the temporary memory alive) would be needed.

fwyzard:
Moved to #260 .

PR #250, opened 2021-10-20
[alpaka] Template StreamCache and EventCache

makortel:
First step in replacing the use of `ALPAKA_ACCELERATOR_NAMESPACE` of #236 with templates in places where templates are more natural fit.


fwyzard:
The changes in general look fine, but I'm worried about one point and we could improve things on another.

The worry is that `ALPAKA_ACCELERATOR_NAMESPACE::Platform` and `ALPAKA_ACCELERATOR_NAMESPACE::Device` point to the same types for the "CPU serial" and "TBB" backends: `alpaka::DevCpu` and `alpaka::PltfCpu`. So they 
While `ALPAKA_ACCELERATOR_NAMESPACE::Queue` is different, but only because one is synchronous (or blocking, i.e. `alpaka::QueueCpuBlocking = alpaka::QueueGenericThreadsBlocking<DevCpu>`) and the other is asynchronous (or non-blocking, i.e. `alpaka::QueueCpuNonBlocking = alpaka::QueueGenericThreadsNonBlocking<DevCpu>`).
Can this cause any problems ?  For example if we try a blocking TBB queue or a non-blocking CPU serial queue ?

The possible improvement is that the `Queue` and `Event` classes are already `shared_ptr`-like objects (they have a data member which is a `shared_ptr` to the underlying implementation).
Can we find a way to use them "by value" in the caches, avoid the extra `shared_ptr` wrapper ?

makortel:
> Can this cause any problems ? For example if we try a blocking TBB queue or a non-blocking CPU serial queue ?

The approach in this PR should work as long as the same `Queue` (or `Event`) type is not used for two `Device` types or two `Platform` types. I can't think of how two `Accelerator`s that use the same `Device` and `Platform` types sharing the same `StreamCache` (or `EventCache`) could be a problem.

> Can we find a way to use them "by value" in the caches, avoid the extra `shared_ptr` wrapper ?

That would be difficult with `ReusableObjectHolder` because the returned `shared_ptr` uses custom deleter that returns the pointed-to object to the cache. I can't quickly think of how to use value types that internally use `shared_ptr` with caches that those types are not aware of.

makortel:
> > > Can we find a way to use them "by value" in the caches, avoid the extra `shared_ptr` wrapper ?
> > 
> > 
> > That would be difficult with `ReusableObjectHolder` because the returned `shared_ptr` uses custom deleter that returns the pointed-to object to the cache. I can't quickly think of how to use value types that internally use `shared_ptr` with caches that those types are not aware of.
> 
> Can we make a `ReusableObjectHolderByValue` that uses a `tbb::concurrent_queue<T> m_availableQueue` and use these objects by value instead of by `unique_ptr`/`shared_ptr` ?

If the cache returns `shared_ptr<T>` (in order to use the custom deleter to return it to the cache), holding the `T` by value internally would lead to heap allocation on each `Cache::get()` call and moving `T` objects around.

But a custom return type that holds `T` by value and upon destruction moves the `T` back to the cache could work. I find it difficult to guess whether that would be more efficient than the additional pointer indirection with `ReusableObjectHolder`.

src/alpaka/AlpakaCore/alpakaConfigAcc.h
```diff
@@ -116,4 +116,11 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
   using AlpakaDeviceSubView = alpaka::ViewSubView<Device, TData, Dim1D, Idx>;
 }  // namespace ALPAKA_ACCELERATOR_NAMESPACE
 
+namespace cms::alpakatools {
+  template <>
+  struct DevType<ALPAKA_ACCELERATOR_NAMESPACE::Event> {
+    using type = ALPAKA_ACCELERATOR_NAMESPACE::Device;
+  };
```
fwyzard, line 123: I guess if we end up using this, we should ask Alpaka developers to add it centrally ?

fwyzard: Done in https://github.com/alpaka-group/alpaka/issues/1428

src/alpaka/AlpakaCore/getDevIndex.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include "AlpakaCore/alpakaConfigAcc.h"
+
+namespace cms::alpakatools {
+  template <typename Device>
+  inline int getDevIndex(Device const& dev) {
+    return 0;
+  }
+
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
```
fwyzard, line 12: Sorry, I messed up with the comment, here is what I wanted to suggest:

```c++
#include <alpaka/alpaka.hpp>

namespace cms::alpakatools {

  // generic interface, for DevOacc and DevOmp5
  template <typename TDev>
  int getDevIndex(TDev const& device) {
    return device.iDevice();
  }

  // specialisation for DevCpu
  template <>
  int getDevIndex<alpaka::DevCpu>(alpaka::DevCpu const& device) {
    return 0;
  }

#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
  // specialisation for DevCudaRt
  template <>
  int getDevIndex<alpaka::DevCudaRt>(alpaka::DevCudaRt const& device) {
    return device.m_iDevice;
  }
#endif  // ALPAKA_ACC_GPU_CUDA_ENABLED

}  // namespace cms::alpakatools
```

That is, use `DevCudaRt` instead of `DevUniformCudaHipRt`, since we use `AccGpuCudaRt` etc.

Good point about `DevCpu` being always defined.

fwyzard:
Sorry, I thought it a templated on TAcc as well ?

Better I stop, I'm too tired...

Issue #218, opened 2021-09-03
[cudadev] Improve caching allocator performance


makortel:
The generalization of the caching allocator in #216 makes it easier to make various improvements to the caching allocator. https://github.com/cms-patatrack/pixeltrack-standalone/pull/211#discussion_r701705367 shows a measurement pointing that the mutex in the caching allocator would be the bottleneck (my studies ~2 years ago pointed more to the mutex in CUDA, but things seem to have evolved). This PR is to discuss improvement ideas, with a(n ordered) plan shown below

- [ ] Generalize the caching allocator, done in #216 
- [ ] Improve the interaction of `ScopedContext` and the caching allocator by having the `ScopedContext` to pass `SharedEventPtr` to the caching allocator (evolution of ideas in https://github.com/cms-patatrack/cmssw/pull/412 and https://github.com/cms-patatrack/cmssw/pull/487)
   * This will reduce the number of CUDA events in flight, remove the need to create and destroy them, and simplify the logic especially on the host allocator side
- [ ] Replace the `multiset` with nested vectors (device, bin) for (much) faster lookup (from https://github.com/cms-patatrack/pixeltrack-standalone/pull/216#discussion_r700131933)
- [ ] Make locking finer grained (e.g. per bin)

fwyzard:
>     * Improve the interaction of `ScopedContext` and the caching allocator by having the `ScopedContext` to pass `SharedEventPtr` to the caching allocator (evolution of ideas in

IIUC this will make the caching allocator very tightly bound with the CMS-specific "CUDA framework", right ?

makortel:
> > ```
> > * Improve the interaction of `ScopedContext` and the caching allocator by having the `ScopedContext` to pass `SharedEventPtr` to the caching allocator (evolution of ideas in
> > ```
> 
> IIUC this will make the caching allocator very tightly bound with the CMS-specific "CUDA framework", right ?

The caching allocator would get a technical dependence on `cms::cuda::SharedEventPtr`, and a "behavioral dependence" that it is up to the caller of the allocator to give the `SharedEventPtr`. (in principle it could be possible to have also API without the `SharedEventPtr`, but that would add a dependence on `cms::cuda::EventCache`).

The `make_device_unique` and `make_host_unique` would get a dependence on some more abstract version of the `Context` classes (in CMSSW it would still reside in `HetegeneousCore/CUDAUtilities`).

Together these would allow
[...]

PR #256, opened 2021-10-26
[cudadev][RFC] Prototype (host|device)_unique_ptr API to use lightweight "Context" object instead of CUDA stream

makortel:
This PR builds on top of #224, but because of the actual developments conflict between the base commit of #224 and `master`, also the #224 part is rebased. The actual developments of this PR are in the last three commits.

The change can be summarized in `make_device_unique<T>(stream)` changing to `make_device_unique(ctx)` where `ctx` can be e.g. the `AcquireContext`/`ProduceContext`, or a "lightweight" `HostAllocatorContext`/`DeviceAllocatorContext`/`Context` (the `AcquireContext`/`ProduceContext` are convertible to the latter `Context` objects). (I'm really overusing the "Context" term here, but haven't figured out better wording yet).

The idea is that
* `HostAllocatorContext` provides the access to pinned host memory allocator (and only that)
* `DeviceAllocatorContext` provides access to device memory allocator (and only that)
* `Context` provides access to both pinned host and device memory allocators (via conversions to the two former types), and also whatever is needed to launch asynchronous kernels or memory transfers (in practice the CUDA stream)

This change would allow e.g.
* Moving the `CachingDeviceAllocator` and `CachingHostAllocator` objects from global variables to be owned (again) by `CUDAService` (in CMSSW only), that would further enable (again) the caching allocator parameters be configured at run time
* Possibly evolving the interplay between the caching allocators and the `AcquireContext`/`ProduceContext` for better performance (see discussion in #218) 
* Pinned host allocation API without explicit use of CUDA stream (evolving possibly later to an interface not requiring the stream, see point above). Such API would be more useful than the current CUDA stream -based API in ESProducers, that initiate transfers from one (pinned) host memory block to all available devices.


makortel:
Made effectively obsolete by https://github.com/cms-sw/cmssw/pull/39428 (although this particular development is not part of the CMSSW PR).

PR #257, opened 2021-10-27
[cudadev][RFC] Prototype an improved model for EventSetup use

makortel:
This PR builds on top of #256 (that builds on top of #224). The actual developments of this PR are in the last 4 commits.

This PR prototypes a major change for the EventSetup data model. Previously (currently) the ESProducer produces a wrapper product that holds the necessary (POD) data on the host memory, and the transfers to devices are done on EDModules when they ask for the actual product on the device memory, queued into the CUDA stream of the EDModule. This approach has (at least) the following downsides
* Requires CUDA runtime API calls in the EventSetup data formats, spreading the CUDA dependence wider in CMSSW
* Effectively prevents kernel calls to produce some of the data (because we haven't succeeded in having device code in library shared objects)
* Makes use of `CachingHostAllocator` complicated as the latter expects each allocation be tied to a CUDA stream, and there is no CUDA stream available at the `ESProducer::produce()` time. Even if there were (or the allocation API would be improved in some way), the necessary lifetime should be controlled by the produced data product.
* The transferred-from host memory is kept alive for the entire duration of the IOV, while it could be sufficient to release them after the data have been transferred to all devices
* `cms::cuda::ESProduct<T>` (that is used as a member data of the aforementioned "wrapper procuct") has complicated synchronization logic to handle concurrent requests on the actual device data

In the proposed model
* The transfers (and possible kernel calls) are launched from the ESProducer.
* The actual product is wrapped into `cms::cudaNew::ESProduct<T>` (to be renamed `cms::cuda::ESProduct<T>` in CMSSW after everything has been migrated to it) that acts similarly to `cms::cuda::Product<T>` for Event data
   * Holds a copy of `T` for each device, along with CUDA event telling if the asynchronous operations to produce `T` have completed
   * Holds a type-erased object (currently `std::any`, to be improved) to hold the temporary (pinned) host data used as a source of the transfers at least until the transfers to all devices have completed.
   * In case the transfer (and/or kernels) have not finished yet when an EDModule asks data, a `cudaStreamWaitEvent()` is called to insert a wait on the EDModule's CUDA stream on the ESProduct's CUDA event
* All CUDA-related operations need to be done within functors passed to `runForHost(<host-functor>).forEachDevice(<device-functor>)`
   * The `<host-functor>` is run once, and it must return an arbitrary object (e.g. `tuple` for multiple objects) holding all temporary host data. This object will be given to the `<device-functor>` as the first argument. The `HostAllocatorContext` gives access to cached pinned host memory allocations, but no other CUDA API calls are allowed (really anything that assumes a current device) 
   * The `<device-functor>` is called once for each device. The current device is set before calling the function. All CUDA operations are allowed.
   * In case no temporary host data are needed, a `runForEachDevice(<device-functor>)` can be used instead.
* Still requires CUDA dependence on the ES data formats because of the `unique_ptr`s. This dependence could be reduced (removed?) by using a smart pointer that erases the deleter type (similar to `std::shared_ptr`), which, I think, we should seriously consider (in light of Alpaka).


makortel:

PR #260, opened 2021-11-02
[alpaka] Caching allocators for host and device

waredjeb:
- Rebase of #248  on top of #255 

## Caching Allocator and Async Allocator Enabled
```
[patatrack02 pixeltrack-standalone]$ CUDA_VISIBLE_DEVICES=0 numactl -N 1 ./alpaka --cuda --numberOfThreads 8 --numberOfStreams 16 --validation --maxEvents 10000; echo; for N in 1 2 3 4; do CUDA_VISIBLE_DEVICES=0 numactl -N 1 ./alpaka --cuda --numberOfThreads 8 --numberOfStreams 16 --maxEvents 10000; done
Processing 10000 events, of which 16 concurrently, with 8 threads.
CountValidator: all 10000 events passed validation
 Average relative track difference 0.00088565 (all within tolerance)
 Average absolute vertex difference 0.0006 (all within tolerance)
Processed 10000 events in 1.472030e+01 seconds, throughput 679.334 events/s.

Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.990729e+00 seconds, throughput 2003.72 events/s.
Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.992631e+00 seconds, throughput 2002.95 events/s.
Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.986976e+00 seconds, throughput 2005.22 events/s.
Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.992776e+00 seconds, throughput 2002.89 events/s.
```
## Disabling the Caching Allocator and the Async Allocator
`make alpaka -j 10 CUDA_BASE=/usr/local/cuda-11.2 USER_CXXFLAGS="-DALPAKA_DISABLE_CACHING_ALLOCATOR -DALPAKA_DISABLE_ASYNC_ALLOCATOR"`

```
[patatrack02 pixeltrack-standalone]$ CUDA_VISIBLE_DEVICES=0 numactl -N 1 ./alpaka --cuda --numberOfThreads 8 --numberOfStreams 16 --validation --maxEvents 10000; echo; for N in 1 2 3 4; do CUDA_VISIBLE_DEVICES=0 numactl -N 1 ./alpaka --cuda --numberOfThreads 8 --numberOfStreams 16 --maxEvents 10000; done
Processing 10000 events, of which 16 concurrently, with 8 threads.
CountValidator: all 10000 events passed validation
 Average relative track difference 0.000888953 (all within tolerance)
 Average absolute vertex difference 0.0005 (all within tolerance)
Processed 10000 events in 4.884283e+01 seconds, throughput 204.738 events/s.

Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.017372e+01 seconds, throughput 248.919 events/s.
Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.029434e+01 seconds, throughput 248.174 events/s.
Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 4.101676e+01 seconds, throughput 243.803 events/s.
Processing 10000 events, of which 16 concurrently, with 8 threads.
Processed 10000 events in 3.935691e+01 seconds, throughput 254.085 events/s.
```

### DIFF
--- a/src/alpaka/AlpakaCore/CachingDeviceAllocator.h
+++ b/src/alpaka/AlpakaCore/CachingDeviceAllocator.h
@@ -1,0 +1,444 @@
+#ifndef HeterogenousCore_AlpakaUtilities_src_CachingDeviceAllocator_h
+#define HeterogenousCore_AlpakaUtilities_src_CachingDeviceAllocator_h
+
+/******************************************************************************
+ * Simple caching allocator for device memory allocations. The allocator is
+ * thread-safe and capable of managing device allocations on multiple devices.
+ ******************************************************************************/
+
+#include <cmath>
+#include <unordered_set>
+#include <memory>
+#include <mutex>
+
+#include "AlpakaCore/alpakaMemoryHelper.h"
+#include "AlpakaCore/deviceAllocatorStatus.h"
+
+/// cms::alpakatools::allocator namespace
+namespace cms::alpakatools::allocator {
+
+  /**
+ * \addtogroup UtilMgmt
+ * @{
+ */
+
+  /******************************************************************************
+ * CachingDeviceAllocator (host use)
+ ******************************************************************************/
+
+  /**
+ * \brief A simple caching allocator for device memory allocations.
+ *
+ * \par Overview
+ * The allocator is thread-safe and is capable of managing cached
+ * device allocations on multiple devices. It behaves as follows:
+ *
+ * \par
+ * - Allocations are categorized and cached by bin size.  A new allocation request of
+ *   a given size will only consider cached allocations within the corresponding bin.
+ * - Bin limits progress geometrically in accordance with the growth factor
+ *   \p bin_growth provided during construction.  Unused device allocations within
+ *   a larger bin cache are not reused for allocation requests that categorize to
+ *   smaller bin sizes.
+ * - Allocation requests below (\p bin_growth ^ \p min_bin) are rounded up to
+ *   (\p bin_growth ^ \p min_bin).
+ * - Allocations above (\p bin_growth ^ \p max_bin) are not rounded up to the nearest
+ *   bin and are simply freed when they are deallocated instead of being returned
+ *   to a bin-cache.
+ * - %If the total storage of cached allocations on a given device will exceed
+ *   \p max_cached_bytes, allocations for that device are simply freed when they are
+ *   deallocated instead of being returned to their bin-cache.
+ *
+ * \par
+ * For example, the default-constructed CachingDeviceAllocator is configured with:
+ * - \p bin_growth          = 8
+ * - \p min_bin             = 3
+ * - \p max_bin             = 7
+ * - \p max_cached_bytes    = 6MB - 1B
+ *
+ * \par
+ * which delineates five bin-sizes: 512B, 4KB, 32KB, 256KB, and 2MB
+ * and sets a maximum of 6,291,455 cached bytes per device
+ *
+ */
+  struct CachingDeviceAllocator {
+    //---------------------------------------------------------------------
+    // Constants
+    //---------------------------------------------------------------------
+
+    /// Out-of-bounds bin
+    static const unsigned int INVALID_BIN = (unsigned int)-1;
+
+    /// Invalid size
+    static const size_t INVALID_SIZE = (size_t)-1;
+
+#ifndef DOXYGEN_SHOULD_SKIP_THIS  // Do not document
+    //---------------------------------------------------------------------
+    // Type definitions and helper types
+    //---------------------------------------------------------------------
+
+    /**
+     * Descriptor for device memory allocations
+     */
+    struct BlockDescriptor {
+      ::ALPAKA_ACCELERATOR_NAMESPACE::AlpakaDeviceBuf<std::byte> buf;  // Device buffer
+      size_t bytes;                                                    // Size of allocation in bytes
+      size_t bytesRequested;  // CMS: requested allocation size (for monitoring only)
+      unsigned int bin;       // Bin enumeration
+
+      // Constructor (suitable for searching maps for a block, given a device and bytes)
+      BlockDescriptor(unsigned int block_bin,
+                      size_t block_bytes,
+                      size_t bytes_requested,
+                      const ::ALPAKA_ACCELERATOR_NAMESPACE::Device& device)
+          : buf{cms::alpakatools::allocDeviceBuf<std::byte>(device, 0u)},
+            bytes{block_bytes},
+            bytesRequested{bytes_requested},  // CMS
+            bin{block_bin} {}
+
+      // Constructor (suitable for searching maps for a specific block, given a device buffer)
+      BlockDescriptor(::ALPAKA_ACCELERATOR_NAMESPACE::AlpakaDeviceBuf<std::byte> buffer)
+          : buf{std::move(buffer)},
+            bytes{0},
+            bytesRequested{0},  // CMS
+            bin{INVALID_BIN} {}
+    };
+
+    struct BlockHashByBytes {
+      size_t operator()(const BlockDescriptor& descriptor) const {
+        size_t h1 = std::hash<int>{}(getIdxOfDev(alpaka::getDev(descriptor.buf)));
+        size_t h2 = std::hash<size_t>{}(descriptor.bytes);
+        return h1 ^ (h2 << 1);
+      }
+    };
+
+    struct BlockEqualByBytes {
+      bool operator()(const BlockDescriptor& a, const BlockDescriptor& b) const {
+        return (getIdxOfDev(alpaka::getDev(a.buf)) == getIdxOfDev(alpaka::getDev(b.buf)) && a.bytes == b.bytes);
+      }
+    };
+
+    struct BlockHashByPtr {
+      size_t operator()(const BlockDescriptor& descriptor) const {
+        size_t h1 = std::hash<int>{}(getIdxOfDev(alpaka::getDev(descriptor.buf)));
+        size_t h2 = std::hash<const std::byte*>{}(alpaka::getPtrNative(descriptor.buf));
+        return h1 ^ (h2 << 1);
+      }
+    };
+
+    struct BlockEqualByPtr {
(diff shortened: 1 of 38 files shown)


src/alpaka/AlpakaCore/CachingDeviceAllocator.h
```diff
@@ -0,0 +1,444 @@
...
+ * - \p min_bin             = 3
+ * - \p max_bin             = 7
+ * - \p max_cached_bytes    = 6MB - 1B
+ *
+ * \par
+ * which delineates five bin-sizes: 512B, 4KB, 32KB, 256KB, and 2MB
+ * and sets a maximum of 6,291,455 cached bytes per device
+ *
+ */
+  struct CachingDeviceAllocator {
```
fwyzard, line 64: This class should be templated on the `Device` type:
```suggestion
  template <typename TDevice>
  struct CachingDeviceAllocator {
  public:
    using Device = TDevice;
```

src/alpaka/AlpakaCore/CachingDeviceAllocator.h
```diff
@@ -0,0 +1,444 @@
...
+      ::ALPAKA_ACCELERATOR_NAMESPACE::AlpakaDeviceBuf<std::byte> buf;  // Device buffer
+      size_t bytes;                                                    // Size of allocation in bytes
+      size_t bytesRequested;  // CMS: requested allocation size (for monitoring only)
+      unsigned int bin;       // Bin enumeration
+
+      // Constructor (suitable for searching maps for a block, given a device and bytes)
+      BlockDescriptor(unsigned int block_bin,
+                      size_t block_bytes,
+                      size_t bytes_requested,
+                      const ::ALPAKA_ACCELERATOR_NAMESPACE::Device& device)
```
fwyzard, line 93: `::ALPAKA_ACCELERATOR_NAMESPACE::Device` should be replaced by the `Device` template type:
```suggestion
                      Device const& device)
```

src/alpaka/AlpakaCore/CachingDeviceAllocator.h
```diff
@@ -0,0 +1,444 @@
...
+          : buf{cms::alpakatools::allocDeviceBuf<std::byte>(device, 0u)},
+            bytes{block_bytes},
+            bytesRequested{bytes_requested},  // CMS
+            bin{block_bin} {}
+
+      // Constructor (suitable for searching maps for a specific block, given a device buffer)
+      BlockDescriptor(::ALPAKA_ACCELERATOR_NAMESPACE::AlpakaDeviceBuf<std::byte> buffer)
```
fwyzard, line 100: `::ALPAKA_ACCELERATOR_NAMESPACE::AlpakaDeviceBuf<std::byte>` should be replaced by a buffer type that depends on the `Device` template parameter:
```suggestion
      using DeviceBuffer = alpaka::Buf<Device, std::byte, Dim1D, Idx>;
      ...
      BlockDescriptor(DeviceBuffer buffer)
```

fwyzard (changes requested):
First round of comments...

src/alpaka/AlpakaCore/CachingDeviceAllocator.h
```diff
@@ -0,0 +1,444 @@
...
+        // Done searching: unlock
+        mutex_locker.unlock();
+      }
+
+      // Allocate the block if necessary
+      if (!found) {
+        search_key.buf = alpaka::allocBuf<std::byte, alpaka_common::Idx>(
+            device, static_cast<alpaka_common::Extent>(search_key.bytes));
+#if CUDA_VERSION >= 11020
+        alpaka::prepareForAsyncCopy(search_key.buf);
```
makortel, line 340: What does `prepareForAsyncCopy()` do? I'd guess it to call `cudaHostRegister()` or similar under the hood for host memory.

Looking at the code I see this is a no-op for CUDA/HIP buffers
https://github.com/alpaka-group/alpaka/blob/ee525dbb1c3e71490a17df80e7d13b3067619b95/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp#L408-L418
(so in principle this call would not be needed for the device allocator)

For host buffers it indeed ends up calling `cudaHostRegister()`
https://github.com/alpaka-group/alpaka/blob/ee525dbb1c3e71490a17df80e7d13b3067619b95/include/alpaka/mem/buf/BufCpu.hpp#L363-L378
https://github.com/alpaka-group/alpaka/blob/ee525dbb1c3e71490a17df80e7d13b3067619b95/include/alpaka/mem/buf/BufCpu.hpp#L269-L303

But given the
```cpp
#if(defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_LANG_CUDA) || (defined(ALPAKA_ACC_GPU_HIP_ENABLED) && BOOST_LANG_HIP)
```
the `cudaHostRegister()` is called only if the source file including this Alpaka header is compiled with `ALPAKA_ACC_GPU_CUDA_ENABLED` and compiling with `nvcc` (which we do when that macro is enabled). I'm afraid mixing that with code using `BufCpu.hpp` compiled without `ALPAKA_ACC_GPU_CUDA_ENABLED` would technically violate ODR.

I also see the definition of `alpaka::detail::BufCpuImpl` depends on these macros, in particular whether the `m_bPinned` member exists or not
https://github.com/alpaka-group/alpaka/blob/ee525dbb1c3e71490a17df80e7d13b3067619b95/include/alpaka/mem/buf/BufCpu.hpp#L103-L105

src/alpaka/AlpakaCore/CachingDeviceAllocator.h
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    /**
+     * \brief Provides a suitable allocation of device memory for the given size on the specified device.
+     *
+     * Once freed, the allocation becomes available immediately for reuse.
+     */
+    auto DeviceAllocate(size_t bytes,  ///< [in] Minimum no. of bytes for the allocation
+                        const ::ALPAKA_ACCELERATOR_NAMESPACE::Device&
+                            device)  ///< [in] The device to be associated with this allocation
```
makortel, line 281: What is the plan towards asynchronous (in a sense garbage-collecting) API?

src/alpaka/AlpakaCore/deviceAllocatorStatus.h
```diff
@@ -0,0 +1,29 @@
...
+        size_t free;
+        size_t live;
+        size_t liveRequested;  // CMS: monitor also requested amount
+        TotalBytes() { free = live = liveRequested = 0; }
+      };
+
+      inline int getIdxOfDev(const ::ALPAKA_ACCELERATOR_NAMESPACE::Device& device) {
+        static const auto devices{alpaka::getDevs<::ALPAKA_ACCELERATOR_NAMESPACE::Platform>()};
+        return (std::find(devices.begin(), devices.end(), device) - devices.begin());
+      }
```
makortel, line 21: (just repeating) Given https://github.com/cms-patatrack/pixeltrack-standalone/blob/master/src/alpaka/AlpakaCore/getDevIndex.h this is not needed.

src/alpaka/AlpakaCore/device_unique_ptr.h
```diff
@@ -0,0 +1,72 @@
...
+#include "AlpakaCore/host_unique_ptr.h"
+
+namespace cms {
+  namespace alpakatools {
+    namespace device {
+      namespace impl {
+        template <typename TData>
+        class DeviceDeleter {
+        public:
+          DeviceDeleter(::ALPAKA_ACCELERATOR_NAMESPACE::AlpakaDeviceBuf<TData> buffer) : buf{std::move(buffer)} {}
```
makortel, line 16: This class would need to be templated over Device.

src/alpaka/AlpakaCore/device_unique_ptr.h
```diff
@@ -0,0 +1,72 @@
...
+      template <typename TData>
+      using unique_ptr =
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+          std::unique_ptr<TData,
+                          impl::DeviceDeleter<
+                              std::conditional_t<allocator::policy == allocator::Policy::Caching, std::byte, TData>>>;
+#else
+          host::unique_ptr<TData>;
+#endif
+    }  // namespace device
```
makortel, line 39: Technically this violates ODR. Templating the `unique_ptr` also over Device and providing a specialization for `DevCudaRt` if `ALPAKA_ACC_GPU_CUDA_ENABLED` is defined would work. In longer term, I think, we should think about hiding the exact deleter type from the type of the `unique_ptr` (or just go directly to `shared_ptr` semantics).

src/alpaka/AlpakaCore/device_unique_ptr.h
```diff
@@ -0,0 +1,72 @@
...
+        auto buf = allocator::getCachingDeviceAllocator().DeviceAllocate(nbytes, device);
+        void* d_ptr = alpaka::getPtrNative(buf);
+        return typename device::unique_ptr<TData>{reinterpret_cast<TData*>(d_ptr),
+                                                  device::impl::DeviceDeleter<std::byte>{buf}};
+      } else {
+        auto buf = alpaka::allocBuf<TData, alpaka_common::Idx>(device, extent);
+#if CUDA_VERSION >= 11020
+        if constexpr (allocator::policy == allocator::Policy::Asynchronous) {
+          alpaka::prepareForAsyncCopy(buf);
+        }
```
makortel, line 60: The behavior of `Asynchronous` policy would be different from `cuda`, right?

src/alpaka/AlpakaCore/CachingDeviceAllocator.h
makortel: Do all scopes creating and destroying `unique_ptr`s have `alpaka::wait()` before the end of scope?

fwyzard:
reimplemented in #301

PR #261, opened 2021-11-08
[alpaka] modify hierarchy for atomics

felicepantaleo:
For atomics that operate on shared memory objects, operations don't need to be atomic among different blocks. 
The hierarchy for these atomics operations has been modified to avoid contention when running on CPU parallel backends.
Speedup of about 10% was observed with tbb backend.

### DIFF
--- a/src/alpaka/AlpakaCore/radixSort.h
+++ b/src/alpaka/AlpakaCore/radixSort.h
@@ -111,7 +111,7 @@ namespace cms::alpakatools::ALPAKA_ACCELERATOR_NAMESPACE {
       // fill bins
       ::cms::alpakatools::ALPAKA_ACCELERATOR_NAMESPACE::for_each_element_in_block_strided(acc, size, [&](uint32_t idx) {
         auto bin = (a[j[idx]] >> d * p) & (sb - 1);
-        alpaka::atomicAdd(acc, &c[bin], 1, alpaka::hierarchy::Blocks{});
+        alpaka::atomicAdd(acc, &c[bin], 1, alpaka::hierarchy::Threads{});
       });
       alpaka::syncBlockThreads(acc);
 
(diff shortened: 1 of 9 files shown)


src/alpaka/AlpakaCore/radixSort.h
```diff
@@ -111,7 +111,7 @@ namespace cms::alpakatools::ALPAKA_ACCELERATOR_NAMESPACE {
       // fill bins
       ::cms::alpakatools::ALPAKA_ACCELERATOR_NAMESPACE::for_each_element_in_block_strided(acc, size, [&](uint32_t idx) {
         auto bin = (a[j[idx]] >> d * p) & (sb - 1);
-        alpaka::atomicAdd(acc, &c[bin], 1, alpaka::hierarchy::Blocks{});
+        alpaka::atomicAdd(acc, &c[bin], 1, alpaka::hierarchy::Threads{});
```
makortel, line 114: Just to make sure I understood correctly, the last argument says that the operation should be atomic among Threads (of a Block) instead of among Blocks (of a Grid)?

fwyzard: Yes.

Maybe we could suggest adding aliases similar to those used in CUDA ?

Unfortunately there isn't a one-to-one correspondence:

CUDA | Alpaka | description
---|---|---
`atomicAdd_block` | `alpaka::atomicAdd(acc, ..., alpaka::hierarchy::Threads{})` | atomic for all threads in the current block
| | `alpaka::atomicAdd(acc, ..., alpaka::hierarchy::Blocks{})` |  atomic for all threads in the current grid
`atomicAdd` | `alpaka::atomicAdd(acc, ..., alpaka::hierarchy::Grids{})` | atomic for all threads in the current device
`atomicAdd_system` | | atomic for all threads in all GPUs and on the CPU

PR #277, opened 2021-12-05
[alpaka] Use stream-ordered device memory buffers

fwyzard:
Change the interface and behaviour of `allocDeviceBuf` to take as argument a `Queue&` instead of a `Device const&`, and allocate stream-ordered (or queue-ordered) memory.

The memory is scheduled for freeing when the buffer is destroyed, but is actually freed only when the queue reaches that point on the device side.

### DIFF
--- a/src/alpaka/AlpakaDataFormats/SiPixelDigisAlpaka.h
+++ b/src/alpaka/AlpakaDataFormats/SiPixelDigisAlpaka.h
@@ -10,14 +10,14 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
   class SiPixelDigisAlpaka {
   public:
     SiPixelDigisAlpaka() = default;
-    explicit SiPixelDigisAlpaka(Device const &device, size_t maxFedWords)
-        : xx_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          yy_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          adc_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          moduleInd_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          clus_d{::cms::alpakatools::allocDeviceBuf<int32_t>(device, maxFedWords)},
-          pdigi_d{::cms::alpakatools::allocDeviceBuf<uint32_t>(device, maxFedWords)},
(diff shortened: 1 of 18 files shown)


fwyzard:
Since we use only async buffers anyway, maybe it would be simpler to change the behaviour of `allocDeviceBuf` rather than introducing `allocDeviceAsyncBuf` ?

src/alpaka/AlpakaDataFormats/SiPixelDigisAlpaka.h
```diff
@@ -10,14 +10,14 @@ namespace ALPAKA_ACCELERATOR_NAMESPACE {
...
     SiPixelDigisAlpaka() = default;
-    explicit SiPixelDigisAlpaka(Device const &device, size_t maxFedWords)
-        : xx_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          yy_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          adc_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          moduleInd_d{::cms::alpakatools::allocDeviceBuf<uint16_t>(device, maxFedWords)},
-          clus_d{::cms::alpakatools::allocDeviceBuf<int32_t>(device, maxFedWords)},
-          pdigi_d{::cms::alpakatools::allocDeviceBuf<uint32_t>(device, maxFedWords)},
-          rawIdArr_d{::cms::alpakatools::allocDeviceBuf<uint32_t>(device, maxFedWords)} {}
+    explicit SiPixelDigisAlpaka(Queue queue, size_t maxFedWords)
```
makortel, line 13: I see `Queue` is taken by reference elsewhere, is copy here intentional?

makortel:
> Since we use only async buffers anyway, maybe it would be simpler to change the behaviour of `allocDeviceBuf` rather than introducing `allocDeviceAsyncBuf` ?

I could see that. The differentiation in naming could be useful (only?) if developers would be expected to deal with both kinds of buffers, but that is not the case here.

src/alpaka/AlpakaDataFormats/SiPixelDigisAlpaka.h
fwyzard: No... thanks for catching it. I've tested that passing by reference works fine (here and in two more places).

fwyzard:
> The differentiation in naming could be useful (only?) if developers would be expected to deal with both kinds of buffers, but that is not the case here.

OK, I'll revert the naming change. In the CUDA case we don't have the two options either.

PR #282, opened 2021-12-08
[alpaka] Add back a required synchronisation

fwyzard:
The `alpaka::wait(queue)` is required to ensure all async operations complete before the host buffer inside the `hits_d` temporary variable goes out of scope and destroyed. It should be possible to remove it once the host buffers are stream-ordered.


fwyzard:
With the TBB backend, the `alpaka::wait(queue)` prevents a segmentation fault or other memory corruption due to the use-after-free.

With the CUDA backend, adding the `alpaka::wait(queue)` increases the throughput by 12%, and I have no idea why.
This was measured with 16 threads/streams on a single T4.

makortel (approved):

fwyzard:
To verify the impact in the performance, here is a scan vs the number of threads:

image

Performed using a single Milan 7543 (32 physical cores) and a single T4 GPU.

PR #285, opened 2021-12-10
[alpaka] Make device initialisation non-static

fwyzard:
Template enumerate() on the device type, and move it out of the `ALPAKA_ACCELERATOR_NAMESPACE`.

The `ALPAKA_ACCELERATOR_NAMESPACE::devices` vectors are no longer initialised as static objects, but inside the `ALPAKA_ACCELERATOR_NAMESPACE::initialise()` functions.

The alpaka-specific Makefile is augmented with new compilation flags that describe all supported backends. These are used by `initialise()` to call `ALPAKA_ACCELERATOR_NAMESPACE::initialise()` for all backends.

`initialise()` is called explicitly by `main()`.


fwyzard:
Apart from generally having a non-deterministic order, the static initialisation breaks CUDA `compute-sanitize`, that complains about
```
COMPUTE-SANITIZER
Internal Sanitizer Error: CUDA initialized before the Sanitizer. The Sanitizer will be disabled.
```

This PR fixes this issue.

fwyzard:
Long term (at least for the migration in CMSSW), devices should be handled by a Service or other centralised, non-static approach.
An other improvement would be to remove the redundancy among different "accelerators" that use the same "device" (_e.g._ CPU serial and TBB both use the same `DevCpu` devices).

src/alpaka/bin/main.cc
```diff
@@ -45,6 +46,9 @@ namespace {
 }  // namespace
 
 int main(int argc, char** argv) {
+  // Initialiase all supported devices
+  initialise();
```
makortel, line 50: In principle this could initialize only the backends chosen by command line arguments, but I guess it makes little difference in practice.

makortel (approved):
Looks reasonable (given the scope).

src/alpaka/bin/main.cc
fwyzard: good point... I'll think if I can find a way to do that, while juggling the namespaces.

PR #286, opened 2021-12-11
[alpaka] Initialise only the selected backends

fwyzard:
Only the backends enabled at build time are supported in the command line options, and listed in the help message.
Only the backends selected on the command line are initialised.


fwyzard:
Building `alpaka` without CUDA support now results in
```
$ ./alpaka --cuda
Invalid parameter --cuda

./alpaka: [--serial] [--tbb] [--numberOfThreads NT] [--numberOfStreams NS] [--maxEvents ME] [--data PATH] [--transfer] [--validation]

Options
 --serial            Use CPU Serial backend
 --tbb               Use CPU TBB backend
 --numberOfThreads   Number of threads to use (default 1, use 0 to use all CPU cores)
 --numberOfStreams   Number of concurrent events (default 0 = numberOfThreads)
 --maxEvents         Number of events to process (default -1 for all events in the input file)
 --runForMinutes     Continue processing the set of 1000 events until this many minutes have passed (default -1 for disabled; conflicts with --maxEvents)
 --data              Path to the 'data' directory (default 'data' in the directory of the executable)
 --transfer          Transfer results from GPU to CPU (default is to leave them on GPU)
 --validation        Run (rudimentary) validation at the end (implies --transfer)
 --histogram         Produce histograms at the end (implies --transfer)
 --empty             Ignore all producers (for testing only)
```

fwyzard:
Building `alpaka` with CUDA enabled, but running with _e.g._ the serial backend, CUDA is never initialised:
```
$ compute-sanitizer --require-cuda-init yes --tool memcheck ./serial
========= COMPUTE-SANITIZER
========= Error: No attachable process found. compute-sanitizer timed-out.
========= Default timeout can be adjusted with --launch-timeout. Awaiting target completion.
Processed 1000 events in 3.383860e+01 seconds, throughput 29.552 events/s.
```

src/alpaka/bin/main.cc
```diff
@@ -42,13 +48,20 @@ namespace {
...
+  enum class Backend {
+#ifdef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_SUPPORTED
+    SERIAL,
+#endif
+#ifdef ALPAKA_ACC_CPU_B_TBB_T_SEQ_SUPPORTED
+    TBB,
+#endif
+#ifdef ALPAKA_ACC_GPU_CUDA_SUPPORTED
+    CUDA,
+#endif
```
makortel, line 60: I'd think the `enum` contents (and the code below adding the EDModules based on the selected backend(s)) would not necessarily need the `#ifdef`s, but I don't see a problem with them either.

I'm a bit surprised of C++ allowing a trailing comma in the definition :)

makortel (approved):
Visually looks ok.

src/alpaka/bin/main.cc
fwyzard: > I'm a bit surprised of C++ allowing a trailing comma in the definition :)

me too, but I tested it, and it works...

fwyzard: > I'd think the `enum` contents (and the code below adding the EDModules based on the selected backend(s)) would not necessarily need the `#ifdef`s, but I don't see a problem with them either.

Yes, maybe I should add back the definition of the `enum`.

I prefer to keep the EDModules part `#ifdef`ed out, since those symbols do not actually exist if the backend is not enabled.

makortel: > I prefer to keep the EDModules part `#ifdef`ed out, since those symbols do not actually exist if the backend is not enabled.

Even then having `#ifdef`s for the command line arguments should prevent non-existing EDModules getting loaded (and if not for some reason, the program would terminate with an obscure exception from `std::unordered_map`). Anyway, I'm fine with either way.

Issue #281, opened 2021-12-08
implement a SoA abstraction


fwyzard:
See https://github.com/cms-patatrack/cmssw/issues/272 for plenty of background information.

fwyzard:
See #211 for a CUDA implementation

fwyzard:
See #258 for the Alpaka implementation

ericcano:
> See #211 for a CUDA implementation

This has been moved to https://github.com/cms-patatrack/pixeltrack-standalone/pull/287 where we now have a feature complete implementation.

PR #287, opened 2021-12-13
Macro based SoA (followup of #PR211)

ericcano:
This PR follows up and replaces #211. Its goal is to implement #281.

The current status of SoA (with future plans) is described in [SoA.md](https://github.com/ericcano/pixeltrack-standalone/blob/macroBasedSoA-PR211-followup/src/cudadev/DataFormats/SoA.md)

### DIFF
--- a/src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
+++ b/src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
@@ -4,48 +4,48 @@
 #include "CUDACore/device_unique_ptr.h"
 #include "CUDACore/host_unique_ptr.h"
 #include "CUDACore/cudaCompat.h"
-#include "DataFormats/SoAStore.h"
+#include "DataFormats/SoALayout.h"
 #include "DataFormats/SoAView.h"
 
 #include <cuda_runtime.h>
 
 class SiPixelClustersCUDA {
 public:
-  generate_SoA_store(DeviceLayoutTemplate,
-    SoA_column(uint32_t, moduleStart),  // index of the first pixel of each module
-    SoA_column(uint32_t, clusInModule), // number of clusters found in each module
-    SoA_column(uint32_t, moduleId),     // module id of each module
+  GENERATE_SOA_LAYOUT(DeviceLayoutTemplate,
+    SOA_COLUMN(uint32_t, moduleStart),  // index of the first pixel of each module
+    SOA_COLUMN(uint32_t, clusInModule), // number of clusters found in each module
+    SOA_COLUMN(uint32_t, moduleId),     // module id of each module
 
     // originally from rechits
-    SoA_column(uint32_t, clusModuleStart) // index of the first cluster of each module
+    SOA_COLUMN(uint32_t, clusModuleStart) // index of the first cluster of each module
   );
   
   // We use all defaults for the template parameters.
   using DeviceLayout = DeviceLayoutTemplate<>;
 
     generate_SoA_view(DeviceViewTemplate,
-    SoA_view_store_list(SoA_view_store(DeviceLayout, deviceLayout)),
-    SoA_view_value_list(
-      SoA_view_value(deviceLayout, moduleStart),  // index of the first pixel of each module
-      SoA_view_value(deviceLayout, clusInModule), // number of clusters found in each module
-      SoA_view_value(deviceLayout, moduleId),     // module id of each module
+    SOA_VIEW_LAYOUT_LIST(SOA_VIEW_LAYOUT(DeviceLayout, deviceLayout)),
+    SOA_VIEW_VALUE_LIST(
+      SOA_VIEW_VALUE(deviceLayout, moduleStart),  // index of the first pixel of each module
+      SOA_VIEW_VALUE(deviceLayout, clusInModule), // number of clusters found in each module
+      SOA_VIEW_VALUE(deviceLayout, moduleId),     // module id of each module
   
       // originally from rechits
-      SoA_view_value(deviceLayout, clusModuleStart) // index of the first cluster of each module
+      SOA_VIEW_VALUE(deviceLayout, clusModuleStart) // index of the first cluster of each module
     )
   );
   
   using DeviceView = DeviceViewTemplate<>;
   
   generate_SoA_const_view(DeviceConstViewTemplate,
-    SoA_view_store_list(SoA_view_store(DeviceView, deviceView)),
-    SoA_view_value_list(
-      SoA_view_value(deviceView, moduleStart),  // index of the first pixel of each module
-      SoA_view_value(deviceView, clusInModule), // number of clusters found in each module
-      SoA_view_value(deviceView, moduleId),     // module id of each module
+    SOA_VIEW_LAYOUT_LIST(SOA_VIEW_LAYOUT(DeviceView, deviceView)),
+    SOA_VIEW_VALUE_LIST(
+      SOA_VIEW_VALUE(deviceView, moduleStart),  // index of the first pixel of each module
+      SOA_VIEW_VALUE(deviceView, clusInModule), // number of clusters found in each module
+      SOA_VIEW_VALUE(deviceView, moduleId),     // module id of each module
   
       // originally from rechits
-      SoA_view_value(deviceView, clusModuleStart) // index of the first cluster of each module
+      SOA_VIEW_VALUE(deviceView, clusModuleStart) // index of the first cluster of each module
     )
   );
   
--- a/src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
+++ b/src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
@@ -11,7 +11,7 @@
 
 class SiPixelClustersCUDA {
 public:
-  generate_SoA_store(DeviceStoreTemplate,
+  generate_SoA_store(DeviceLayoutTemplate,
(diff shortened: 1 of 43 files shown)


ericcano:
@fwyzard, @makortel, this version does not contain all the intended features, but could be considered for merging into master.

The missing features I intend to cover later are:
- Eigen support. This probably will require templating the way we describe columns in views, and I did not figure it out yet, and will probably require quite from testing, as this is macro-generated code. I also intend to make the inclusion of Eigen headers optional if Eigen columns are not used.
- Convert the dump member function into an operator<<.

I tried merging with the current master and this compiles and runs fine.

fwyzard:
A silly thing: could you run it all through `clang-format` ?

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
```diff
@@ -4,12 +4,54 @@
...
+    SOA_VIEW_LAYOUT_LIST(SOA_VIEW_LAYOUT(DeviceView, deviceView)),
+    SOA_VIEW_VALUE_LIST(
+      SOA_VIEW_VALUE(deviceView, moduleStart),  // index of the first pixel of each module
+      SOA_VIEW_VALUE(deviceView, clusInModule), // number of clusters found in each module
+      SOA_VIEW_VALUE(deviceView, moduleId),     // module id of each module
+  
+      // originally from rechits
+      SOA_VIEW_VALUE(deviceView, clusModuleStart) // index of the first cluster of each module
+    )
+  )
```
fwyzard, line 50: Why is this "derived" from `DeviceView` and not from `DeviceLayout` ?

ericcano: The "idea" was that the class has its mutable view internally for initialization, and then down grades it to a const view for usage by the user in member function `view()`. Both path are possible, but we need to use the type we will pass to the constructor. The "hard work" is in the layout constructor, and in both cases, we're copying pointers (and the compiler will optimize this copying things out).

That makes the name of the `SOA_VIEW_LAYOUT_LIST` macro too restrictive, maybe. What about `SOA_VIEW_DATA_SOURCE_LIST` (and accordingly for `SOA_VIEW_LAYOUT`)?

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
```diff
@@ -4,12 +4,49 @@
...
 #include "CUDACore/cudaCompat.h"
+#include "DataFormats/SoALayout.h"
+#include "DataFormats/SoAView.h"
 
 #include <cuda_runtime.h>
 
 class SiPixelClustersCUDA {
 public:
-  SiPixelClustersCUDA() = default;
+  GENERATE_SOA_LAYOUT(DeviceLayoutTemplate,
```
fwyzard, line 14: can this be simplified using `GENERATE_SOA_LAYOUT_VIEW_AND_CONST_VIEW` ?

src/cudadev/DataFormats/SoALayout.h
```diff
@@ -0,0 +1,373 @@
...
+                  , /* Column */                                                                                      \
+                  BOOST_PP_CAT(NAME, _) = reinterpret_cast<CPP_TYPE*>(curMem);                                        \
+                  curMem += (((nElements_ * sizeof(CPP_TYPE) - 1) / byteAlignment) + 1) * byteAlignment;              \
+                  , /* Eigen column */                                                                                \
+                  BOOST_PP_CAT(NAME, _) = reinterpret_cast<CPP_TYPE::Scalar*>(curMem);                                \
+                  curMem += (((nElements_ * sizeof(CPP_TYPE::Scalar) - 1) / byteAlignment) + 1) * byteAlignment *     \
+                            CPP_TYPE::RowsAtCompileTime * CPP_TYPE::ColsAtCompileTime;                                \
+                  BOOST_PP_CAT(NAME, Stride_) = (((nElements_ * sizeof(CPP_TYPE::Scalar) - 1) / byteAlignment) + 1) * \
+                                                byteAlignment / sizeof(CPP_TYPE::Scalar);)                            \
```
fwyzard, line 176: what happens if `byteAlignment` is not a multiple of `sizeof(CPP_TYPE::Scalar)` ?
what happens if `byteAlignment` is less than `sizeof(CPP_TYPE::Scalar)` ?

src/cudadev/CUDADataFormats/SiPixelClustersCUDA.h
ericcano: Indeed, this code predates the introduction of this utility macro.

src/cudadev/DataFormats/SoALayout.h
```diff
@@ -0,0 +1,373 @@
...
+  if constexpr (alignmentEnforcement == AlignmentEnforcement::Enforced)                                               \
+    if (reinterpret_cast<intptr_t>(BOOST_PP_CAT(NAME, _)) % byteAlignment)                                            \
+      throw std::out_of_range("In layout constructor: misaligned column: " #NAME);
```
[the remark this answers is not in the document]
ericcano: I did not find a perfect match in `std::exceptions` and felt this could be the closest (maybe `std::length_error` ? ). Otherwise, we could just create our own exceptions, but we mostly rely on `std::`  ones, don't we?

src/cudadev/DataFormats/SoALayout.h
ericcano: Indeed, bad things might happen. Usual values are powers of 2 and a typical byte alignment is bigger than a built-in scalar, but this is not guaranteed. I think asserting (compile-time) that both `sizeof(CPP_TYPE::Scalar)` and `byteAlignment` are powers of 2 and aligning the columns to the biggest of the two should forbid any unwanted scenario while remaining a reasonable constraint on the parameters.

src/cudadev/DataFormats/SoALayout.h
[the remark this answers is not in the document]
fwyzard: I'd go with a generic `std::runtime_error`.

src/cudadev/DataFormats/SoALayout.h
fwyzard: The alignment must always be a power of 2.
The size of a type can be anything, but c++ (almost) mandates that it is a multiple of its own alignment.

But "scalar" in this context does not necessarily mean a primitive type - it could also be an aggregate like class or an array (e.g. `char name[1024]`).
Then the stride becomes 0, which I don't think is the intended behaviour ?
If `sizeof(CPP_TYPE::Scalar)` is greater that the alignment, the stride should just be `nElements_ * sizeof(CPP_TYPE::Scalar)` .

fwyzard: Actually, shouldn't the stride simply be `(((nElements_ * sizeof(CPP_TYPE::Scalar) - 1) / byteAlignment) + 1) * byteAlignment`, without the `/ sizeof(CPP_TYPE::Scalar)` part ?

fwyzard:
Here are some ideas after reading most of the PR and the discussions of the past couple of days.

My understanding is that the basic SoA-based structure would look similar to
```c++
// define the SoALayout and SoAView types
GENERATE_SOA_LAYOUT_VIEW_AND_CONST_VIEW(
    SoALayoutTemplate,
    SoAViewTemplate,
    SoAConstViewTemplate,
    ...
);

using SoALayout = SoALayoutTemplate<>;
using SoAView = SoAViewTemplate<>;

class SoA {
public:
  SoA(size_t size) :
    storage_{std::make_unique_ptr<std::byte[]>(SoALayout::computeDataSize(size))},
    layout_{storage_.get(), size)},
    view_{layout_}
  {}

  SoAView view() { return view_; }  // return by value or by reference ?
  SoAConstView view() const { return const_view(); }
  SoAConstView const_view() const { return SoAConstView(layout_); }  // can we create the const view from the view instead ?

private:
  std::unique_ptr<std::byte[]> storage_;  // use a different storage type for GPUs, etc.
  SoALayout layout_;
  SoAView view_;
};
```

Then, I would have suggestions to simplify the usage.

#### Make the `View` and `ConstView` nested types of the `Layout`, and always declare them
```c++
```c++
// define the SoALayout and SoAView types
GENERATE_SOA_LAYOUT(
    SoALayoutTemplate,
    ...
);
```
would always generate
  - `SoALayoutTemplate<...>`
  - `SoALayoutTemplate<...>::View`
  - `SoALayoutTemplate<...>::ConstView`

and 
```c++
using SoALayout = SoALayoutTemplate<>;
```
would give
  - `SoALayout`
  - `SoALayout::View`
  - `SoALayout::ConstView`


#### Make the `view` a data member of the `layout`

If I understood correctly the content of the `Layout` and `View` types, the former is a superset of the latter.
Then we could simplify the `Layout` type by making it own a `View` data member, instead of duplicating those fields ?

So the `SoALayoutTemplate<...>` template would have a `SoALayoutTemplate<...>::View view_;` data member instead of the individual pointers.
It should also have the `view()`, `view() const` and `const_view() const` methods to return a `View` and `ConstView`

ericcano:
I attempted to achieve this, but failed in the current 2 classes model as the `View` is defined from the metadata of the `Layout` (column type are deduced from the metadata of the source layout/views). Instantiating the `View` early enough to be used inside the `Layout`

This schema can be achieved nevertheless if we create a third type of generated class (`Trivial_view`?) where the types of columns are directly generated, like for the `Layout`. It will be used like the `View` but generated in a different way.

PR #290, opened 2022-01-13
Update Kokkos to 3.5.00

makortel:


fwyzard:
No objections, of course, but out of curiosity what improvements does the new version bring ?

makortel:
Changelog can be found from https://github.com/kokkos/kokkos/blob/master/CHANGELOG.md#3500-2021-10-19

Of immediate interest are portable API for creating multiple execution space objects (~ CUDA streams), and Serial backend becoming thread safe (but simply protected with a mutex so not efficient); latter means that `kokkos` can be run with more than 1 threads, even if the throughput won't scale. They also added optional support for `cudaMallocAsync()`

PR #293, opened 2022-01-20
[alpaka] Update the alpaka external to the develop branch

fwyzard:


Makefile
```diff
@@ -550,7 +550,7 @@ $(HWLOC_BASE):
 external_alpaka: $(ALPAKA_BASE)
 
 $(ALPAKA_BASE):
-	git clone git@github.com:cms-patatrack/alpaka.git -b 0.8.0-patatrack $@
+	git clone git@github.com:alpaka-group/alpaka.git -b develop $@
```
makortel, line 553: Would picking a specific commit from `develop` be feasible? (now each `clone` could result in a different commit being used without any changes here)

fwyzard:
Good point... yes, I have a PR still open upstream, as soon it's merged
I'll put the commit hash in the Makefile.

fwyzard:
Done in https://github.com/cms-patatrack/pixeltrack-standalone/pull/299 .

PR #294, opened 2022-01-20
[alpaka] Simplify the interface for host and device buffers

fwyzard:
0-dimensional (scalar) buffers now implement the dereference `*` and member access `->` operators.
1-dimensional buffers now implement the array subscript `[]` operators.
all buffers implement the `.data()` method to access the underlying pointer.

### DIFF
--- a/src/alpaka/AlpakaCore/AtomicPairCounter.h
+++ b/src/alpaka/AlpakaCore/AtomicPairCounter.h
@@ -3,6 +3,8 @@
 
 #include <cstdint>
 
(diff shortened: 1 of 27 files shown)


makortel (approved):
Visually looks reasonable

fwyzard:
Thanks.
This (and the following changes in #296 and later) are not expected to jhave any impact on the performance: they only change the interfaces, but not what underlying code gets called.
The usual `--validation` checks pass, and I could not observe and change in performance - but currently the measurment are very noisy, so only major regressions could be visible,

PR #296, opened 2022-01-21
[alpaka] Rework the use of preprocessor symbols

fwyzard:
Rework the use of preprocessor symbols used to determine which backends are supported and active in a given compilation unit:
  - the `ALPAKA_ACC_..._SUPPORTED` symbols are not used any more;
  - the `ALPAKA_ACC_..._ENABLED` symbols (`ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED`, `ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED`, `ALPAKA_ACC_GPU_CUDA_ENABLED`, _etc._) indicate all the backends supported at compile time, and all are always defined;
  - the `ALPAKA_HOST_ONLY` symbol is defined by default, making it possible to compile the host-side API (device queries, memory handling, etc.) of all accelerators with the standard host compiler (e.g. `g++`);
  - the source code under the `src/alpaka/.../alpaka/` subdirectories is compiled multiple times, once per backend, and each time the corresponding `ALPAKA_ACC_..._[A]SYNC_BACKEND` symbol is defined;
  - only one  `ALPAKA_ACC_..._[A]SYNC_BACKEND` symbol should be defined at a given time;
  - when compiling for the CUDA backend, the `ALPAKA_HOST_ONLY` symbol is undefined, and `nvcc` is used as the compiler.

This allow moving more of the accelerator-independent code outside of the `src/alpaka/.../alpaka/` subdirectories.


fwyzard:
This PR is built on top of, and contains the changes from, #294.
It can be rebased once that is merged.

fwyzard:
Rebased.

makortel (approved):
Visually looks ok.

makortel:
Are the `ALPAKA_ACC_..._ENABLED` and `ALPAKA_HOST_ONLY` also recognized by Alpaka itself (in `develop`)?

fwyzard:
Thanks.
As the previous ones, it passes the usual variation, and does not show any impact on the performance (as expected).

fwyzard:
> Are the `ALPAKA_ACC_..._ENABLED` ...

Yes, this is the mechanism used by alpaka itself since ever.

> ... and `ALPAKA_HOST_ONLY` also recognized by Alpaka itself (in `develop`)?

Yes, this was added in `develop` via https://github.com/alpaka-group/alpaka/pull/1567 .

makortel:
> > Are the `ALPAKA_ACC_..._ENABLED` ...
> 
> Yes, this is the mechanism used by alpaka itself since ever.

Ah, right (what was I thinking). I should have written `ALPAKA_ACC_.._BACKEND`, and I see it continues to be our own construct. 

I guess I'm wondering how the macro setup would extend to a case where we'd want to use also other device backends than CUDA (even if only "one at a time"). But of course there are also other hurdles in the way.

fwyzard:
> Ah, right (what was I thinking). I should have written `ALPAKA_ACC_.._BACKEND`, and I see it continues to be our own construct.

Ah, yes: basically this PR gives back its original meaning to the `ENABLED` macros, and adds `BACKEND` to mean "the one to compile the device code and kernel calls for".


> I guess I'm wondering how the macro setup would extend to a case where we'd want to use also other device backends than CUDA (even if only "one at a time"). But of course there are also other hurdles in the way.

I've been thinking about it, and I think we could make it work with two set of changes (still to be discussed with the alpaka developers):
  - the CUDA- and HIP-related objects should be distinct types, instead of being just typedefs for the same type;
  - the logic of `ALPAKA_HOST_ONLY` should be reversed: host-only should become the default, and a separate macro for each backend should be used to enable compiling the corresponding device code and kernel launches.

I still don't have much experience with the HIP compiler, hopefully I can catch up next week.

fwyzard:
There is one item I don't think I know how to handle, though: pinned host memory.

If we support both NVIDIA/CUDA and AMD/HIP, how do we pin memory ?
Can we call _both_ `cudaHostRegister` and `hipHostRegister` on the same memory buffer ?
Or do we need to change the interface to "please pin _this_ memory area for _that_ device/backend" ?

fwyzard:
> Kokkos treats "CUDA pinned host memory" and "HIP pinned host memory" as separate things (but also doesn't even try to support both in one process).

The downside of this approach is that the hypothetical `pin()` method can no longer be a property of CPU memory buffer, but needs to be told the target memory space, and the buffer needs a way to record it or how to "unpin" the memory.

Which is doable, it's just not the current approach used in alpaka.

makortel:
> The downside of this approach is that the hypothetical `pin()` method can no longer be a property of CPU memory buffer, but needs to be told the target memory space, and the buffer needs a way to record it or how to "unpin" the memory.

Right. I'm actually a bit confused of the usefulness of the buffer's `pin()` method way for us. I would imagine it to be useful when only some of the host buffers are transferred to the device. But how would it work with caching of the pinned host memory? Wouldn't the point of "call `pin()` in `memcpy()`" be too late for obtaining an existing memory block from the cache? So far the approach of deciding at memory buffer allocation time seems to have worked pretty well.

Ok, I can think of one use case where pinning later than allocation could be useful: transferring an existing data product to device memory without copying from regular memory to pinned memory. But if the pinning/unpinning still synchronizes everything, I have doubts on the overall performance with multiple CPU threads.

fwyzard:
> Right. I'm actually a bit confused of the usefulness of the buffer's `pin()` method way for us. I would imagine it to be useful when only some of the host buffers are transferred to the device. But how would it work with caching of the pinned host memory? Wouldn't the point of "call `pin()` in `memcpy()`" be too late for obtaining an existing memory block from the cache? So far the approach of deciding at memory buffer allocation time seems to have worked pretty well.

I agree, and a simple way would be to just `pin()` every host buffer - whether they are used for transfer or not - immediately after allocating them, basically replacing `cudaMallocHost(&ptr)` with `ptr = malloc(); cudaHostRegister(ptr)`.
But I have not investigated whether that's good or bad for performance.

Or we can keep `make_host_buffer()` returning non-pinned memory, add `make_pinned_buffer()` that returns pinned buffer, and apply the caching only to the latter.

ericcano:
Looking around, it seems pinning memory and keeping track of it will be a driver duty, so I would suspect each driver has its own management of the pinning (making pinned memory a per-device thing like kokkos does). https://www.kernel.org/doc/html/latest/core-api/pin_user_pages.html#

For the cache-DMA issue, as far as I understand, this is handled by the CPU's cache pretty much like other memory accesses.

PR #301, opened 2022-02-01
[alpaka] Host and device memory caching allocator

fwyzard:
Implement a caching allocator based on the `cuda`/`cudadev` one for device memory and pinned host memory.

The performance profile of the `alpaka --cuda` program now follows the one of the `cuda` program, while still having a ~5% performance degradation (shown in the plot after merging #307 and #308):

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,25 @@ $ make kokkos ...
 | `-DKOKKOS_SERIALONLY_DISABLE_ATOMICS`  | Disable Kokkos (real) atomics, can be used with Serial-only build |
 
 
+#### `alpaka`
+
+The `alpaka` code base is loosely based on the `cuda` code base, with some minor changes introduced during the porting.
+
+The use of caching allocator can be disabled at compile time setting the
+`ALPAKA_DISABLE_CACHING_ALLOCATOR` preprocessor symbol:
+```
+make alpaka ... USER_CXXFLAGS="-DALPAKA_DISABLE_CACHING_ALLOCATOR"
+```
+
+If the caching allocator is disabled and CUDA version is 11.2 or greater is detected,
+device allocations and deallocations will use the stream-ordered CUDA functions
+`cudaMallocAsync` and `cudaFreeAsync`. Their use can be disabled explicitly at
+compile time setting also the `ALPAKA_DISABLE_ASYNC_ALLOCATOR` preprocessor symbol:
+
+```
+make alpaka ... USER_CXXFLAGS="-DALPAKA_DISABLE_CACHING_ALLOCATOR -DALPAKA_DISABLE_ASYNC_ALLOCATOR"
+```
+
 ## Code structure
 
 The project is split into several programs, one (or more) for each
--- a/src/alpaka/AlpakaCore/CachedBufAlloc.h
+++ b/src/alpaka/AlpakaCore/CachedBufAlloc.h
@@ -1,0 +1,95 @@
+#ifndef AlpakaCore_CachedBufAlloc_h
+#define AlpakaCore_CachedBufAlloc_h
+
+#include <alpaka/alpaka.hpp>
+
+#include "AlpakaCore/getDeviceCachingAllocator.h"
+#include "AlpakaCore/getHostCachingAllocator.h"
+#include "Framework/demangle.h"
+
+namespace cms::alpakatools {
+
+  namespace traits {
+
+    //! The caching memory allocator trait.
+    template <typename TElem, typename TDim, typename TIdx, typename TDev, typename TQueue, typename TSfinae = void>
+    struct CachedBufAlloc {
+      static_assert(alpaka::meta::DependentFalseType<TDev>::value, "This device does not support a caching allocator");
+    };
+
+    //! The caching memory allocator implementation for the CPU device
+    template <typename TElem, typename TDim, typename TIdx, typename TQueue>
+    struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, TQueue, void> {
+      template <typename TExtent>
+      ALPAKA_FN_HOST static auto allocCachedBuf(alpaka::DevCpu const& dev, TQueue queue, TExtent const& extent)
+          -> alpaka::BufCpu<TElem, TDim, TIdx> {
+        // non-cached host-only memory
+        return alpaka::allocAsyncBuf<TElem, TIdx>(queue, extent);
+      }
+    };
+
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+
+    //! The caching memory allocator implementation for the pinned host memory
+    template <typename TElem, typename TDim, typename TIdx>
+    struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, alpaka::QueueUniformCudaHipRtNonBlocking, void> {
+      template <typename TExtent>
+      ALPAKA_FN_HOST static auto allocCachedBuf(alpaka::DevCpu const& dev,
+                                                alpaka::QueueUniformCudaHipRtNonBlocking queue,
+                                                TExtent const& extent) -> alpaka::BufCpu<TElem, TDim, TIdx> {
+        ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
+
+        auto& allocator = getHostCachingAllocator<alpaka::QueueUniformCudaHipRtNonBlocking>();
+
+        // FIXME the BufCpu does not support a pitch ?
+        size_t size = alpaka::extent::getExtentProduct(extent);
+        size_t sizeBytes = size * sizeof(TElem);
+        void* memPtr = allocator.allocate(sizeBytes, queue);
+
+        // use a custom deleter to return the buffer to the CachingAllocator
+        auto deleter = [alloc = &allocator](TElem* ptr) { alloc->free(ptr); };
+
+        return alpaka::BufCpu<TElem, TDim, TIdx>(dev, reinterpret_cast<TElem*>(memPtr), std::move(deleter), extent);
+      }
+    };
+
+    //! The caching memory allocator implementation for the CUDA/HIP device
+    template <typename TElem, typename TDim, typename TIdx, typename TQueue>
+    struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevUniformCudaHipRt, TQueue, void> {
+      template <typename TExtent>
+      ALPAKA_FN_HOST static auto allocCachedBuf(alpaka::DevUniformCudaHipRt const& dev,
+                                                TQueue queue,
+                                                TExtent const& extent)
+          -> alpaka::BufUniformCudaHipRt<TElem, TDim, TIdx> {
+        ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
+
+        auto& allocator = getDeviceCachingAllocator<alpaka::DevUniformCudaHipRt, TQueue>(dev);
+
+        size_t width = alpaka::extent::getWidth(extent);
+        size_t widthBytes = width * static_cast<TIdx>(sizeof(TElem));
+        // TODO implement pitch for TDim > 1
+        size_t pitchBytes = widthBytes;
+        size_t size = alpaka::extent::getExtentProduct(extent);
+        size_t sizeBytes = size * sizeof(TElem);
+        void* memPtr = allocator.allocate(sizeBytes, queue);
+
+        // use a custom deleter to return the buffer to the CachingAllocator
+        auto deleter = [alloc = &allocator](TElem* ptr) { alloc->free(ptr); };
+
+        return alpaka::BufUniformCudaHipRt<TElem, TDim, TIdx>(
+            dev, reinterpret_cast<TElem*>(memPtr), std::move(deleter), pitchBytes, extent);
+      }
+    };
+
+#endif  // ALPAKA_ACC_GPU_CUDA_ENABLED
+
+  }  // namespace traits
+
+  template <typename TElem, typename TIdx, typename TExtent, typename TQueue, typename TDev>
+  ALPAKA_FN_HOST auto allocCachedBuf(TDev const& dev, TQueue queue, TExtent const& extent = TExtent()) {
+    return traits::CachedBufAlloc<TElem, alpaka::Dim<TExtent>, TIdx, TDev, TQueue>::allocCachedBuf(dev, queue, extent);
+  }
+
+}  // namespace cms::alpakatools
+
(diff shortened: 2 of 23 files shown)


fwyzard:
This is a first working implementation of a caching allocator written using alpaka buffers as its backend.

It does achieve better performance than using native CUDA memory allocations (in large part because of the caching of the host-side pinned buffers), even though the performance is not as good as that of the native CUDA version:

image

src/alpaka/AlpakaCore/CachingAllocator.h
```diff
@@ -0,0 +1,424 @@
...
+
+      // iterate through the range of cached blocks in the same bin
+      const auto [begin, end] = cachedBlocks_.equal_range(block.bin);
+      for (auto iBlock = begin; iBlock != end; ++iBlock) {
+        if (
+            // FIXME figure out which combination of devices allow reusing non-complete events in the same queue
+            // e.g.
+            //     for pinned host memory a freed block without completed event can not be re-used
+            //     even for operations in the same queue (?)
+            /* (*block.queue == *(iBlock->second.queue) or */
```
fwyzard, line 281: @makortel could you remind me why this is disabled for the host case ?

src/alpaka/AlpakaCore/alpakaMemory.h
```diff
@@ -73,21 +75,68 @@ namespace cms::alpakatools {
...
+  // the memory is pinned by the caching allocator, or explicitly if it is not used
+
+  template <typename T, typename TQueue>
+  std::enable_if_t<not std::is_array_v<T>, host_buffer<T>> make_host_buffer(TQueue const& queue [[maybe_unused]]) {
+    if constexpr (allocator_policy<alpaka::Dev<TQueue>> == AllocatorPolicy::Caching) {
+      return allocCachedBuf<T, Idx>(host, queue, Scalar{});
+    } else {
+      auto buffer = alpaka::allocBuf<T, Idx>(host, Scalar{});
+      alpaka::prepareForAsyncCopy(buffer);
+      return buffer;
```
makortel, line 114: How about
```suggestion
      return make_host_buffer<T>();
```
to make it explicit that the code is intended to be the same? (same for the other two cases below)

src/alpaka/AlpakaCore/getCachingAllocator.h
```diff
@@ -0,0 +1,42 @@
...
+  struct CompareDevice {
+    bool operator()(TDevice const& a, TDevice const& b) const { return getDevIndex(a) < getDevIndex(b); }
+  };
+
+  template <typename TDevice, typename TQueue>
+  inline CachingAllocator<TDevice, TQueue>& getCachingAllocator(TDevice const& device) {
+    static std::mutex mutex;
+    static std::map<TDevice, CachingAllocator<TDevice, TQueue>, CompareDevice<TDevice>> allocators;
+
+    auto keyval = allocators.find(device);
```
makortel, line 20: I believe the search needs to be protected with the `mutex` too. (e.g. one thread could be doing `find()` while other is doing `try_emplace()`).

I'm not sure if `std::shared_mutex` would be worth here, or `tbb::concurrent_unordered_map` (which would require defining a hash for the devices?).

src/alpaka/AlpakaCore/CachingAllocator.h
makortel: If enabled it could result in data race, toy example:
```cpp
auto d_buf = make_device_buffer<Foo>(queue);
{
  auto h_buf = make_host_buffer<Foo>(queue);
  // fill h_buf
  alpaka::memcpy(queue, d_buf, h_buf);
  // h_buf goes out of scope and the memory block is returned to the cache
}
{
  auto h_buf = make_host_buffer<Foo>(queue);
  // h_buf can be the same memory block as in earlier block
  // writing to h_buf before the memcpy() (or any other operation) has completed results in data race
}
```

fwyzard: I see.

The same does not happen for a device buffer because all operations there are only scheduled along the queue, right ?

README.md
```diff
@@ -364,6 +364,25 @@ $ make kokkos ...
...
 
 
+#### `alpaka`
+
+The `alpaka` code base is loosely based on the `cuda` code base, with some minor changes introduced during the porting.
+
+The use of caching allocator can be disabled at compile time setting the
+`ALPAKA_DISABLE_CACHING_ALLOCATOR` preprocessor symbol:
+```
+make cuda ... USER_CXXFLAGS="-DALPAKA_DISABLE_CACHING_ALLOCATOR"
```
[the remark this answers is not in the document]
fwyzard: Thanks, fixed.

README.md
```diff
@@ -364,6 +364,25 @@ $ make kokkos ...
...
+```
+
+If the caching allocator is disabled and CUDA version is 11.2 or greater is detected,
+device allocations and deallocations will use the stream-ordered CUDA functions
+`cudaMallocAsync` and `cudaFreeAsync`. Their use can be disabled explicitly at
+compile time setting also the `ALPAKA_DISABLE_ASYNC_ALLOCATOR` preprocessor symbol:
+
+```
+make cuda ... USER_CXXFLAGS="-DALPAKA_DISABLE_CACHING_ALLOCATOR -DALPAKA_DISABLE_ASYNC_ALLOCATOR"
```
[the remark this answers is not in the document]
fwyzard: Thanks, fixed.

src/alpaka/AlpakaCore/CachedBufAlloc.h
```diff
@@ -0,0 +1,78 @@
...
+    template <typename TElem, typename TDim, typename TIdx, typename TDev, typename TQueue, typename TSfinae = void>
+    struct CachedBufAlloc {
+      static_assert(alpaka::meta::DependentFalseType<TDev>::value, "This device does not support a caching allocator");
+    };
+
+    //! The caching memory allocator implementation for the CPU device
+    template <typename TElem, typename TDim, typename TIdx, typename TQueue>
+    struct CachedBufAlloc<TElem, TDim, TIdx, alpaka::DevCpu, TQueue, void> {
+      template <typename TExtent>
+      ALPAKA_FN_HOST static auto allocCachedBuf(alpaka::DevCpu const& dev, TQueue queue, TExtent const& extent)
```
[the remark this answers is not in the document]
fwyzard: Yes, `DevCpu` is always defined (though a corresponding accelerator like `AccCpuSerial` might not be).

Thanks, I'll the CUDA-related `#ifdef`s.

src/alpaka/AlpakaCore/CachedBufAlloc.h
```diff
@@ -0,0 +1,78 @@
...
+
+        auto& allocator = getCachingAllocator<alpaka::DevCpu, TQueue>(dev);
+
+        // FIXME the BufCpu does not support a pitch ?
+        size_t size = alpaka::extent::getExtentProduct(extent);
+        size_t sizeBytes = size * sizeof(TElem);
+        void* memPtr = allocator.allocate(sizeBytes, queue);
+
+        // use a custom deleter to return the buffer to the CachingAllocator
+        auto deleter = [alloc = &allocator](TElem* ptr) { alloc->free(reinterpret_cast<void*>(ptr)); };
```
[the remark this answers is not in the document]
fwyzard: Indeed, the compiler is happy without it.

src/alpaka/AlpakaCore/CachingAllocator.h
makortel: > The same does not happen for a device buffer because all operations there are only scheduled along the queue, right ?

Right.

src/alpaka/AlpakaCore/CachingAllocator.h
```diff
@@ -0,0 +1,424 @@
...
+      } catch (std::runtime_error const& e) {
+        // the allocation attempt failed: free all cached blocks on the device and retry
+        if (debug_) {
+          std::ostringstream out;
+          out << "\t" << deviceType_ << " " << alpaka::getName(device_) << " failed to allocate " << block.bytes
+              << " bytes for queue " << block.queue->m_spQueueImpl.get()
+              << ", retrying after freeing cached allocations" << std::endl;
+          std::cout << out.str() << std::endl;
+        }
+        // TODO implement a method that frees only up to block.bytes bytes
```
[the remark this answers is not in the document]
fwyzard: > Just to note that freeing many small blocks with total size of `block.bytes` does not necessarily help for a single allocation of `block.bytes` to succeed.

I think that should suffice if the blocks are large enough to free whole pages (which I think are 2MB in size, at least on my laptop with CUDA 11.6):
```
$ ./test 
        CUDA memory: 4014604288 / 4103208960 bytes

Allocate 1913 2 MB blocks
0       CUDA memory: 4014604288 / 4103208960 bytes
100     CUDA memory: 4012507136 / 4103208960 bytes
200     CUDA memory: 3802791936 / 4103208960 bytes
300     CUDA memory: 3593076736 / 4103208960 bytes
400     CUDA memory: 3383361536 / 4103208960 bytes
500     CUDA memory: 3173646336 / 4103208960 bytes
600     CUDA memory: 2963931136 / 4103208960 bytes
700     CUDA memory: 2754215936 / 4103208960 bytes
800     CUDA memory: 2544500736 / 4103208960 bytes
900     CUDA memory: 2334785536 / 4103208960 bytes
1000    CUDA memory: 2125070336 / 4103208960 bytes
1100    CUDA memory: 1915355136 / 4103208960 bytes
1200    CUDA memory: 1705639936 / 4103208960 bytes
1300    CUDA memory: 1495924736 / 4103208960 bytes
1400    CUDA memory: 1286209536 / 4103208960 bytes
1500    CUDA memory: 1076494336 / 4103208960 bytes
1600    CUDA memory:  866779136 / 4103208960 bytes
1700    CUDA memory:  657063936 / 4103208960 bytes
1800    CUDA memory:  447348736 / 4103208960 bytes
1900    CUDA memory:  237633536 / 4103208960 bytes
1913    CUDA memory:    2752512 / 4103208960 bytes

Free 16 random blocks
1897    CUDA memory:   36306944 / 4103208960 bytes

Allocate a 32 MB block
        CUDA memory:    2752512 / 4103208960 bytes
```

And, of course, it should be enough to free _one_ block _larger_ than what we need.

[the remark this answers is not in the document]
fwyzard: > Would it actually make sense to monitor the number of times this code path is taken? Ideally we would never use it, and it starting to be used could be a sign to take a deeper look on the application itself (thinking now CMSSW specifically).

Good suggestion, I'll see if I can add it to the statistics.

src/alpaka/AlpakaCore/alpakaMemory.h
fwyzard: mhm... I think I need to wrok a bit more on the combination of all the cases:
  - accelerator GPU memory (sync, async, cached) with a GPU accelerator queue
  - accelerator GPU memory (sync, async ?, cached ?) without a queue
  - accelerator CPU memory¹ (sync, async, cached) with a CPU accelerator queue
  - accelerator CPU memory¹ (sync, async ?, cached ?) without a queue
  - pinned host CPU memory (sync, async? cached) with a GPU accelerator queue
  - pinned host CPU memory (sync, async? cached?) without a queue
  - (pinned) host CPU memory (sync, async? cached?) with a CPU accelerator queue
  - (pinned) host CPU memory (sync, async? cached?) without a queue

¹ is for the case where the backend is a CPU accelerator, like TBB or serial; in these case what can we assume about race conditions between host and "device" memory access ?

src/alpaka/AlpakaCore/getCachingAllocator.h
fwyzard: mhm... maybe just an `std::vector<CachingAllocator>` or `std::unique_ptr<CachingAllocator[]>`, that we fill at the start of the job for all available devices ?
it adds some overhead for the devices that are available but not used - but the allocators don'r allocate any memory by themselves, so not much.

makortel: > maybe just an `std::vector<CachingAllocator>` or `std::unique_ptr<CachingAllocator[]>`, that we fill at the start of the job for all available devices ?

I agree that would probably be best.

fwyzard: The approach used by CUDA for `cudaMallocAsync` _etc._ seems to be that
  - a pool allocator deals with a single device
  - the allocation functions can use an explicit allocator, or fall back to a default per-device allocator
  
Even leaving out the option of using an explicit allocator for the time being, this would suggest that it makes sense to keep the current approach, where the allocator deals with a single device, and there is one allocator per device.

fwyzard:
Rebased after merging #302 and #307 .

image

src/alpaka/AlpakaCore/alpakaMemory.h
fwyzard: Splitting `getCachingAllocator` into `getHostCachingAllocator` and `getDeviceCachingAllocator` should help here as well.

fwyzard: > How about [...] to make it explicit that the code is intended to be the same? (same for the other two cases below)

OK, after the splitting of the host and device allocators,I think I've followed your suggestion for the fallback host allocations.

fwyzard:
Rebased after merging #308.

image

fwyzard:
The part about the statistics needs to be reorganised... the rest should be ready to go.

makortel (commented):
Otherwise looks good.

src/alpaka/AlpakaCore/getDeviceCachingAllocator.h
```diff
@@ -0,0 +1,47 @@
...
+    // one allocator per device
+    static std::vector<std::optional<CachingAllocator<TDevice, TQueue>>> allocators(devices.size());
+    auto& allocator = allocators[index];
+
+    // thread safe initialisation of the allocator, if needed
+    static std::mutex mutex;
+    if (not allocator) {
+      std::scoped_lock lock(mutex);
+      if (not allocator) {
+        allocator.emplace(devices[index],
```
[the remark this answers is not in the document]
fwyzard: > I suppose the simplest way would be to just construct the allocators for all devices

OK, done.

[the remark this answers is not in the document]
fwyzard: BTW, this did not have any observable impact on the performance, at least on a single GPU system.

fwyzard:
I've decided to descope the addition of the statistics to a follow up PR.

PR #302, opened 2022-02-02
[alpaka] Implement a host-only task running in the CUDA callback thread

fwyzard:


fwyzard:
For the moment this is a draft.
It works, but I want to understand if it can be optimised a bit.

makortel (commented):
Looks ok to me.

src/alpaka/AlpakaCore/HostOnlyTask.h
```diff
@@ -0,0 +1,55 @@
...
+                               void* arg) {
+        //ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(status);
+        std::unique_ptr<HostOnlyTask> pTask(static_cast<HostOnlyTask*>(arg));
+        (*pTask)();
+      }
+
+      ALPAKA_FN_HOST static auto enqueue(QueueUniformCudaHipRtNonBlocking& queue, HostOnlyTask task) -> void {
+        auto pTask = std::make_unique<HostOnlyTask>(std::move(task));
+        ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
+            ALPAKA_API_PREFIX(StreamAddCallback)(queue.m_spQueueImpl->m_UniformCudaHipQueue,
```
j-stephan, line 43: `cudaStreamAddCallback` is deprecated in current CUDA versions. This should probably be exchanged with something based on `cudaLaunchHostFunc`. Otherwise this approach looks fine to me.

fwyzard: I know, but we do want to be notified in case of errors in the CUDA stream.

PR #304, opened 2022-02-04
[alpakatest] Update AlpakaCore from alpaka, and split shared objects to backend-specific libraries

makortel:
Update of `AlpakaCore` from `alpaka` serves as a base for further prototyping (actually forward-porting of that).

The changes in `Makefile` for splitting the shared objects to backend-specific libraries should be straightforward to propagate to `alpaka` as well (perhaps in a separate PR?), I had just done it in `alpakatest` first (as part of other prototyping).


fwyzard:
The reduced number of memory allocations gives a small improvement in performance:

image

fwyzard (commented):
Overall the changes are fine.

For importing the AlpakaCore files from `alpaka` to `alpakatest`, it might be better to wait until #301 is merged (or redo the updated afterwards).

I still have to go through the `Makefile` changes.

makortel:
This PR is now updated to reflect he HEAD of master (as of merge of #305). The first force push was a pure rebase to `master`, second force push added updates from `alpaka/AlpakaCore`.

src/alpakatest/Makefile
```diff
@@ -40,26 +40,37 @@ define LIB_template
...
 $(1)_CUDA_OBJ := $$(patsubst $(SRC_DIR)%,$(OBJ_DIR)%,$$($(1)_PORTABLE_SRC:%=%.cuda.o))
 $(1)_CUDA_DEP := $$($(1)_CUDA_OBJ:$.o=$.d)
+$(1)_CUDA_LIB := $(LIB_DIR)/$(TARGET_NAME)/lib$(1)_cuda.so
+LIBS += $$($(1)_CUDA_LIB)
+$(1)_CUDA_LDFLAGS := -l$(1)_cuda
 #
 $(1)_CUOBJ := $$($(1)_CUDA_OBJ)
 $(1)_CUDADLINK := $$(if $$(strip $$($(1)_CUOBJ)),$(OBJ_DIR)/$(TARGET_NAME)/$(1)/lib$(1)_cudalink.o)
 endif
+endif # if PORTABLE_SRC is empty
```
fwyzard, line 72:
```suggestion
endif # if PORTABLE_SRC is not empty
```
?

fwyzard:
The changes for `alpakatest` are fine, however using the same build rules for `alpaka` fails with
```bash
$ g++ .../obj/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc.serial.o .../obj/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToClusterGPUKernel.cc.serial.o .../obj/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelFedCablingMapESProducer.cc.serial.o .../obj/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelGainCalibrationForHLTESProducer.cc.serial.o -O2 -fPIC -pthread -Wl,-E -lstdc++fs -ldl -shared -Wl,-z,defs -L.../lib/alpaka -lFramework -lAlpakaCore -lAlpakaDataFormats -lDataFormats -lCondFormats      -L.../external/tbb/lib -ltbb   -L.../external/boost/lib -L.../external/libbacktrace/lib -lbacktrace -L/usr/local/cuda/lib64 -lcudart -lcudadevrt -o .../lib/alpaka/pluginSiPixelClusterizer_serial.so

/opt/rh/gcc-toolset-10/root/usr/bin/ld: /opt/rh/gcc-toolset-10/root/usr/bin/ld: DWARF error: could not find variable specification at offset 4e137
.../obj/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc.serial.o: in function `alpaka_serial_sync::SiPixelRawToCluster::acquire(edm::Event const&, edm::EventSetup const&, edm::WaitingTaskWithArenaHolder)':
.../src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc:98: undefined reference to `ErrorChecker::ErrorChecker()'
/opt/rh/gcc-toolset-10/root/usr/bin/ld: .../src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc:120: undefined reference to `ErrorChecker::checkCRC(bool&, int, unsigned long const*, std::map<unsigned int, std::vector<SiPixelRawDataError, std::allocator<SiPixelRawDataError> >, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, std::vector<SiPixelRawDataError, std::allocator<SiPixelRawDataError> > > > >&)'
/opt/rh/gcc-toolset-10/root/usr/bin/ld: .../src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc:130: undefined reference to `ErrorChecker::checkHeader(bool&, int, unsigned long const*, std::map<unsigned int, std::vector<SiPixelRawDataError, std::allocator<SiPixelRawDataError> >, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, std::vector<SiPixelRawDataError, std::allocator<SiPixelRawDataError> > > > >&)'
/opt/rh/gcc-toolset-10/root/usr/bin/ld: .../src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc:139: undefined reference to `ErrorChecker::checkTrailer(bool&, int, unsigned int, unsigned long const*, std::map<unsigned int, std::vector<SiPixelRawDataError, std::allocator<SiPixelRawDataError> >, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, std::vector<SiPixelRawDataError, std::allocator<SiPixelRawDataError> > > > >&)'
collect2: error: ld returned 1 exit status
```

I think the reason is that `ErrorChecker` is implemented in `src/alpaka/plugin-SiPixelClusterizer/`, not in `src/alpaka/plugin-SiPixelClusterizer/alpaka/`:
```bash
$ ls -1 src/alpaka/plugin-SiPixelClusterizer/

alpaka/
ErrorChecker.cc
ErrorChecker.h
gpuClusterChargeCut.h
SiPixelFedIdsESProducer.cc
```

Thinking also about CMSSW, how do we want to deal with common code (in the sense that it does not depend on the backend) in a plugin directory ?
I guess it needs to be moved to a shared library ?

fwyzard:
OK, the `alpaka` code base can be fixed rearranging the `ErrorChecker` into a separate library:
```
renamed:    src/alpaka/plugin-SiPixelClusterizer/ErrorChecker.cc -> src/alpaka/SiPixelRawToDigi/ErrorChecker.cc
renamed:    src/alpaka/plugin-SiPixelClusterizer/ErrorChecker.h  -> src/alpaka/SiPixelRawToDigi/ErrorChecker.h
```

and then updating the other files accordingly:
```
modified:   src/alpaka/Makefile
modified:   src/alpaka/Makefile.deps
modified:   src/alpaka/plugins.txt
modified:   src/alpaka/plugin-SiPixelClusterizer/alpaka/SiPixelRawToCluster.cc
```

fwyzard:
I'll take this opportunity to draft what the allowed builds and dependencies could look like.
Currently we have in CMSSW:
```
Subsystem/
 └─ Package/
     ├─ interface/
     ├─ plugins/
     ├─ src/
     └─ test/
```

which could produce
  - `libSubsystemPackage.so`
  - `pluginSubsystemPackagePlugins.so`
  - some stuff under `test/` that I'm choosing to ignore for the moment

With the introduction of `alpaka` subdirectories we would have:
```
Subsystem/
 └─ Package/
     ├─ interface/
     │   └─ alpaka/
     ├─ plugins/
     │   └─ alpaka/
     ├─ src/
     │   └─ alpaka/
     └─ test/
         └─ alpaka/
```
and produce something like:
  - `libSubsystemPackage.so`
  - `libSubsystemPackage_cpu-serial.so`
  - `libSubsystemPackage_gpu-cuda.so`
  - `pluginSubsystemPackagePlugins.so`
  - `pluginSubsystemPackagePlugins_cpu-serial.so`
  - `pluginSubsystemPackagePlugins_gpu-cuda.so`
  - plus some more stuff under `test/`

That is, the code under `.../path/alpaka` is considered to be independent from the code under `.../path`.

The possible dependencies could be:
  - nothins can depend on a plugin, so _e.g._ `pluginSubsystemPackagePlugins_gpu-cuda.so` _cannot_ depend on `pluginSubsystemPackagePlugins.so`
  - an alpaka-enabled library automatically depends on its common part, so _e.g._ `libSubsystemPackage_gpu-cuda.so` depends on `libSubsystemPackage.so`
  - dependencies are declared for the package as a whole, not for the individual libraires, so a `BuildFile.xml` would still have `<use name="Subsystem/Package">` - _not_ `<use name="Subsystem/Package/alpaka">`
  - dependencies on an alpaka-enabled library automatically pick up both the standard library and the corresponding alpaka one, so _e.g._ if `Subsystem/Package/plugins` depends on `Subsystem/Package`, then `pluginSubsystemPackagePlugins_gpu-cuda.so` will depend on `libSubsystemPackage_gpu-cuda.so` and `libSubsystemPackage.so`

This are the most minimal set of changes I can think of right now... they exclude the possibility that a non-alpaka library or plugin may depend on an alpaka one; while this would be tempting, I think we already do not support this use case with CUDA (or, at least, I don't think we ever got it to work) so it doesn't seem to be a big problem.
If it turns out it would be useful to have, and we find a way to support it, we can always extend our rules later on.

src/alpakatest/Makefile
makortel: Right, thanks.

makortel:
> Thinking also about CMSSW, how do we want to deal with common code (in the sense that it does not depend on the backend) in a plugin directory ?
> I guess it needs to be moved to a shared library ?

I agree, in this model it would need to go into a shared library.

> This are the most minimal set of changes I can think of right now...

I pretty much agree with your proposal.

> they exclude the possibility that a non-alpaka library or plugin may depend on an alpaka one; 

I would strive for this approach also from dependency management point of view.

> while this would be tempting, I think we already do not support this use case with CUDA (or, at least, I don't think we ever got it to work) so it doesn't seem to be a big problem.

I might have misunderstood, but in CMSSW we do have libraries/plugins that themselves do not use CUDA but depend on libraries that use CUDA (that use only the runtime API, but not device code). For example, CPU-only EDModules (say DQM) reading SoA data formats that currently reside in CUDADataFormats.

Which then circles back to the treatment of DataFormats. If the data formats used by Alpaka EDModules are always accessed only by Alpaka EDModules, i.e. they are converted to e.g. legacy data formats for downstream EDModules, having data format classes for each backend should work fine.

But if we'd want to avoid that conversion, i.e. downstream EDModules could use directly the SoAs in (pinned) host memory produced by Alpaka EDModules, I believe the straightforward way would be to require these data formats to not depend on Alpaka or any device technology at build time (run time dependence e.g. in the form of `shared_ptr` deleter should still be fine).One option that just came to my mind could be to use backend-specific data formats within the Alpaka EDModule chains, but in the device-to-host transfer produce a generic, Alpaka/backend independent SoA that can be consumed by downstream EDModules without any extra dependencies. But I haven't thought this thoroughly. Anyway, at this point something to think about more.

fwyzard:
> > I think we already do not support this use case with CUDA (or, at least, I don't think we ever got it to work) so it doesn't seem to be a big problem.
> 
> I might have misunderstood, but in CMSSW we do have libraries/plugins that themselves do not use CUDA but depend on libraries that use CUDA (that use only the runtime API, but not device code). For example, CPU-only EDModules (say DQM) reading SoA data formats that currently reside in CUDADataFormats.

Sorry I wasn't clear, what I meant is non-CUDA (or non-Alpaka) libraries or plugins that make use of CUDA (or Alpaka) libraries that run GPU (accelerator) kernels.

In principle, it should be feasible to have a shared library that uses CUDA kernels internally, but only exposes a standard C/C++ API. In practice, if I remember correctly it ended up not working - and we don't have any uses cases for it at the moment, so we simply dropped it.

fwyzard:
> One option that just came to my mind could be to use backend-specific data formats within the Alpaka EDModule chains, but in the device-to-host transfer produce a generic, Alpaka/backend independent SoA that can be consumed by downstream EDModules without any extra dependencies. But I haven't thought this thoroughly. Anyway, at this point something to think about more.

Could you remind me what are the downsides of using the host-only part of alpaka in DataFormats ?
An alpaka (host) buffer is not very different from a `shared_ptr`.

makortel:
> > One option that just came to my mind could be to use backend-specific data formats within the Alpaka EDModule chains, but in the device-to-host transfer produce a generic, Alpaka/backend independent SoA that can be consumed by downstream EDModules without any extra dependencies. But I haven't thought this thoroughly. Anyway, at this point something to think about more.
> 
> Could you remind me what are the downsides of using the host-only part of alpaka in DataFormats ? An alpaka (host) buffer is not very different from a `shared_ptr`.

With `-DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_HOST_ONLY` Alpaka still has a compile-time dependence on CUDA (while the code can be compiled with the host compiler). Mixing code with `-DALPAKA_ACC_GPU_CUDA_ENABLED` and without leads to ODR violation because of
https://github.com/alpaka-group/alpaka/blob/654cda1c8bd349a09fb406826f5f5f6452b08f35/include/alpaka/mem/buf/BufCpu.hpp#L60-L62

For **transient** data formats using only CUDA backend along the CPU backends would probably technically work (as fine as current CUDA data formats). I'm more concerned on the evolution towards persistent data formats, and for more device backends (more thoughts on that below). I think we have a window of opportunity to try to be a bit more general now than minimally needed in an attempt to address those concerns to some degree, in the hope that spending a bit more effort for the generalization now would save more (migration) effort later. On the other hand, I think we can't solve everything upfront, and need to take one step at a time.

---

For **persistent** data formats I want to avoid build-time dependencies on specific device technologies. Potential problems of the dependencies includes e.g. some future host platform not supporting CUDA (e.g. because of CPU architecture, or compiler version; possibly temporarily). Or replace CUDA with HIP or DPC++. `DataFormats` having minimal dependencies has been beneficial also for FWLite and Fireworks (although that is moving towards web, but I presume they still use `DataFormats` in some way), and for I/O R&D.

For **transient** data formats we have much more flexibility, and I don't have specific potential problems in mind. Nevertheless it would feel a bit strange for downstream EDModule reading a host data product to have to depend on CUDA. It would require amending https://github.com/cms-patatrack/pixeltrack-standalone/pull/304#issuecomment-1041576835 with an ability for non-Alpaka package to depend on CPU-Serial version of Alpaka package, or having some other way to deal with CPU-Serial as a special case.

One complication is how would we extend the system to HIP and SYCL. Ok, currently CUDA and HIP are mutually exclusive, but maybe it will evolve. I took a quick look on the SYCL PR, and that appears to take a different approach (apparently also implying a different host buffer type?). Would the concept be that we enable all three device backends, and then `alpaka.hpp` brings compile-time dependence on all three (*)? Treating each device backend as parallel universe (that, to leading order, would never be loaded in the same program with the others) sounds to me like the path of least problems.

(*) `nvcc` is probably ok with HIP headers, I don't know what `hipcc` would think of CUDA headers. Or what DPC++ compiler would think of CUDA and HIP headers, or `nvcc` and `hipcc` on SYCL headers.

For that, I believe, all the host backend code would need to be independent of any device backend code. Or we'd always compile the host backends for each "device parallel universe" (leading to a somewhat similar shared object setup that use of Kokkos would imply). But even then the data formats passed on to downstream EDModules would need to be independent of any device platform.

fwyzard:
Thanks for the reminder and for the whole analysis !

I haven't read the SYCL-related PR in alpaka, yet, so I'll keep SYCL out of my comments - but do let me know if anything I write *does not* work with it.

> One complication is how would we extend the system to HIP ...

My idea is to replace `ALPAKA_HOST_ONLY` with a preprocessor symbol for each backend, similar to our `ALPAKA_ACC_GPU_CUDA_ASYNC_BACKEND`: so maybe `ALPAKA_ACC_GPU_CUDA_BACKEND`, `ALPAKA_ACC_GPU_HIP_BACKEND`, _etc._
Then `alpaka.hpp` will bring in the host-side API for all `ENABLED` accelerators, and the device-side parts (the I expect will need the accelerator-specific compiler) only for the specified `BACKEND`.
But, again, I have no idea if this can work with SYCL 🤔

This extension will require a way to handle pinning buffers for different accelerators; a simple option could be to pin for _all_ the enabled accelerators - but I have no idea if this can actually work :-/
Something cleaner would be to type-pun the pin/unpin functions, and move them to an `std::function<void(alpaka::Buf&)>` data member... then, depending on which accelerator is being targeted, the correct call can be made to pin the memory, and the corresponding call can be store to unpin it later.
What do you think ? I can try to check if this can work for the SYCL case, and discuss it with the alpaka developers next week.

This _might_ also have the benefit of making the host buffers independent from any third-party API, at least in the ODR sense, and possibly also in the link-time-dependency sense. I have to think it through.

---

If this doesn't work (for any reason) then it looks like we need to design our own simple host memory buffer, and use it for all host-side collections.

fwyzard:
There's one aspect that I hadn't considered until now: when reading products back from a root file, should they ever go into pinned memory ?  And then, what kind ?

makortel:
> Something cleaner would be to type-pun the pin/unpin functions, and move them to an `std::function<void(alpaka::Buf&)>` data member... then, depending on which accelerator is being targeted, the correct call can be made to pin the memory, and the corresponding call can be store to unpin it later.

I think this could work. Although for our use case I'd find it more natural to control the pinning on the allocator rather than in the buffer, but re-inventing a different wheel has also a cost (plus a possible counter argument below).

> There's one aspect that I hadn't considered until now: when reading products back from a root file, should they ever go into pinned memory ? And then, what kind ?

Good question. I think it would be interesting possibility to be able to read products directly to pinned memory (or pin those buffers later, assuming that operation would become cheaper synchronization-wise), but whether or not to do that should be somehow a property of the application reading the data (along "if the job uses a GPU, read products into pinned memory, but if it doesn't, read into regular memory"). Being able to (cheaply) pin on the fly would certainly make that easy (speaking for the `pin()` function in the buffer class).

PR #305, opened 2022-02-04
Migrate to tbb::task_group

### LINKED ISSUE #295: TBB removed tbb::task
`tbb::task` was deprecated and has been removed from oneAPI distribution (eg 2021.3)
The downloaded version of TBB (2020 update 3) still has it.

The `env.sh` script will initialize oneAPI if it is present on the system (`source /opt/intel/oneapi/setvar.sh`).  The build will pick up that version of TBB rather than the downloaded version (I'm not sure how - the oneapi path is not present in the command lines printed out during the makefile run.  Probably through the `CPATH` environment variable.)

The issue shows up when trying to build the serial version (and other versions) with the error
```
                 from /home/mdewing/physics/hep/tracking/patatrack/pixeltrack/pixeltrack-standalone/src/serial/bin/EventProcessor.cc:1:
/home/mdewing/physics/hep/tracking/patatrack/pixeltrack/pixeltrack-standalone/src/serial/Framework/WaitingTask.h:37:44: error: expected class-name before ‘{’ token
   37 |   class WaitingTask : public tbb::task {
      |                                            ^
...
```
A workaround is to remove the oneapi set up line in `env.sh`

What is the longer term plan for dealing with this?  Move to `task_group`?

makortel:
Resolves #295. Following the approach in cms-sw/cmssw#32804.


fwyzard:
All the backends I tested (all but `hip` and `kokkos`) compile, run, and pass the validation.

With 8 or more threads, the performance is the same as before.
With less than 8 threads the performance is visibly worse than before:

image

Is this expected ?

makortel:
The culprit appears to be this spinloop
https://github.com/cms-patatrack/pixeltrack-standalone/blob/796f790fa6763b6bea0e8ed3b58f297a8e0d906f/src/alpaka/bin/EventProcessor.cc#L41-L43
which is needed to guarantee the program won't terminate while there is no other work than waiting for an asynchronous work launched from ExternalWork `acquire()` to complete. `perf` showed high load on this, and the "best" way to hit to this code path is one-thread (or low-thread) jobs, where the impact on throughput is visible.

I gave a try to the "deferred task" preview feature of recent oneTBB versions (as suggested to us in https://github.com/oneapi-src/oneTBB/issues/346#issuecomment-1023036397), and that seemed to restore the performance (at least within ~2 %  in my limited tests). I used oneTBB v2021.4.0 (which we use in CMSSW_12_3_X) for that test. While exploring this option I noticed GCC 8.4.0 to crash when compiling oneTBB v2021.4.0, while GCC 9.3.0 worked fine. I'd suggest we just bump the minimum GCC version to 9; I suppose we don't have hard requirement to keep supporting GCC 8 (I've kept that because that was used in the CMSSW version the `cuda` code was originally exported).

I had been thinking to update the oneTBB anyway after this PR. The main question is then if it would be better to move to recent oneTBB update and use of deferred task already in this PR, or in two subsequent PRs. I don't have strong feelings at this point. @fwyzard, which would you prefer?

fwyzard:
@makortel thanks for the investigation !

I would make one PR to bump the GCC requirement, and them update this PR to include the newer version of TBB and the use of deferred tasks.

makortel:
> I would make one PR to bump the GCC requirement, and them update this PR to include the newer version of TBB and the use of deferred tasks.

Sounds good, I'll prepare another PR to bump the GCC requirement.

makortel:
Updates
* #246 missed `cudatest`, so it is updated here
* Update TBB to v2021.4.0
* Replace the [aforementioned](https://github.com/cms-patatrack/pixeltrack-standalone/pull/305#issuecomment-1036294829) spinloop with deferred tasks

src/alpaka/Framework/WaitingTaskWithArenaHolder.cc
```diff
@@ -21,9 +21,23 @@ namespace edm {
   // Note that the arena will be the one containing the thread
   // that runs this constructor. This is the arena where you
   // eventually intend for the task to be spawned.
-  WaitingTaskWithArenaHolder::WaitingTaskWithArenaHolder(WaitingTask* iTask)
-      : m_task(iTask), m_arena(std::make_shared<tbb::task_arena>(tbb::task_arena::attach())) {
+  WaitingTaskWithArenaHolder::WaitingTaskWithArenaHolder(tbb::task_group& iGroup, WaitingTask* iTask)
+      : m_task(iTask), m_group(&iGroup), m_arena(std::make_shared<tbb::task_arena>(tbb::task_arena::attach())) {
     m_task->increment_ref_count();
+    m_handle = std::make_shared<tbb::task_handle>(m_group->defer([task = m_task]() {
```
makortel, line 27: I'm not terribly happy with the approach I took for `tbb::task_handle` here, e.g. there are now three members doing reference counting (two `shared_ptr`s implicitly and `m_task` explicitly), but I'd leave improvements (likely a redesign) to a later time.

src/alpaka/Framework/WaitingTaskWithArenaHolder.cc
```diff
@@ -73,9 +94,7 @@ namespace edm {
     auto task = m_task;
     m_task = nullptr;
     if (0 == task->decrement_ref_count()) {
-      // The enqueue call will cause a worker thread to be created in
-      // the arena if there is not one already.
```
makortel, line 77: Whoops, this comment should've stayed.

src/alpaka/Framework/WaitingTaskHolder.h
```diff
@@ -72,20 +95,29 @@ namespace edm {
...
       // try to reuse this object. Resetting
       // before spawn avoids problems
       auto task = m_task;
       m_task = nullptr;
       if (0 == task->decrement_ref_count()) {
-        tbb::task::spawn(*task);
+        m_group->run([task]() {
+          TaskSentry s{task};
+          task->execute();
+        });
```
fwyzard, line 108: Trying to understand what this pattern does:
  - `m_group` is a pointer to the `tbb::task_group` that will run the task
  - `m_task` is the a pointer to the `WaitingTask` to be run
  - `task` is a copy of `m_task` (the comment above explains why the extra copy is needed)
  - `m_group` takes a functor (here a lambda) to be run
  - the lambda makes a copy of `task` 
  - the `TaskSentry` takes ownership of the `task`, guaranteeing that it will be deleted after it has run (also in case an exception is thrown)
  - the task is executed inside the lambda inside the `task_group`
  
The only confusing aspect is that `TaskSentry` will delete the task once the lambda is complete, except if it is a `FinalWaitingTask`, in which case the destructor of the `TaskSentry` effectively does not do anything.

fwyzard:
With these latest changes, the performance using `tbb::task_group` is the same as using `tbb::task`:
image

Zoomed-in version of the CUDA performance:
image

Zoomed-in version of the CPU performance:
image

Note that these plots were made using the `tbb::spin_mutex` in the caching allocators.

fwyzard:
@makortel do you prefer to keep the individual commits or squash them ?

makortel:
Thanks a lot Andrea for the detailed checks!

> do you prefer to keep the individual commits or squash them ?

In general I've been trying to touch only one test program code in one commit, and I'd like to keep that. I'd like to also separate "migrate to `tbb::task_group`", "TBB update", and "use of deferred tasks" as separate sets of commits. But I see now that I have some spurious "ScopedContext" etc commits that I intended to squash, so let me fix those.

makortel:
> But I see now that I have some spurious "ScopedContext" etc commits that I intended to squash, so let me fix those.

Done now.

src/alpaka/Framework/WaitingTaskHolder.h
makortel: That difference comes from objects of all other task types than `FinalWaitingTask` being allocated with `new` and intended to be deleted right after the task execution has finished. The `FinalWaitingTask` object is allocated on the stack of the function that initiates the "asynchronous region" and does a wait on the `tbb::task_group`. It must be kept alive until all the asynchronous tasks of the region have finished (e.g. to check if exception was thrown).

PR #306, opened 2022-02-07
Use Kokkos partition_space

markdewing:
Use Kokkos partition_space to abstract stream management on Cuda and HIP platforms.

This enables multiple streams to work on HIP platforms.

Remove code related to stream management: SharedStreamPtr, StreamCache, ExecSpaceWrapper.


fwyzard:
What impact do these changes have on the runimte performance ?

makortel:
Here is a test on V100
kokkos_cuda_throughput
fullsocket_kokkos_serial_throughput

PR #307, opened 2022-02-07
[alpaka] Restructure TrackingRecHit2DAlpaka with SoA's like the cuda version

fwyzard:
Replaces the 14 memory buffers used in the `alpaka` version with 2 larger memory buffers, as is done in the `cuda` version.


fwyzard:
image

fwyzard:
The improvement is even greater when combined with #301:
image

PR #308, opened 2022-02-09
[alpaka] Restructure CAHitNtupletGeneratorKernels with SoA's like the cuda version

fwyzard:


fwyzard:
The reduced number of memory allocations gives a small improvement in performance:

image

PR #309, opened 2022-02-11
[kokkos] Disable Desul atomics temporarily to improve performance

makortel:
The Kokkos update to 3.5 turned out to decrease the throughput by 2-3x. Disabling "Desul atomics" (that Kokkos was changed to use by default in 3.5) appears to cure the issue, but this is only a temporary workaround, because the "old atomics implementation" is going to be phased out. The issue is being followed up with Kokkos developers in https://github.com/kokkos/kokkos/issues/4780.

In addition, this PR adds support for using Kokkos' profiling tools via the `KOKKOS_PROFILE_LIBRARY` environment variable. (functionality that we were missing because of heavily customized initialization of Kokkos).,


makortel:
Superseded by #333 (but I'll close only after testing that one for performance)

makortel:
Based on discussion in #333 proceeding whit this one for now after all.

PR #310, opened 2022-02-11
Bump minimum GCC version to 9

makortel:
Needed in order to update oneTBB to something recent. E.g. v2021.4.0 lead to GCC 8.4.0 to crash during compilation (not really worth of further investigation). See #305 for more details.


makortel:
@fwyzard Could you double-check the update in README?

fwyzard:
Is the minimum 9.0.0 or some later version ?

makortel:
I tested now 9.1.0 (from LCG) and that worked. On the other hand, TBB claims to support 
> GNU Compilers (gcc) 4.8.5 - 11.1.1
https://github.com/oneapi-src/oneTBB/blob/master/SYSTEM_REQUIREMENTS.md

I tested now again with 8.4.0, and was unable to reproduce the earlier crash when compiling TBB. So I was too hasty with the conclusion in https://github.com/cms-patatrack/pixeltrack-standalone/pull/305#issuecomment-1036294829.

Since there is no strict need to bump the minimum GCC version, I'll just close this PR.

PR #312, opened 2022-02-14
[alpaka] Sync TB host copies only when needed

fwyzard:
Synchronise the queues and wait for the host-device copies to complete only when using the TBB async backend, without the caching allocator.


fwyzard:
Recovers some of the performance when using the CUDA backend with a large number of threads:
image

fwyzard:
Thanks to @waredjeb for pointing this out !

PR #313, opened 2022-02-15
[cuda] Minimal changes taking advantage of C++17 features

fwyzard:


makortel (approved):
(on visual inspection)

fwyzard:
Builds and runs fine...

Just to clarify: the main reason for these changes is to make it easier to test different `mutex` types in the caching allocators.
Then I had the `constexpr` changes lying around for a while, so I included them as well.

PR #321, opened 2022-02-25
[cudadev][RFC] Prototype edm::async() and replace cudaStreamAddCallback with it

makortel:
This PR prototypes `edm::async()` planned in https://github.com/cms-sw/cmssw/issues/29188, and use it in `cudadev` instead of `cudaStreamAddCallback()`.

I made a preliminary test on RTX 2080 (single ~1 minute run on each case, CPU utilization is peak of those given reported by `ps` checked every 10 s). I tested three cases, `cudadev`, `cudadev --transfer`, and `cudadev --histogram` to see if behavior is different with more device-to-host transfers (i.e. synchronization calls) and CPU work.

I tried first to use `cudaStreamSynchronize()` with each of `cudaDeviceScheduleSpin`, `cudaDeviceScheduleYield`, and `cudaDeviceScheduleBlockingSync` (for `cudaSetDeviceFlags()`), but didn't see any substantial difference between them: they give a bit higher throughput at low number of threads, but come with significant use of CPU compared to `cudaStreamAddCallback()`.

Then I tried `cudaEventSynchronize()`, but in first test that didn't seem to improve (I didn't even bother to measure it). Adding `cudaEventBlockingSync` to the CUDA event creation flags finally lead to significantly lower CPU utilization, in case of minimal-transfer `cudadev` much lower than with `cudaStreamAddCallback()`. The throughput on low number of CPU threads takes up to 15 % hit compared to `cudaStreamAddCallback()` and `cudaStreamSynchronize()` (depending on the test case).

image
image
image

Given that I didn't repeat the measurements, the uncertainty may be high, and therefore I think the measurements should be repeated. I'm planning to do that, but didn't want to hold the prototype for those.

### DIFF
--- a/src/cudadev/CUDACore/ScopedContext.cc
+++ b/src/cudadev/CUDACore/ScopedContext.cc
@@ -56,20 +56,20 @@ namespace cms::cuda {
     void ScopedContextHolderHelper::enqueueCallback(int device, cudaStream_t stream) {
       SharedEventPtr event = getEventCache().get();
       cudaEventRecord(event.get(), stream);
-      edm::async(waitingTaskHolder_, [device, stream, event=std::move(event)]() mutable {
-          auto status = cudaEventSynchronize(event.get());
-          event.reset();
-          if (status != cudaSuccess) {
-            auto error = cudaGetErrorName(status);
-            auto message = cudaGetErrorString(status);
-            throw std::runtime_error("Callback of CUDA stream " +
-                                     std::to_string(reinterpret_cast<unsigned long>(stream)) + " in device " +
-                                     std::to_string(device) + " error " + std::string(error) + ": " + std::string(message));
-          } else {
-            //std::cout << " GPU kernel finished (in callback) device " << device << " CUDA stream "
-            //          << stream << std::endl;
-          }
-        });
+      edm::async(waitingTaskHolder_, [device, stream, event = std::move(event)]() mutable {
+        auto status = cudaEventSynchronize(event.get());
+        event.reset();
+        if (status != cudaSuccess) {
+          auto error = cudaGetErrorName(status);
+          auto message = cudaGetErrorString(status);
+          throw std::runtime_error(
+              "Callback of CUDA stream " + std::to_string(reinterpret_cast<unsigned long>(stream)) + " in device " +
+              std::to_string(device) + " error " + std::string(error) + ": " + std::string(message));
+        } else {
+          //std::cout << " GPU kernel finished (in callback) device " << device << " CUDA stream "
+          //          << stream << std::endl;
+        }
+      });
     }
   }  // namespace impl
 
--- a/src/cudadev/CUDACore/ScopedContext.cc
+++ b/src/cudadev/CUDACore/ScopedContext.cc
@@ -54,19 +54,22 @@ namespace cms::cuda {
     }
 
     void ScopedContextHolderHelper::enqueueCallback(int device, cudaStream_t stream) {
-      edm::async(waitingTaskHolder_, [device, stream]() {
-        auto status = cudaStreamSynchronize(stream);
-        if (status != cudaSuccess) {
-          auto error = cudaGetErrorName(status);
-          auto message = cudaGetErrorString(status);
-          throw std::runtime_error(
-              "Callback of CUDA stream " + std::to_string(reinterpret_cast<unsigned long>(stream)) + " in device " +
-              std::to_string(device) + " error " + std::string(error) + ": " + std::string(message));
-        } else {
-          //std::cout << " GPU kernel finished (in callback) device " << device << " CUDA stream "
-          //          << stream << std::endl;
-        }
-      });
+      SharedEventPtr event = getEventCache().get();
+      cudaEventRecord(event.get(), stream);
+      edm::async(waitingTaskHolder_, [device, stream, event=std::move(event)]() mutable {
+          auto status = cudaEventSynchronize(event.get());
+          event.reset();
+          if (status != cudaSuccess) {
+            auto error = cudaGetErrorName(status);
+            auto message = cudaGetErrorString(status);
+            throw std::runtime_error("Callback of CUDA stream " +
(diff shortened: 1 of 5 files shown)


fwyzard:
I'd be curious if the CPU usage from top matches the values from https://github.com/cms-patatrack/pixeltrack-standalone/pull/324 ?

makortel:
I re-ran subset of the test cases with #324 and the `ps`-based monitoring, and plotted the thread efficiency (i.e. the number reported by #324)
image
image
image

In majority of the cases #324 gives a slightly larger value, but I'd say overall the two methods are in rather good agreement.

fwyzard:
Thanks for the check !
I'll extend #324 to all backends and merge it.

src/cudadev/CUDACore/ScopedContext.h
```diff
@@ -227,11 +227,10 @@ namespace cms {
     namespace impl {
       template <typename F>
       void ScopedContextHolderHelper::pushNextTask(F&& f, ContextState const* state) {
-        replaceWaitingTaskHolder(edm::WaitingTaskWithArenaHolder{
-            edm::make_waiting_task_with_holder(std::move(waitingTaskHolder_),
-                                               [state, func = std::forward<F>(f)](edm::WaitingTaskWithArenaHolder h) {
-                                                 func(ScopedContextTask{state, std::move(h)});
-                                               })});
+        replaceWaitingTaskHolder(edm::WaitingTaskWithArenaHolder{edm::make_waiting_task_with_holder(
+            std::move(waitingTaskHolder_), [state, func = std::forward<F>(f)](edm::WaitingTaskWithArenaHolder h) {
```
fwyzard, line 231: this is just a change in formatting ?

src/cudadev/bin/main.cc
```diff
@@ -104,11 +104,29 @@ int main(int argc, char** argv) {
...
+    if (cudaSuccess != status) {
+      std::cout << "Failed to set device status flags" << std::endl;
+      return EXIT_FAILURE;
+    }
+  }
+  status = cudaSetDevice(0);
+  if (cudaSuccess != status) {
+    std::cout << "Failed to set device" << std::endl;
+    return EXIT_FAILURE;
+  }
```
fwyzard, line 128: do we care to re-set the current device to 0 ?

src/cudadev/CUDACore/ScopedContext.cc
```diff
@@ -81,8 +54,22 @@ namespace cms::cuda {
...
+        if (status != cudaSuccess) {
+          auto error = cudaGetErrorName(status);
+          auto message = cudaGetErrorString(status);
+          throw std::runtime_error(
+              "Callback of CUDA stream " + std::to_string(reinterpret_cast<unsigned long>(stream)) + " in device " +
+              std::to_string(device) + " error " + std::string(error) + ": " + std::string(message));
+        } else {
+          //std::cout << " GPU kernel finished (in callback) device " << device << " CUDA stream "
+          //          << stream << std::endl;
+        }
```
fwyzard, line 71: This could simply be
```suggestion
        cudaCheck(cudaEventSynchronize(event.get()));
```
except for the mention of the callback in the error message ?

src/cudadev/CUDACore/ScopedContext.cc
```diff
@@ -81,8 +54,22 @@ namespace cms::cuda {
     }
 
     void ScopedContextHolderHelper::enqueueCallback(int device, cudaStream_t stream) {
-      cudaCheck(
-          cudaStreamAddCallback(stream, cudaScopedContextCallback, new CallbackData{waitingTaskHolder_, device}, 0));
+      SharedEventPtr event = getEventCache().get();
+      cudaEventRecord(event.get(), stream);
+      edm::async(waitingTaskHolder_, [device, stream, event = std::move(event)]() mutable {
```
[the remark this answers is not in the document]
makortel: Thanks for point it out, this indeed could use `std::move()` (both of the calls to `enqueueCallback()` come from destructors of classes with `ScopedContextHolderHelper` member, and therefore nothing will use the the `waitingTaskHolder_`).

Copying is technically tine too (just a little bit less efficient) because indeed necessary internals are shared among copies and reference counted. The framework also holds one copy with a lifetime a tiny bit longer than the `EDModule::acquire()` call; to guarantee that the `produce()` can not be called too early.

src/cudadev/CUDACore/ScopedContext.cc
makortel: I agree, I wasn't sure if the "Callback on device X" would be useful information or not, so I decided to keep the message from the `cudaStreamAddCallback()` version. Ok, reminding myself of `cudaCheck()` again I could use the `description` argument for that information. In that case I'd be tempted to change the `description` to `std::string_view` (following similar change in CMSSW). On the other hand, using `description` argument would imply (unnecessarily) constructing the error message also for the successful cases

src/cudadev/CUDACore/ScopedContext.h
makortel: Good catch, it indeed seems so. I'll take it away from not creating confusion here (as this PR does not include any other formatting changes).

src/cudadev/bin/main.cc
makortel: Nothing in our code should care it, i.e. we should always be setting the device explicitly. I'd be happy to remove this `cudaSetDevice()`. On the other hand, it might be worth to test also if with `cudaEventBlockingSync` this `cudaSetDeviceFlags(cudaDeviceScheduleBlockingSync)` has any effect, and if not, just remove all changes in this file.

src/cudadev/CUDACore/ScopedContext.h
makortel: The latest push removed these changes.

makortel:
This idea was integrated into CMSSW in https://github.com/cms-sw/cmssw/pull/44901, so I think this PR can be closed.

PR #322, opened 2022-02-26
[alpaka] Split shared objects to backend-specific libraries

fwyzard:
Move common code shared between plugins to a shared library.


fwyzard:
@makortel we don't expect any impact on the performance from this, do we ?

fwyzard:
However, `alpakatest` does not seem to be working either:
```
$ ./alpakatest --cuda
Found 0 devices.
Floating point exception (core dumped)
```

fwyzard:
OK, that was probably a hardware problem, since just running `nvidia-smi` seems to have crashed the machine :-/

makortel:
> we don't expect any impact on the performance from this, do we ?

In the leading order there should no performance impact. In the sub-leading order shared library reorganization _might_ result in performance differences.

Visually looks good to me.

fwyzard:
OK, now it runs fine, and the perfornance and CPU usage look unchanged:
image
image

PR #323, opened 2022-02-27
Abort on the first compilation error

fwyzard:
`-Wfatal-errors` makes GCC stop after the first compilation error.


fwyzard:
I find this useful to avoid the compiler wasting too much time after the first error, and to spot the first error more easily, since it is often the cause (or a repetirion) of many subsequent failures.

@makortel what do you think ?

makortel (approved):
I'm fine with giving it a try.

In some occasions I have found the compiler listing multiple different errors on the same file useful, but I think all those have been in context of mass migrations in CMSSW, which is unlikely to be useful here.

PR #324, opened 2022-02-27
Monitor the CPU usage of the event loop

fwyzard:


fwyzard:
Ideally this should give a result similar to `top`, but normalised to the number of threads being used.
On the other hand, it may give too much importance the impact of the initialisation part.

@makortel what do you think ?

Maybe we should add a way to "warm up" the program, skipping the first N events ?

fwyzard:
@makortel @markdewing I wasn't able to build with Kokkos; do you want to test the changes before merging this PR, or do you trust it to work the same as for the other backends ?

fwyzard:
Yes, I skipped it because I didn't have a way to test it at the moment.

makortel:
`kokkos` and `kokkostest` build and work fine with this PR.

I would nevertheless propagate the update to `sycltest`. I think the update working in all other cases makes it likely to work for `sycltest` too, and fixing any (possible) compilation errors would be more straightforward than figuring out which infrastructure updates are missing.

fwyzard:
OK, will do.

PR #326, opened 2022-03-02
Update Boost to version 1.78.0

fwyzard:
Building `hip` and `hiptest` on CentOS 7 with GCC 9.2 requires Boost 1.78 or later, in order to pick the fix for https://github.com/boostorg/config/issues/392 .

Tested with various versions of ROCm from 4.0.0 to 5.0.1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,8 +133,8 @@ export EIGEN_LDFLAGS :=
 export EIGEN_NVCC_CXXFLAGS := --diag-suppress 20014
 
 BOOST_BASE := /usr
-# Minimum required version of Boost, e.g. 1.73.0
-BOOST_MIN_VERSION := 107300
+# Minimum required version of Boost, e.g. 1.78.0
+BOOST_MIN_VERSION := 107800
 # Check if an external version of Boost is present and recent enough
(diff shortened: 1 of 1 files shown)


fwyzard:
@makortel is this OK for you ?

By the way, the version of Boost used in CMSSW was recently bumped to 1.78.0, too.

makortel:
Visually yes, let me recompile bunch of stuff with this.

What about reverting the workaround added in https://github.com/cms-patatrack/pixeltrack-standalone/pull/207, would that be better done here or as a separate PR?

fwyzard:
I'm just testing that, and if it works I'd do it in a separate PR.

fwyzard:
Yes, with Boost 1.78.0 it compiles (and runs) fine also after reverting #207.

fwyzard:
By the way, I'm using the sourceforge.net mirrors to get Boost (_e.g._ https://sourceforge.net/projects/boost/files/boost/1.78.0/boost_1_78_0.tar.bz2) because the original location times out for me.

And I've mirrored the mirror at /cvmfs/patatrack.cern.ch/externals/src/boost/download/ ...

makortel (approved):
Seems to work fine.

PR #327, opened 2022-03-03
Revert "Fix compile issue with BOOST_NOINLINE and HIP"

fwyzard:
This reverts commit 4ad85d8c368658629244122ef48a5eca1b63555c.

The workaround is no longer necessary after the move to Boost 1.78.0.


markdewing (approved):

markdewing:
The change works on my machine with the updated boost.

PR #329, opened 2022-03-07
Take internal directories from variables instead of hardcoding

makortel:
This pattern is already followed everywhere else in the Makefile. I encountered this while overriding the `OBJ_DIR` from command line and wondering why `make clean` didn't remove the pointed-to directory.


makortel:
@fwyzard Do you want to take a look?

fwyzard:
Looks OK to me.

PR #330, opened 2022-03-07
[alpaka] Initialize devices in many alpaka unit test

makortel:
Fixes assertion failure in DeviceCachingAllocator. With this PR all the unit tests in `alpaka` run successfully for Serial, TBB, and CUDA backends.

### DIFF
--- a/src/alpaka/test/alpaka/AtomicPairCounter_t.cc
+++ b/src/alpaka/test/alpaka/AtomicPairCounter_t.cc
@@ -5,6 +5,7 @@
 #include "AlpakaCore/alpakaConfig.h"
 #include "AlpakaCore/alpakaMemory.h"
 #include "AlpakaCore/alpakaWorkDiv.h"
+#include "AlpakaCore/initialise.h"
 
 using namespace cms::alpakatools;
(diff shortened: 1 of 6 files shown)


fwyzard:
The changes look good - but we need the same changes also in `src/alpaka/test/alpaka/HistoContainer_t.cc`, `src/alpaka/test/alpaka/clustering_t.cc`, _etc._.

Maybe we need a more robust solution, that doesn't require an explicit initialisation ?

fwyzard (commented):
The changes look good - but we need the same changes also in `src/alpaka/test/alpaka/HistoContainer_t.cc`, `src/alpaka/test/alpaka/clustering_t.cc`, _etc._.

Maybe we need a more robust solution, that doesn't require an explicit initialisation ?

makortel:
> The changes look good - but we need the same changes also in `src/alpaka/test/alpaka/HistoContainer_t.cc`, `src/alpaka/test/alpaka/clustering_t.cc`, _etc._.

Thanks, somehow I missed those.

makortel:
Done now. `make test_alpaka` runs (although so I thought before too).

fwyzard:
:+1:

Issue #297, opened 2022-01-21
Performance of kokkos serial backend vs. plain serial code


markdewing:
The time from the kokkos serial backend (`kokkos --serial`) is slower for one thread than the standalone serial code (`serial`).

Looking at profiles, the function `kernel_connect` ( in `plugin-PixelTriplets`) takes significantly more time in the kokkos version.   From the instructions retired, it is clearly performing more operations in the kokkos version. 

The loops do not perform their iterations in the same order.
```
<outer loop from kokkos>
  <kernel_connect function body>
     for (int idx = firstCellIndex, nt = (nCells()); idx < nt; idx += leagueSize * blockDim) {
     ...
       for (int j = first; j < numberOfPossibleNeighbors; j += stride) {
```
The serial version has no outer loop.   However, printing the values for `idx` and `j` shows the same values get accessed, just in a different order between the versions.

It looks like (based on instructions retired), the kokkos version is doing more work (by a factor of 2x or more), in routines like areAlignedRZ.  But based on printing out how many times it's called, they should be the same.

markdewing:
In plugin-SiPixelClusterizer, there's a difference between the serial code and Kokkos code.
In  gpuClusterChargeCut.h, there is a loop that sets some variables to zero.

Kokkos version (around line 97)
```
         for (uint32_t i = teamMember.team_rank(); i < ::gpuClustering::MaxNumClustersPerModules;
               i += teamMember.team_size()) {
            charge(i) = 0;
            ok(i) = 0;
            newclusId(i) = 0;
          }
```

The serial version (around line 58)
```
      for (auto i = threadIdx.x; i < nclus; i += blockDim.x) {
        charge[i] = 0;
      }
```

Two reasons the Kokkos version is doing more work:
* the serial version of the loop bound is `nclus` (about 10) but the Kokkos version loop bound is MaxNumClustersPerModules (1024).
* the Kokkos version also sets `ok` and `newclusid` to zero.  These get set in a later loop, I'm not sure this is necessary.

If the Kokkos loop is changed to match the serial loop, the performance becomes similar.

Notes
 * The starting value (0) and increment (1) of the loops is the same for both.
 * The performance is similar at least for a sufficiently large number of events - there must be some startup overhead to Kokkos.  1000 events is sufficient - both get 27-28 events/s on my machine.  Using 100 events shows no drop in performance of serial code (27 events/s) where the kokkos serial version clearly gets lower performance (24.5 events/s)
[...]

PR #338, opened 2022-03-10
[kokkos] serial backend performance fix

markdewing:
The Kokkos serial backend is slower than the plain serial code  - see #297 for more details.

One cause is addressed here, where the Kokkos backend loops over the entire array (size 1024) rather than the number of clusters (about 10).
This loop also sets ok and newclusId to zero.  Those get set later and these assignments can be removed.


makortel:
On Cori I see ~5 % improvement on `kokkos --serial` when running one process on otherwise empty node
kokkos_serial_throughput

1-thread case is now witin 4 % of the `serial` program.

On a fully loaded socket the improvement is smaller but still clearly visible

fullsocket_kokkos_serial_throughput

PR #343, opened 2022-03-17
[alpaka] Remove unneeded variables and rules from the Makefiles

fwyzard:
Replace `ALPAKA_CUFLAGS` with `CUDA_CUFLAGS`, since dropping some of the compiler options is no longer needed.

Drop the rules for building .cu files, which are not used with Alpaka.


makortel (approved):
Looks visually ok.

PR #344, opened 2022-03-17
[cupla] Drop support for building with Cupla

fwyzard:


fwyzard:
Currently includes #343 to avoid conflicts with it.

makortel (approved):
Looks visually ok

PR #347, opened 2022-03-18
HIP/ROCm updates

fwyzard:
Update the ROCm Makefile variables:
  - rename the `HIP` tool to `ROCM`, and all `HIP_...` Makefile variables to `ROCM_...` .
  - take the ROCm clang paths from `clang -print-resource-dir` and `clang -print-runtime-dir` instead of `hipconfig`.

Update the list of failing test for the `hip` application.
Update the `hiptest` common files from `hip`.


makortel (approved):
Looks visually ok.

PR #352, opened 2022-03-27
Replace the hipLaunchKernelGGL macro with the CUDA chevron syntax

fwyzard:


markdewing (approved):

README.md
```diff
@@ -47,8 +47,8 @@ In addition, the individual programs assume the following be found from the syst
...
 | `kokkos`     | :heavy_check_mark: | :white_check_mark: (1)      | :white_check_mark: (2) |                                                                                                                  |
-| `alpakatest` |                    | :heavy_check_mark:          |                        |                                                                                                                  |
-| `alpaka`     |                    | :heavy_check_mark:          |                        |                                                                                                                  |
+| `alpakatest` |                    | :heavy_check_mark:          | :heavy_check_mark:     |                                                                                                                  |
+| `alpaka`     |                    | :heavy_check_mark:          | :heavy_check_mark:     |                                                                                                                  |
```
makortel, line 51: I'd suggest to use `:white_check_mark:` for both CUDA and ROCm, and add a similar kind of footnote as for Kokkos, since both dependencies are optional (and currently exclusive).

makortel (commented):
Looks visually ok. IIRC the `hipLaunchKernelGGL()` was the result of `hipify-perl` at the time I ran it.

README.md
fwyzard: Ah, yes, I was going to ask what was the difference between :heavy_check_mark: and :white_check_mark: ...

PR #365, opened 2022-07-10
[hip] Simplify HIP compilation flags

fwyzard:
Looks like `CLANG_RESOURCE_DIR` and `CLANG_RUNTIME_DIR` are only used internally, so we don't need to pass them explicitly.

The use of `$(ROCM_BASE)/hip/include` and `$(ROCM_BASE)/hsa/include` is not necessary, because they are symlinked to (or from) `$(ROCM_BASE)/include`, and their use is deprecated in ROCm 5.2.


fwyzard:
@makortel I've tested the simplified flags building the `hip` and `alpaka` back-ends.
Let me know if we should check anything else.

PR #367, opened 2022-07-22
[stdpar] NVHPC support

esseivaju:
Update the `std::par` program to build with the `nvc++` compiler. Tested on zeus with

- NVHPC 22.5
- CUDA 11.7

# Changelist

- Update makefiles to use `nvc++`
- Workaround bugs in `nvc++`
- Dropped CPU version of some algorithms which did not work with the `nvc++` unified compilation model
- Port of `setHitsLayer` to `std::par`
- Update `_CUDACC__`, `__CUDA_ARCH__` macro to work with `nvc++`
- Temporarily disabled compilation of RiemannFit kernels and related tests, which do not compile with `nvc++` and `Eigen`

# Notes

- While this compiles, running the program and most tests results in runtime errors, even with the original `setHitsLayer`cuda kernel (illegal memory access on device). 
- Two translation units, `PixelRecHits.cu`and `CAHitNtupletGeneratorKernels.cu` can't be compiled with optimization because of bugs in `nvc++`. This has already been reported to Nvidia and should be fixed in the next compiler release.


src/stdpar/Framework/CountingIterator.h
```diff
@@ -0,0 +1,69 @@
+struct counting_iterator {
```
makortel, line 1: I had the impression the use of C++20 for `nvc++` would have provided the `counting_iterator` from the standard library. Did I understand wrong?

src/stdpar/Makefile
```diff
@@ -20,8 +20,15 @@ EXE_OBJ := $(patsubst $(SRC_DIR)%,$(OBJ_DIR)%,$(EXE_SRC:%=%.o))
...
 LIBNAMES := $(filter-out plugin-% bin test Makefile% plugins.txt%,$(wildcard *))
-PLUGINNAMES := $(patsubst plugin-%,%,$(filter plugin-%,$(wildcard *)))
-MY_CXXFLAGS := -I$(TARGET_DIR) -DLIB_DIR=$(LIB_DIR)/$(TARGET_NAME)
+
+# PLUGINNAMES are compiled with nvcc, PLUGINNAMES_NVCXX with nvc++
+# PLUGINNAMES_NVCXX := SiPixelRecHits
+# PLUGINNAMES := $(filter-out $(PLUGINNAMES_NVCXX),$(patsubst plugin-%,%,$(filter plugin-%,$(wildcard *))))
+
+PLUGINNAMES_NVCXX := $(patsubst plugin-%,%,$(filter plugin-%,$(wildcard *)))
+PLUGINNAMES :=
```
makortel, line 29: Do you see a need to keep the `PLUGINNAMES` around?

In principle I'd like to try out at some point what happens if one tries to limit the `nvc++` to the minimum (i.e. code with memory management or device code) and use `gcc` for the rest, but I think such a test would be better done later, and the build rules may be better to figure out then rather than thinking about them prematurely.

src/stdpar/plugin-PixelTriplets/HelixFitOnGPU.h
```diff
@@ -47,7 +47,7 @@ class HelixFitOnGPU {
   void launchRiemannKernels(HitsView const *hv, uint32_t nhits, uint32_t maxNumberOfTuples, cudaStream_t cudaStream);
   void launchBrokenLineKernels(HitsView const *hv, uint32_t nhits, uint32_t maxNumberOfTuples, cudaStream_t cudaStream);
 
-  void launchRiemannKernelsOnCPU(HitsView const *hv, uint32_t nhits, uint32_t maxNumberOfTuples);
+  //void launchRiemannKernelsOnCPU(HitsView const *hv, uint32_t nhits, uint32_t maxNumberOfTuples);
```
makortel, line 50: Would
```suggestion
#ifndef DISABLE_RFIT
  void launchRiemannKernelsOnCPU(HitsView const *hv, uint32_t nhits, uint32_t maxNumberOfTuples);
#endif
```
do the same?

src/stdpar/plugin-PixelTriplets/gpuPixelDoublets.h
```diff
@@ -99,30 +99,211 @@ namespace gpuPixelDoublets {
...
-                      phicuts,
-                      minz,
-                      maxz,
-                      maxr,
-                      ideal_cond,
-                      doClusterCut,
-                      doZ0Cut,
-                      doPtCut,
-                      maxNumOfDoublets);
+    // ysize cuts (z in the barrel)  times 8
```
makortel, line 109: (note to self) This is a direct copy from `gpuPixelDoubletsAlgos.h`.

Can that header be removed now?

src/stdpar/plugin-PixelVertexFinding/gpuSortByPt2.h
```diff
@@ -55,15 +56,15 @@ namespace gpuVertexFinder {
...
-#ifdef __CUDA_ARCH__
-    __shared__ uint16_t sws[1024];
-    // sort using only 16 bits
-    radixSort<float, 2>(ptv2, sortInd, sws, nvFinal);
-#else
-    for (uint16_t i = 0; i < nvFinal; ++i)
-      sortInd[i] = i;
-    std::sort(sortInd, sortInd + nvFinal, [&](auto i, auto j) { return ptv2[i] < ptv2[j]; });
-#endif
+    if target (nv::target::is_device) {
```
makortel, line 59: Just curious, is `if target` `nvc++` extension?

Makefile
```diff
@@ -26,8 +26,10 @@ endif
...
 export CXXFLAGS := -std=c++17 $(HOST_CXXFLAGS) $(USER_CXXFLAGS) -g
+export NVCXXFLAGS := -std=c++20 -O0 -cuda -gpu=managed -stdpar -fpic -gopt $(USER_CXXFLAGS)
 export LDFLAGS := -O2 -fPIC -pthread -Wl,-E -lstdc++fs -ldl
 export LDFLAGS_NVCC := -ccbin $(CXX) --linker-options '-E' --linker-options '-lstdc++fs'
+export LDFLAGS_NVCXX := $(NVCXXFLAGS) -Wl,-E -ldl
```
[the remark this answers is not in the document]
esseivaju: A few are, I can update it to only include necessary flags

src/stdpar/Framework/CountingIterator.h
esseivaju: You are correct, we can use [`std::views::iota`](https://en.cppreference.com/w/cpp/ranges/iota_view). I updated stdpar in 6553e97 but kept the iterator around, I'll remove it

src/stdpar/Makefile
esseivaju: `PLUGINNAMES` can be used to specifiy plugins that we want to compile with `gcc`. Since all of them can be compiled with `nvc++` I can remove it. 

`nvc++` has some restriction as to which code can be compiled with which compiler. Any code that does heap memory allocation that is to be used in a stdpar algorithm will need to be compiled by `nvc++`so that it can instrument the code to use unified memory.

src/stdpar/plugin-PixelVertexFinding/gpuSortByPt2.h
esseivaju: yes, this the `nvc++` alternative to `__CUDA_ARCH__`. `nvcc` also support this syntax. It is useful to generate different instructions for a function that is compiled for both device and host contexts or different device architecture. It has some limitation, mainly that it can only be used within function scope and can't leak into type system (e.g. change return type depending of arch).

src/stdpar/plugin-PixelTriplets/HelixFitOnGPU.h
esseivaju: I actually removed this function definition ins 126abbd46991d384eec8baaf030b76a0d9a7fc1a as it was confusing `nvc++`.

The problem here is that with the `nvcc` split compilation model, `__global__` was redefined in `cudaCompat.h` to work on cpu. With `nvc++`, we can't redefine `__global__`, so the kernel invocation in `launchRiemannKernelsOnCPU` (e.g. `kernelCircleFit<3>`) where causing compilation error as the compiler was expecting standard kernel invocation syntax with `<<<>>>`.

src/stdpar/CondFormats/PixelCPEFast.cc
```diff
@@ -17,17 +14,16 @@
...
 
-  cudaCheck(cudaMallocManaged(&m_params, sizeof(pixelCPEforGPU::ParamsOnGPU)));
-  cudaCheck(cudaMallocManaged(&m_commonParams, sizeof(pixelCPEforGPU::CommonParams)));
-  cudaCheck(cudaMallocManaged(&m_layerGeometry, sizeof(pixelCPEforGPU::LayerGeometry)));
-  cudaCheck(cudaMallocManaged(&m_averageGeometry, sizeof(pixelCPEforGPU::AverageGeometry)));
-
+  m_params = new pixelCPEforGPU::ParamsOnGPU;
+  m_commonParams = new pixelCPEforGPU::CommonParams;
+  m_layerGeometry = new pixelCPEforGPU::LayerGeometry;
+  m_averageGeometry = new pixelCPEforGPU::AverageGeometry;
```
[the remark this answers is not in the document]
esseivaju: Definitely better, I will update it. I was using new as a quick way to debug what was causing the illegal memory access as data members where raw pointers.
As a side note, we can't pass `std::unique_ptr`to a stdpar algorithm as the copy constructor is deleted so we'll still need to pass a raw pointer using `std::unique_ptr::get`.

src/stdpar/plugin-PixelVertexFinding/gpuSortByPt2.h
fwyzard: > `nvcc` also support this syntax. 

it does ?

esseivaju: this is what Bryce Lelbach said in his [GTC talk](https://www.nvidia.com/en-us/on-demand/session/gtcspring21-s31358/) last year but I did not try it myself so I might be wrong on that

src/stdpar/CondFormats/PixelCPEFast.cc
```diff
@@ -14,57 +13,31 @@
 //-----------------------------------------------------------------------------
 //!  The constructor.
 //-----------------------------------------------------------------------------
-PixelCPEFast::PixelCPEFast(std::string const &path) {
+PixelCPEFast::PixelCPEFast(std::string const &path)
+    : m_params{std::make_shared<pixelCPEforGPU::ParamsOnGPU>()},
+      m_commonParams{std::make_shared<pixelCPEforGPU::CommonParams>()},
+      m_layerGeometry{std::make_shared<pixelCPEforGPU::LayerGeometry>()},
+      m_averageGeometry{std::make_shared<pixelCPEforGPU::AverageGeometry>()} {
```
makortel, line 20: Why `shared_ptr` and not `unique_ptr`?

src/stdpar/CondFormats/SiPixelGainCalibrationForHLTGPU.cc
```diff
@@ -10,28 +10,17 @@
...
-#ifndef CUDAUVM_DISABLE_ADVISE
-    cudaCheck(cudaMemAdvise(gainForHLT_, sizeof(SiPixelGainForHLTonGPU), cudaMemAdviseSetReadMostly, device));
-    cudaCheck(cudaMemAdvise(gainData_, gainData.size(), cudaMemAdviseSetReadMostly, device));
+                                                                 std::vector<DecodingStructure> const& gainData)
+    : gainForHLT_{std::make_unique<SiPixelGainForHLTonGPU>(gain)},
+//make_shared for array
+#if __GNUC__ >= 12
+      gainData_{std::make_shared<DecodingStructure[]>(gainData.size())}
+#else
+      gainData_{new DecodingStructure[gainData.size()]}
```
makortel, line 19: The `gainData` could probably be also `std::vector<DecodingStructure>`, right? That would help to avoid the `#if`. (otherwise, why `shared_ptr` instead of `unique_ptr?)

src/stdpar/plugin-PixelTriplets/gpuPixelDoublets.h
```diff
@@ -1,8 +1,6 @@
 #ifndef RecoLocalTracker_SiPixelRecHits_plugins_gpuPixelDoublets_h
 #define RecoLocalTracker_SiPixelRecHits_plugins_gpuPixelDoublets_h
 
-#include "gpuPixelDoubletsAlgos.h"
```
makortel, line 4: Can the file `gpuPixelDoubletsAlgos.h` be removed too?

makortel (commented):
@esseivaju Thanks for the update! I had a few comments/questions more, after those I think this starts to be good-enough for merging.

src/stdpar/CondFormats/PixelCPEFast.cc
esseivaju: I actually had a local commit which use `unique_ptr`. I was using a `shared_ptr` because data members were also taking a pointer to these structures.

esseivaju:
@makortel All the current comments have been resolved, however, as mentioned in the original comment, running the stdpar application still results in runtime memory access error: 

```
/bld4/home/esseivaj/devel/pixeltrack-standalone/src/stdpar/CUDACore/allocate_managed.cc, line 32:
cudaCheck(cudaFree(ptr));
cudaErrorIllegalAddress: an illegal memory access was encountered
```

Replacing the ```cudaMalloc*``` with heap allocation and letting ```nvc++``` fixes the issue, however, you'll then get similar errors when destroying ```cudaStream_t``` / ```cudaEvent_t``` which require more important changes in how the framework processes events. Should we already merge this pull request and fix the issue in a subsequent one or do you want to keep it open until runtime issues are fixed?

PR #368, opened 2022-08-06
[stdpar] Update explicit memory management through CUDA API to smart pointers

esseivaju:
This pull request aims to update ```CUDACore``` and ```CUDADataFormats``` libraries and plugins to remove all explicit memory management through the CUDA API and replace it with the use of ```std::unique_ptr``` /  ```std::shared_ptr``` which will be instrumented by ```nvc++``` during compilation to use unified memory.

This will include changes such as:

- Removal of classes providing portability between Host and Device which are no longer needed
- Removal of most CUDA runtime API calls
- Plugins update to be compatible with the updated ```CUDACore``` and ```CUDADataFormats``` libraries
- Renaming of files, classes, libraries to reflect the changes

Most notably, plugins will still use CUDA kernels but these changes will help in porting kernels to stdpar algorithms as most the framework will already be standard c++.


esseivaju:
@makortel The pull request is starting to become quite large even though most changes are deletion so we should probably review it now. 

As mentioned in the first comment, the core changes are to memory management; allocate memory on the heap through ```std::unique_ptr``` and let ```nvc++``` instrument the unified memory usage. This led a lot of extra changes as a lot of classes were designed to be generic over the allocation method / pointer type, which is not needed in the stdpar version

src/stdpar/plugin-SiPixelRawToDigi/SiPixelDigisSoAFromCUDA.cc
```diff
@@ -50,21 +42,12 @@ void SiPixelDigisSoAFromCUDA::acquire(const edm::Event& iEvent,
...
-  rawIdArr_ = gpuDigis.rawIdArrToHostAsync(ctx.stream());
-  adc_ = gpuDigis.adcToHostAsync(ctx.stream());
-  clus_ = gpuDigis.clusToHostAsync(ctx.stream());
-#else
-  gpuDigis.pdigiPrefetchAsync(cudaCpuDeviceId, ctx.stream());
-  gpuDigis.rawIdArrPrefetchAsync(cudaCpuDeviceId, ctx.stream());
-  gpuDigis.adcPrefetchAsync(cudaCpuDeviceId, ctx.stream());
-  gpuDigis.clusPrefetchAsync(cudaCpuDeviceId, ctx.stream());
+  // This memory is already owned by a unique pointer, therefore class members cannot be unique_ptr themselves
+  // *this must have a shorter lifetime than gpuDigis object retrieved from the context or we risk using dangling pointers
```
makortel, line 46: Just to clarify, the objects retrieved from `edm::Event` are alive until everything has been processed for a given event. In absence of asynchronous execution the use of `edm::EDProducerExternalWork` and the split to `acquire()` and `produce()` functions is not necessary (i.e. all this could be simplified in some subsequent PR).

makortel (commented):
Thanks, looks good! My only question that _might_ be (or not) worth of considering further in this PR is the added initialization of data used first time in device as a target of a write operation.

src/stdpar/DataFormats/PixelErrors.h
```diff
@@ -10,10 +10,10 @@
...
 
 struct PixelErrorCompact {
-  uint32_t rawId;
-  uint32_t word;
-  uint8_t errorType;
-  uint8_t fedId;
+  uint32_t rawId {0};
+  uint32_t word {0};
+  uint8_t errorType {0};
+  uint8_t fedId {0};
```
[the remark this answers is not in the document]
esseivaju: It was mostly to play safe as I did not check how this particular struct was used. If it's written to by the device before being read and you noticed important cost I'll revert it back.

src/stdpar/plugin-BeamSpotProducer/BeamSpotToCUDA.cc
```diff
@@ -23,38 +17,18 @@ class BeamSpotToCUDA : public edm::EDProducer {
   void produce(edm::Event& iEvent, const edm::EventSetup& iSetup) override;
 
 private:
-  edm::EDPutTokenT<cms::cuda::Product<BeamSpotCUDA>> bsPutToken_;
-
-#ifdef CUDAUVM_DISABLE_MANAGED_BEAMSPOT
-  cms::cuda::host::noncached::unique_ptr<BeamSpotPOD> bsHost;
-#endif
+  edm::EDPutTokenT<cms::cuda::Product<BeamSpot>> bsPutToken_;
```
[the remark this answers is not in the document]
esseivaju: Thanks for the info. I was planning on removing ```cudaStream_t``` and ```cudaEvent_t``` usage  in the next PR so it'll probably also involve those two.

src/stdpar/plugin-SiPixelRawToDigi/SiPixelDigisSoAFromCUDA.cc
esseivaju: In that case, the class ```edm::EDProducerExternalWork``` would no longer be necessary since its purpose was to provide ```acquire()```?  Parallel algorithms seem to be a similar abstraction to what ```edm::EDProducerExternalWork``` does; offloading the work "somewhere" to parallelize it.

src/stdpar/DataFormats/PixelErrors.h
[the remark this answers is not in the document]
makortel: For this particular struct the initialization likely has negligible impact. I was more commenting about the general approach. Could you undo the added initializations elsewhere as well (since you already started that route :))?

src/stdpar/plugin-SiPixelRawToDigi/SiPixelDigisSoAFromCUDA.cc
makortel: > In that case, the class `edm::EDProducerExternalWork` would no longer be necessary since its purpose was to provide `acquire()`?

Correct (and I'd leave the cleanup in a later PR).

> Parallel algorithms seem to be a similar abstraction to what `edm::EDProducerExternalWork` does; offloading the work "somewhere" to parallelize it.

I'd say the abstractions are not quite the same. The part of "offload work somewhere" is indeed similar. But the key point of `edm::EDProducerExternalWork` is that while the offloaded work is being run (i.e. the work has been launched in `acquire()`) the framework (really TBB) can run other CPU tasks on the same CPU worker thread that run `acquire()`. The `produce()` is called after the offloaded work has finished, but not necessary by the same CPU worker thread.

src/stdpar/plugin-PixelTriplets/CAHitNtupletGeneratorOnGPU.cc
```diff
@@ -81,45 +81,24 @@ CAHitNtupletGeneratorOnGPU::CAHitNtupletGeneratorOnGPU(edm::ProductRegistry& reg
...
-  if (m_params.onGPU_) {
-#ifdef CUDAUVM_MANAGED_TEMPORARY
-    cudaCheck(cudaMallocManaged(&m_counters, sizeof(Counters)));
-#else
-    cudaCheck(cudaMalloc(&m_counters, sizeof(Counters)));
-#endif
-    cudaCheck(cudaMemset(m_counters, 0, sizeof(Counters)));
-  } else {
-    m_counters = new Counters();
-    memset(m_counters, 0, sizeof(Counters));
```
esseivaju, line 94: This memset has been replaced by in-class initializer of ```Counters``` which should both have similar cost.

src/stdpar/DataFormats/PixelErrors.h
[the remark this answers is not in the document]
esseivaju: Done except for Counters where it was memset to 0 before use.

src/stdpar/plugin-SiPixelRawToDigi/SiPixelDigisSoAFromCUDA.cc
esseivaju: Yes we lose the asynchronous offloading of work until the c++ standard library has better async support, most likely with c++26 and senders / receivers.

makortel:
@esseivaju Just to confirm, from your side this PR is ready to be merged, right?

esseivaju:
@makortel Yes it should be ready to be merged

PR #372, opened 2022-08-17
[stdpar] Update stdpar to remove Product, ScopedContext, device abstraction

esseivaju:
As discussed in #368 this PR contains an updated to cudart api utilization to get closer to standard c++ and no cuda dependencies. 

## Changes

- No longer use the cudaStream and cudaEvent APIs.
- Removed multiple classes in ```CUDACore``` which are no longer necessary
- Updated Plugins and tests to be compatible with the previously mentioned changes. 
- Refactored multiple producers to inherit from ```EDProducer``` instead of ```EDProducerExternalWork```

# Current issues to be fixed

- Multiple tests still manage memory through ```cudaMalloc``` / ```cudaMemcpy```, this should be updated to use ```std::unique_ptr```
- There are still runtime errors. I am currently working on figuring out exactly what's the issue. I've noticed that some kernels seem to produce incorrect results (e.g. the```SiPixelRawToClusterCUDA``` module). I updated the ```gpuClustering_t``` test to work with ```std::unique_ptr``` and it fails with the following output:

```
[esseivaj@zeus pixeltrack-standalone]$ test/stdpar/gpuClustering_t 
created 175031 digis in 8960 clusters
CUDA countModules kernel launch with 2000 blocks of 256 threads
CUDA findModules kernel launch with 2000 blocks of 256 threads
before charge cut found 0 clusters
ERROR!!!!! wrong number of cluster found
found 897 Modules active
gpuClustering_t: /bld4/home/esseivaj/devel/pixeltrack-standalone/src/stdpar/test/gpuClustering_t.h:284: int main(): Assertion `h_clus[i] < int(nclus[h_id[i]])' failed.
Aborted (core dumped)
```


src/stdpar/plugin-PixelTrackFitting/PixelTrackSoAFromCUDA.cc
```diff
@@ -1,42 +1,23 @@
...
-#include "CUDACore/Product.h"
 #include "CUDADataFormats/PixelTrack.h"
 #include "Framework/EventSetup.h"
 #include "Framework/Event.h"
 #include "Framework/PluginFactory.h"
 #include "Framework/EDProducer.h"
-#include "CUDACore/ScopedContext.h"
 
-class PixelTrackSoAFromCUDA : public edm::EDProducerExternalWork {
+class PixelTrackSoAFromCUDA : public edm::EDProducer {
```
makortel, line 7: In principle this module is now useless (e.g. the simple mock framework can't handle multiple modules producing the same type of product). In `cudauvm` it played to role of issuing prefetches to the host memory, but in `stdpar` even that does not make much sense. I suppose for the "transfer to host" tests we need to measure the time of something actually accessing all the memory (e.g. `HistoValidator`).

src/stdpar/plugin-PixelVertexFinding/PixelVertexSoAFromCUDA.cc
```diff
@@ -1,48 +1,29 @@
...
 #include "CUDADataFormats/ZVertex.h"
 #include "Framework/EventSetup.h"
 #include "Framework/Event.h"
 #include "Framework/PluginFactory.h"
 #include "Framework/EDProducer.h"
 #include "Framework/RunningAverage.h"
-#include "CUDACore/ScopedContext.h"
 
-class PixelVertexSoAFromCUDA : public edm::EDProducerExternalWork {
+class PixelVertexSoAFromCUDA : public edm::EDProducer {
```
makortel, line 8: In principle this module is also useless now (well, actually earlier but I realized it only now).

src/stdpar/plugin-SiPixelRawToDigi/SiPixelDigisSoAFromCUDA.cc
```diff
@@ -1,54 +1,24 @@
...
 #include "CUDADataFormats/SiPixelDigis.h"
 #include "DataFormats/SiPixelDigisSoA.h"
 #include "Framework/EventSetup.h"
 #include "Framework/Event.h"
 #include "Framework/EDProducer.h"
 #include "Framework/PluginFactory.h"
-#include "CUDACore/ScopedContext.h"
 
-class SiPixelDigisSoAFromCUDA : public edm::EDProducerExternalWork {
+class SiPixelDigisSoAFromCUDA : public edm::EDProducer {
```
esseivaju, line 8: I would assume that this module is also not needed as it's never user

esseivaju:
Changes listed in the original comment have been implemented. The tests below are still failing: 

* ```HistoContainer_t```
* ```OneToManyAssoc_t```
* ```prefixScan_t```
* ```radixSort_t```
* Tests depending on a plugin

Fixing those will require updating many Kernels so it should probably be done in a future PR as this one is getting fairly large. Some of these tests (e.g. ```prefixScan_t``` / ```prefixScan.h```) are still pretty much identical to the original ```cudauvm``` version, the only change being update to ```#define``` directives to compile the correct code with ```nvc++``` so it's suggesting some issues with ```nvc++```

esseivaju:
Yes we can merge now I will fix the remaining issues with HistoContainer, prefixScan and radixSort in the next pull request.

PR #375, opened 2022-08-31
[stdpar] Formatting and Readme update

esseivaju:
## Changes
* Format stdpar program using clang-format
* Updated Readme description of stdpar
* Updated .clang-format file

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -4,7 +4,7 @@ BasedOnStyle:  Google
 ColumnLimit:     120
 #didn't we want to change this?
 NamespaceIndentation: All
-SortIncludes:    false
+SortIncludes:    Never
 IndentWidth:     2
 AccessModifierOffset: -2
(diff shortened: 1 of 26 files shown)


.clang-format
```diff
@@ -16,4 +16,3 @@ BinPackParameters: false
 AlwaysBreakTemplateDeclarations: Yes
 ReflowComments: false
 BinPackArguments: false
-BinPackParameters: false
```
makortel, line 19: Are these updates related to using newer `clang`?

esseivaju: The `BinPackParameters` was just a duplicate entry. `SortIncludes` is a change in `clang-13`, since the Makefile specifically uses `clang-format-10` I can revert that change.

makortel: Yeah, maybe we could update the clang version used for formatting at some point.

PR #376, opened 2022-09-02
[stdpar] Porting radixSort and sortByPt2 Kernels to stdpar 

esseivaju:
## Changes 

* use ```std::sort``` instead of the CUDA radix sort implementation
*  Updated sortByPt2 device function which was calling the radix sort implementation. Now a host function making ```stdpar``` algorithms calls.

## Issues

946a5eebe3c4829b18fdff6577d766d393105418 introduces a link error with ```nvc++``` 
```
nvlink error   : Undefined reference to '__atomic_compare_exchange' in '/bld4/home/esseivaj/devel/pixeltrack-standalone/obj/stdpar/plugin-PixelVertexFinding/gpuVertexFinder.cu.o'
nvlink error   : Undefined reference to '__atomic_load' in '/bld4/home/esseivaj/devel/pixeltrack-standalone/obj/stdpar/plugin-PixelVertexFinding/gpuVertexFinder.cu.o'
```


esseivaju:
The linker error occurs because `nvc++` doesn't yet support `std::atomic_ref<T>` when `T` is a floating_point type on device code. I added a workaround, dispatching `atomicAdd()` of floating_point types to cuda atomic or `std::atomic_ref` depending on the compiler / execution space.

PR #380, opened 2022-10-06
[cuda] Implementing an autotuning interface and script

asubah:
This PR introduces an autotuning interface and a python script to autotune the CUDA kernel launch configurations.

Refer to the [`README.md`](https://github.com/asubah/pixeltrack-standalone/blob/autotuning/autotuning/README.md) for instructions on how to use the script.

The interface is simple. It just reads the block size from a file in [`autotuning/kernel_configs`](https://github.com/asubah/pixeltrack-standalone/tree/autotuning/autotuning/kernel_configs).

Tested on a node with Intel Xeon Silver 4214R and a single T4.
Baseline configuration: [`autotuning/tunables-baseline.csv`](https://github.com/asubah/pixeltrack-standalone/blob/autotuning/autotuning/tunables-baseline.csv).
Average Throughput: **1,195.66 ± 1.96 events/s.**

Best config found by autotuning:  [`autotuning/tunables.csv`](https://github.com/asubah/pixeltrack-standalone/blob/autotuning/autotuning/tunables.csv).
Average Throughput: **1,269.46 ± 3.52 events/s.**

The search space is too big, hence, I selected a subset of the kernels that has the highest runtime and randomly autotuned them. Then fixed the configurations on the best found configuration, and tuned another set of kernels until all kernels are visited.

I appreciate your comments and suggestions regarding the interface or the script.


src/cuda/plugin-PixelTriplets/BrokenLineFitOnGPU.cu
```diff
@@ -18,13 +15,31 @@ void HelixFitOnGPU::launchBrokenLineKernels(HitsView const *hv,
   auto fast_fit_resultsGPU_ = cms::cuda::make_device_unique<double[]>(
       maxNumberOfConcurrentFits_ * sizeof(Rfit::Vector4d) / sizeof(double), stream);
 
+  cms::cuda::ExecutionConfiguration exec;
```
makortel, line 18: Could all these be done in the `cudadev` program instead (or "fork" a separate program, e.g. `cudaautotune`)? I'd like to keep the reference `cuda` program as close to the original as possible.

asubah: I forked it to cudaautotune, and changed the tuning script directory to be under `src/cudaautotune/autotuining`.

Makefile
```diff
@@ -66,7 +66,7 @@ export CUDA_DEPS := $(CUDA_LIBDIR)/libcudart.so
 export CUDA_ARCH := 35 50 60 70
 export CUDA_CXXFLAGS := -I$(CUDA_BASE)/include
 export CUDA_TEST_CXXFLAGS := -DGPU_DEBUG
-export CUDA_LDFLAGS := -L$(CUDA_LIBDIR) -lcudart -lcudadevrt
+export CUDA_LDFLAGS := -L$(CUDA_LIBDIR) -lcudart -lcudadevrt -lcudaautotunert
```
makortel, line 69: Could this library be added only in `src/cudaautotune/Makefile`?

src/cudaautotune/CUDACore/ExecutionConfiguration.h
```diff
@@ -0,0 +1,42 @@
+#ifndef HeterogeneousCore_CUDAUtilities_interface_ExecutionConfiguration_h
+#define HeterogeneousCore_CUDAUtilities_interface_ExecutionConfiguration_h
+
+#include <fstream>
+
+namespace cms {
+  namespace cuda {
+
+    class ExecutionConfiguration {
```
makortel, line 9: Currently this class appears to be used only as a "namespace", because the objects have no state.

On the other hand, if I understood correctly, the launch parameters are read from the file for each kernel for each event. Would it be feasible to read each file only once in some way?

src/cudaautotune/CUDACore/ExecutionConfiguration.h
```diff
@@ -0,0 +1,42 @@
...
+        size_t blockSize;
+
+        std::fstream file;
+        file.open(CONFIG_PATH + filename, std::ios::in);
+        if (file.is_open())
+        {
+          file >> blockSize;
+          file.close();
+        } else {
+            std::cout << "Error in opening file " + filename + "\n";
```
makortel, line 31: Does it make sense to continue the program if a file can not be opened? (currently the function returns an indeterminate value)

src/cudaautotune/CUDACore/HistoContainer.h
```diff
@@ -0,0 +1,325 @@
...
+    inline __attribute__((always_inline)) void launchFinalize(Histo *__restrict__ h,
+                                                              cudaStream_t stream
+#ifndef __CUDACC__
+                                                              = cudaStreamDefault
+#endif
+    ) {
+#ifdef __CUDACC__
+      uint32_t *poff = (uint32_t *)((char *)(h) + offsetof(Histo, off));
+      int32_t *ppsws = (int32_t *)((char *)(h) + offsetof(Histo, psws));
+      cms::cuda::ExecutionConfiguration exec;
```
makortel, line 81: This doesn't seem to be used.

src/cudaautotune/Makefile
```diff
@@ -0,0 +1,160 @@
+TARGET_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+TARGET_NAME := $(notdir $(TARGET_DIR))
+TARGET := $(BASE_DIR)/$(TARGET_NAME)
+include Makefile.deps
+EXTERNAL_DEPENDS := $(cuda_EXTERNAL_DEPENDS)
```
makortel, line 5: This should be 
```suggestion
EXTERNAL_DEPENDS := $(cudaautotune_EXTERNAL_DEPENDS)
```

src/cudaautotune/Makefile.deps
```diff
@@ -0,0 +1,12 @@
+cuda_EXTERNAL_DEPENDS := TBB CUDA EIGEN BOOST BACKTRACE
```
makortel, line 1: This should be 
```suggestion
cudaautotune_EXTERNAL_DEPENDS := TBB CUDA EIGEN BOOST BACKTRACE
```

src/cudaautotune/autotuning/tuner.py
```diff
@@ -0,0 +1,149 @@
+import argparse
```
makortel, line 1: How about adding
```suggestion
#!/usr/bin/env python3
import argparse
```
so that the script could be run "directly" (./src/cudaautotune/autotuning/tuner.py)?

src/cudaautotune/autotuning/tuner.py
```diff
@@ -0,0 +1,149 @@
...
+import subprocess
+import itertools
+import random
+import datetime
+
+parser = argparse.ArgumentParser(description='Autotuner script.')
+parser.add_argument('-p', '--process', type=pathlib.Path, nargs=1, required=True,
+        help='path to the program to be autotuned')
+parser.add_argument('-c', '--configurations', type=pathlib.Path, nargs=1,
+        default=[pathlib.Path('src/cudaautotune/autotuning/kernel_configs')], help='path to save the configurations for the tunable process to read them. Default = autotuning/kernel_configs/')
```
makortel, line 12: Why the default value is in a list? (same for the following two arguments too)

src/cudaautotune/autotuning/tuner.py
```diff
@@ -0,0 +1,149 @@
...
+        print('Validation result:', validation)
+        
+    # Benchmarking
+    time = "NaN"
+    throughput = "NaN"
+    cpu_efficiency = "NaN"
+    status = ""
+
+    cpu_threads = config[tunables["cpu_threads"]]
+    gpu_streams = cpu_threads + config[tunables["gpu_streams"]]
```
makortel, line 119: It is not clear to me why the number of concurrent events (`--numberOfStreams` below) is set to "number of CPU threads" + something named "GPU streams". Could you clarify what is the intended behavior here wrt. CPU threads and concurrent events?

PR #382, opened 2022-11-01
[openmp] Add OpenMP compilers and flags

markdewing:
Add OpenMP compilers to the Makefile.  Change the compiler name and flags in src/openmp/Makefile.


Makefile
```diff
@@ -309,6 +309,36 @@ export SYCL_CUDA_FLAGS  := --cuda-path=$(CUDA_BASE) -Wno-unknown-cuda-version
...
+ifeq ($(OPENMP_COMPILER), LLVM)
+  export OPENMP_CXX := clang++
+  # CUDA
+  export OPENMP_TARGETS := -fopenmp-targets=nvptx64-nvidia-cuda --cuda-path=$(CUDA_BASE)
+  # HIP
+  #export OPENMP_TARGETS := -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdhsa-amd-amdhsa  -march=gfx900
+  export OPENMP_CXXFLAGS := -fopenmp-version=51 -foffload-lto -O2 -fPIC -std=c++17 -fopenmp $(OPENMP_TARGETS) -fopenmp-offload-mandatory $(OPENMP_EIGEN_CXXFLAGS) --gcc-toolchain=$(GCC_TOOLCHAIN)
+  export OPENMP_LDFLAGS := $(LDFLAGS) -foffload-lto -O2 --gcc-toolchain=$(GCC_TOOLCHAIN)
+else ifeq ($(OPENMP_COMPILER), AMD)
+  echo "aompcc not supported"
```
makortel, line 330: How about
```suggestion
  $(error aompcc not supported)
```
? (like done here
https://github.com/cms-patatrack/pixeltrack-standalone/blob/03b27ca1239077c9d62da04e2075a52192df7f23/Makefile#L14-L17
)

makortel:
@markdewing I took the liberty of pushing directly a commit that will allow the `all` target to not include `openmp` by default. I'm not completely happy with the way I did it, but I think it is reasonable compromise that allows to move on. What do you think?

markdewing:
It looks reasonable to me.

PR #385, opened 2022-11-10
Upgrade oneTBB to 2021.7.0

felicepantaleo:
Upgrading the oneTBB external to `2021.7.0`.
Compiles, runs and passes validation.


makortel:
Just curious, is there any specific reason to motivate the update? (I am going to merge this after some tests)

fwyzard:
Marco has issues building on Ubuntu 21.10 with the previous release, while
this one worked.

Luca and Aurora were also asking to update, but I don't recall the exact
reason.

PR #387, opened 2022-12-09
[SYCL] Add SYCL implementation

AuroraPerego:
Implementation for the SYCL/oneAPI backend based on `cuda` with some inspiration taken from `alpaka`.

The SYCL implementation can be compiled with `dpcpp` or with `clang++`. The latter is the default since it supports the CUDA backend. To compile with `dpcpp` set `USE_SYCL_ONEAPI=1`. In this case TBB from oneAPI is used and it's not cloned as an external.
```bash
make environment
source env.sh
make -j `nproc` sycl
```

The device(s) can be chosen at runtime with `--device`:
- `cpu`: the CPU can be selected, but it hangs after a couple of events (currently under investigation with Intel because there seems to be a conflict between TBB of the framework and of the runtime openCL)
- `gpu`: targets all the GPUs
- `level_zero`: for Intel GPUs with the level_zero backend
- `cuda`: for NVIDIA GPUs (at the moment that only one that is consistent with the result)
- `hip`: for AMD GPUs (doesn't compile, there is a bug)

Some changes (e.g. different kernels on CPU and GPU, shared variables...) are due to bugs in SYCL and can be reverted when those will be solved (marked with `SYCL_BUG_`).

**NOTE**: the line 155 of `src/sycl/plugin-PixelVertexFinding/gpuClusterTracksIterative.h` has been commented due to a bug in the compiler, so `gpuClusterTracksIterative` cannot be used
 
Regarding performance, a comparison on a NVIDIA GPU with native CUDA has been carried out, showing that there is still a lot of work to do to reach the performance of native CUDA.

More details on what has been changed and on the performance are available [here](https://codimd.web.cern.ch/s/F6JEoLNGU)


Makefile
```diff
@@ -114,6 +114,79 @@ export ROCM_LDFLAGS := -L$(ROCM_LIBDIR) -lamdhip64
...
+AOT_HIP_FLAGS     := -fsycl-targets=amdgcn-amd-amdhsa -Xsycl-target-backend --offload-arch=gfx900 --rocm-path=$(ROCM_BASE) -Wno-linker-warnings 
+
+# INTEL flags: compile AOT for all the CPUs and for the GPU on olice-05
+# CUDA flags : compile AOT for for NVIDIA GPUs
+# HIP flags  : compile AOT for architectures with ID gfx900 (e.g. the Radeon PRO WX 9100)
+
+# -Wno-linker-warnings will not be needed be needed anymore with https://github.com/intel/llvm/pull/7245
+
+ifdef USE_SYCL_ONEAPI
+ONEAPI_BASE       := /cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2022
```
makortel, line 134: Could the `if` blocks be indented? They get complicated enough to be difficult to follow without.

Makefile
```diff
@@ -114,6 +114,79 @@ export ROCM_LDFLAGS := -L$(ROCM_LIBDIR) -lamdhip64
...
+ONEAPI_VERSION    := latest
+TBB_BASE          := $(ONEAPI_BASE)/tbb/latest
+TBB_LIBDIR        := $(TBB_BASE)/lib/intel64/gcc4.8
+ifneq ($(wildcard $(ONEAPI_BASE)),)
+ONEAPI_ENV        := $(ONEAPI_BASE)/setvars.sh # --config="/eos/user/a/aperego/dev/pixeltrack-standalone/config.txt"
+                                               # the config.txt file can be used to source only specific tools 
+                                               # or a specific version of a tool of the oneAPI package
+SYCL_BASE         := $(ONEAPI_BASE)/compiler/$(ONEAPI_VERSION)/linux
+USER_SYCLFLAGS    := -fp-model=precise -fimf-arch-consistency=true -no-fma
```
makortel, line 143: Is there are reason to use `USER_SYCLFLAGS` specifically? The `USER_*` variables are intended to be overridden from command line, so these would be lost. I'd suggest to use a different name for the variable (or move directly as part of `SYCL_CXXFLAGS`)

Makefile
```diff
@@ -114,6 +114,79 @@ export ROCM_LDFLAGS := -L$(ROCM_LIBDIR) -lamdhip64
...
+endif
+endif
+# check if libraries are under lib or lib64
+ifdef SYCL_BASE
+ifneq ($(wildcard $(SYCL_BASE)/lib/libsycl.so),)
+SYCL_LIBDIR := $(SYCL_BASE)/lib
+else ifneq ($(wildcard $(SYCL_BASE)/lib64/libsycl.so),)
+SYCL_LIBDIR := $(SYCL_BASE)/lib64
+else
+SYCL_BASE :=
```
makortel, line 186: Undefining `SYCL_BASE` basically means disabling SYCL. Should the `TBB_BASE` be explicitly undefined as well in this case?

Makefile
```diff
@@ -274,40 +355,11 @@ ifdef KOKKOS_HOST_PARALLEL
...
-USER_SYCLFLAGS :=
-ifdef SYCL_BASE
-export SYCL_CXX      := $(SYCL_BASE)/bin/dpcpp
-export SYCL_CXXFLAGS := -fsycl $(DPCT_CXXFLAGS) $(filter-out $(SYCL_UNSUPPORTED_CXXFLAGS),$(CXXFLAGS)) $(USER_SYCLFLAGS)
 ifdef CUDA_BASE
 export SYCL_CUDA_PLUGIN := $(wildcard $(SYCL_LIBDIR)/libpi_cuda.so)
 export SYCL_CUDA_FLAGS  := --cuda-path=$(CUDA_BASE) -Wno-unknown-cuda-version
 endif
-endif
+# endif
```
makortel, line 362: Could this commented line be removed?

Makefile
```diff
@@ -570,24 +625,24 @@ $(TBB_LIB):
 	mkdir -p $(TBB_TMP)
 	mkdir -p $(TBB_TMP_SRC)
 	mkdir -p $(TBB_TMP_BUILD)
-	git clone --branch v2021.4.0 https://github.com/oneapi-src/oneTBB.git $(TBB_TMP_SRC)
+	git clone --branch v2021.7.0 https://github.com/oneapi-src/oneTBB.git $(TBB_TMP_SRC)
```
makortel, line 628: Reminds me that I should resume testing #385,

Makefile
```diff
@@ -570,24 +625,24 @@ $(TBB_LIB):
...
 $(EIGEN_BASE):
-	# from Eigen master branch as of 2021.08.18
-	git clone -b cms/master/82dd3710dac619448f50331c1d6a35da673f764a https://github.com/cms-externals/eigen-git-mirror.git $@
-	# include all Patatrack updates
-	cd $@ && git reset --hard 6294f3471cc18068079ec6af8ceccebe34b40021
-
+	# from Eigen master branch on gitlab as of 2022.08.09
+	git clone https://gitlab.com/AuroraPerego/eigen.git $@
+	# include all SYCL updates
+	cd $@ && git reset --hard eb029b6bc9c33172f0b77524d226d2bca52945a2
```
makortel, line 645: @fwyzard Would it make sense to clone the `eigen` under `cms-patatrack`? I'd also like to test its impact for other programs separately.

src/sycl/CondFormats/PixelCPEFast.cc
```diff
@@ -0,0 +1,62 @@
...
+                  &this->m_averageGeometry,
+                  sizeof(pixelCPEforGPU::AverageGeometry));
+    stream.memcpy((void *)data.h_paramsOnGPU.m_layerGeometry.get(),
+                  &this->m_layerGeometry,
+                  sizeof(pixelCPEforGPU::LayerGeometry));
+    stream
+        .memcpy((void *)data.h_paramsOnGPU.m_detParams.get(),
+                this->m_detParamsGPU.data(),
+                this->m_detParamsGPU.size() * sizeof(pixelCPEforGPU::DetParams))
+        .wait();
```
makortel, line 57: Why is synchronization needed here? (I'm not repeating the question for the other ESProducts that follow similar pattern)

src/sycl/CondFormats/PixelCPEFast.h
```diff
@@ -0,0 +1,38 @@
...
+  pixelCPEforGPU::CommonParams m_commonParamsGPU;
+  pixelCPEforGPU::LayerGeometry m_layerGeometry;
+  pixelCPEforGPU::AverageGeometry m_averageGeometry;
+
+  struct GPUData {
+    ~GPUData();
+    // not needed if not used on CPU...
+    pixelCPEforGPU::ParamsOnGPU h_paramsOnGPU;
+    cms::sycltools::device::unique_ptr<pixelCPEforGPU::ParamsOnGPU> d_paramsOnGPU =
+        nullptr;  // copy of the above on the Device
```
makortel, line 31:
```suggestion
    // copy of the above on the Device
    cms::sycltools::device::unique_ptr<pixelCPEforGPU::ParamsOnGPU> d_paramsOnGPU = nullptr;
```
_might_ format better

src/sycl/CondFormats/SiPixelFedCablingMapGPUWrapper.h
```diff
@@ -0,0 +1,51 @@
...
+  bool hasQuality_;
+
+  SiPixelFedCablingMapGPU *cablingMapHost_ = nullptr;  // pointer to struct in CPU
+
+  struct GPUData {
+    GPUData() = default;
+    ~GPUData() {}
+
+    std::unique_ptr<SiPixelFedCablingMapGPU, cms::sycltools::device::impl::DeviceDeleter>
+        cablingMapDevice;  // pointer to struct in GPU
```
makortel, line 37: Would
```suggestion
    cms::sycltools::device::unique_ptr<SiPixelFedCablingMapGPU> cablingMapDevice;  // pointer to struct in GPU
```
work here? (I haven't gotten yet to the definition of `cms::sycltools::device::unique_ptr`)

src/sycl/DataFormats/approx_atan2.h
```diff
@@ -0,0 +1,290 @@
...
+  print("}") >> filename;
+end;
+*/
+
+#include <cstdint>
+#include <cmath>
+#include <limits>
+#include <algorithm>
+
+using sycl::abs;
```
makortel, line 35: Does this `using` declaration rely on a file including this header to include also `CL/sycl.hpp`? Or do the SYCL compilers provide some magic?

src/sycl/Framework/RunningAverage.h
```diff
@@ -0,0 +1,53 @@
...
+    explicit RunningAverage(unsigned int k = 4) : m_mean(N * k), m_curr(0) {
+      for (auto& i : m_buffer)
+        i = k;
+    }
+
+    int mean() const { return m_mean / N; }
+
+    int upper() const {
+      auto lm = mean();
+      return lm += (sycl::abs(m_buffer[0] - lm) + sycl::abs(m_buffer[N / 2] - lm));
```
makortel, line 32: With explicit use of `sycl::abs()` I'd suggest to add `#include <CL/sycl.hpp>` to make the header dependence more explicit.

makortel (commented):
Got up to `Makefile.deps`, more will follow.

Makefile
AuroraPerego: Not really, I can move them

Makefile
AuroraPerego: Should I revert back to 2021.4 and wait for the PR to add the change?

Makefile
AuroraPerego: When I started, the version of `eigen` in the Makefile wasn't working with SYCL (neither the new version cloned from gitlab).  I took the newest version and put on top of that the changes needed for SYCL (https://gitlab.com/libeigen/eigen/-/merge_requests/1132).
Before this PR I tested serial, native CUDA and  alpaka (serial, CUDA, HIP) with both this version of Eigen and the original one in the Makefile. I've only looked at the throughput and it doesn't change.

src/sycl/CondFormats/SiPixelFedCablingMapGPUWrapper.h
AuroraPerego: yes, I'll change it. In `cms::sycltools::device`:
```cpp
template <typename T>
      using unique_ptr = std::unique_ptr<T, impl::DeviceDeleter>;
```

src/sycl/DataFormats/approx_atan2.h
AuroraPerego: Yes to the first question, otherwise it wouldn't compile (but I think I've just been lucky and that's better to include `CL/sycl.hpp` here). 
The answer to the second question could be also a yes for what I've seen :)

src/sycl/CondFormats/PixelCPEFast.cc
AuroraPerego: Here and for the other ESProducts is not needed, changed it.

Makefile
AuroraPerego: Regarding this, I did some plots to compare the `eigen` under `cms-patatrack` and this one. 
Eigen_GPU
Eigen_CPU
- new = `eigen` under `gitlab.com/AuroraPerego`
- old  = `eigen` under `cms-patatrack`
They show no loss of performance with the new version

makortel:
At this stage probably best to just merge as it is

PR #389, opened 2022-12-13
`std::par` implementation of `plugin-PixelVertexFinding`

esseivaju:
Port of `plugin-PixelVertexFinding` to `std::par`.

The original CUDA threading model was to process one module per thread block for most kernels. Parallel algorithms don't currently allow to express this model; as single execution of a `std::for_each` lambda will map to one thread so we now have one thread per module. 

There is still an issue when compiling `gpuVertexFinder::loadTracks`, `nvc++` hangs and never finishes


esseivaju:
Hanging of `nvc++`fixed by 564f669

PR #395, opened 2023-05-03
Update Eigen to version that works with CUDA, ROCm, and SYCL

makortel:
I made this by taking the branch by @AuroraPerego in https://github.com/cms-patatrack/pixeltrack-standalone/pull/387 and rebasing the earlier patches on top of that branch. I checked that it compiles with gcc, CUDA, ROCm, and SYCL with an intel-llvm build with cuda backend enabled.


makortel:
@AuroraPerego Could you check that this version of Eigen compiles and runs in your SYCL setup?

Makefile
```diff
@@ -655,9 +655,9 @@ external_eigen: $(EIGEN_BASE)
 
 $(EIGEN_BASE):
 	# from Eigen master branch on gitlab as of 2022.08.09
-	git clone https://gitlab.com/AuroraPerego/eigen.git $@
-	# include all SYCL updates
-	cd $@ && git reset --hard eb029b6bc9c33172f0b77524d226d2bca52945a2
+	git clone https://github.com/makortel/eigen.git $@
```
makortel, line 658: @fwyzard I wonder if we should setup eigen repository under cms-patatrack organization

fwyzard: Sure, we can do that.
Eventually we should push the changes to the CMSSW version...

fwyzard: (... or upstream ...)

AuroraPerego:
> @AuroraPerego Could you check that this version of Eigen compiles and runs in your SYCL setup?

I checked on CPU, NVIDIA GPU and Intel GPU and everything is fine (compiles and passes validation)

makortel:
I only updated the repository where the eigen is pulled from.

AuroraPerego:
FYI they have (finally!) added USM support for SYCL in eigen ([PR](https://gitlab.com/libeigen/eigen/-/merge_requests/1285)), so with just one additional change ([this](https://gitlab.com/libeigen/eigen/-/merge_requests/1326)) it will be possible to use, at least for SYCL, the eigen master branch.

PR #398, opened 2023-08-12
Warm up before measuring the throughput (old)

fwyzard:


fwyzard:
@makortel what do you think ?

I've considered a couple different approaches, mostly about where to put the `Timestamp` objects, but this was the easier to implement.

fwyzard:
Closed in favour of the approach used in #401 .

PR #402, opened 2023-08-22
Make alpaka platforms full objects

AuroraPerego:
Since [41e9956](https://github.com/alpaka-group/alpaka/commit/41e99568f1aa24a8b40b84f09b9db78dfdf98816) in Alpaka the `platform`s are actual objects and a couple of [commits](https://github.com/alpaka-group/alpaka/commit/8cf861bd6a3f8c511d21ae896a6fe13211944c4c) later they have been renamed from `Pltf` to `Platform`.
This PR propagates these changes to `alpaka` and `alpakatest`.

It also replaces `cms::alpakatools::enumerate()` with `alpaka::getDevs()` (thanks @fwyzard).

### DIFF
--- a/src/alpaka/AlpakaCore/alpakaDevices.h
+++ b/src/alpaka/AlpakaCore/alpakaDevices.h
@@ -11,31 +11,17 @@
 
 namespace cms::alpakatools {
 
-  // alpaka host device
+  // alpaka host platform and device
   inline const alpaka_common::PlatformHost platformHost{};
   inline const alpaka_common::DevHost host = alpaka::getDevByIdx(platformHost, 0u);
 
-  // alpaka accelerator devices
+  // alpaka accelerator platform and devices
+  // these objects are filled by a call to cms::alpakatools::initialise<TPlatform>()
+  template <typename TPlatform>
+  inline std::optional<TPlatform> platform;
+
   template <typename TPlatform>
   inline std::vector<alpaka::Dev<TPlatform>> devices;
-
-  template <typename TPlatform>
-  std::vector<alpaka::Dev<TPlatform>> enumerate() {
-    assert(getDeviceIndex(host) == 0u);
-
-    using Device = alpaka::Dev<TPlatform>;
-    using Platform = TPlatform;
-    platform = Platform{};
-
-    std::vector<Device> devices;
-    uint32_t n = alpaka::getDevCount(*platform);
-    devices.reserve(n);
-    for (uint32_t i = 0; i < n; ++i) {
-      devices.push_back(alpaka::getDevByIdx(*platform, i));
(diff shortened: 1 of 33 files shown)


src/alpaka/AlpakaCore/alpakaDevices.h
```diff
@@ -11,29 +11,17 @@
...
-    std::vector<Device> devices;
-    uint32_t n = alpaka::getDevCount<Platform>();
-    devices.reserve(n);
-    for (uint32_t i = 0; i < n; ++i) {
-      devices.push_back(alpaka::getDevByIdx<Platform>(i));
-      assert(getDeviceIndex(devices.back()) == static_cast<int>(i));
-    }
-    return devices;
-  }
+  inline std::vector<alpaka::Dev<TPlatform>> devices;
```
makortel, line 24: Would it make sense (now) to move to the same approach as in CMSSW, i.e. have a function that stores the global object in a function-scope `static` variable and returns a (const?) reference to the object?

makortel (commented):
What are the thread-safety guarantees of the Platform objects?

fwyzard:
> What are the thread-safety guarantees of the Platform objects?

Most platform types (all but SYCL IIRC) are empty objects, so they are implicitly thread safe.

I _think_ the SYCL platforms are thread safe because the underlying SYCL objects are thread safe - but I have not thought too much about this.

src/alpaka/AlpakaCore/alpakaDevices.h
fwyzard: I think so, but I would need to try and see if anything breaks (and then fix it).

fwyzard: Something like
```c++
namespace cms::alpakatools {
      
  // return the alpaka accelerator platform for the given platform
  template <typename TPlatform, typename = std::enable_if_t<alpaka::isPlatform<TPlatform>>>
  inline TPlatform const& platform() {
    static const auto platform = TPlatform{};
    return platform;
  } 
  
  // return the alpaka accelerator devices for the given platform
  template <typename TPlatform, typename = std::enable_if_t<alpaka::isPlatform<TPlatform>>>
  inline std::vector<alpaka::Dev<TPlatform>> const& devices() {
    static const auto devices = alpaka::getDevs(platform());
    return devices;
  }

}  // namespace cms::alpakatools
```
?

I would also take the opportunity to try and make then not `inline`, to avoid problems with linking.

fwyzard: An attempt is in https://github.com/cms-patatrack/pixeltrack-standalone/pull/404 .

PR #406, opened 2023-08-27
[sycl] Install Intel OpenCL CPU runtime 2022.14.8.0.04 as an external

fwyzard:


fwyzard:
@AuroraPerego can you try if this works for you ?

AuroraPerego:
Thanks for adding it :)
The only issue I see is that `INTELOCLCPU_BASE` is used at line 162 but defined at line 266. If I move the definition before line 162, everything works. I'll try it also with alpaka.

PR #408, opened 2024-01-02
Add `-rpath` to the shared libraries

fwyzard:
On Ubuntu 22.04 using the system compiler (gcc 11.4.0) the calls to `dlopen()` fail because the plugins are not able to find the shared libraries.

For example
```
$ ./cudatest
Found 1 devices
  what():  unable to load /home/fwyzard/src/pixeltrack-standalone/lib/cudatest/pluginTest1.so because libCUDACore.so: cannot open shared object file: No such file or directory
Aborted (core dumped)
```

And according to `ldd`:
```
$ ldd lib/cudatest/pluginTest1.so
        linux-vdso.so.1 (0x00007ffc3977e000)
        libFramework.so => not found
        libCUDACore.so => not found
        libDataFormats.so => not found
        libtbb.so.12 => /home/fwyzard/src/pixeltrack-standalone/external/tbb/lib/libtbb.so.12 (0x00007fd0bb44f000)
        libcudart.so.12 => /usr/local/cuda/lib64/libcudart.so.12 (0x00007fd0bb000000)
        libbacktrace.so.0 => /home/fwyzard/src/pixeltrack-standalone/external/libbacktrace/lib/libbacktrace.so.0 (0x00007fd0bb43b000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fd0bac00000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fd0bb41b000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd0ba800000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fd0bb4f9000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fd0bb414000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fd0bb40f000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fd0bb40a000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fd0bb323000)
```

Passing the `-rpath` option when building the plugins fixes the issue:
```
$ ./cudatest 
Found 1 devices
Processing 1000 events, with 1 concurrent events and 1 threads.
TestProducer2::acquire Event 1 stream 0 array 0x7f317f200000
TestProducer3 Event 1 stream 0
TestProducer2::produce Event 1 stream 0
...
Processed 1000 events in 1.066554e+01 seconds, throughput 93.7599 events/s, CPU usage per thread: 14.8%
```

For simplicity and consistency, it is now passed when building all shared libraries.

### DIFF
--- a/src/alpaka/Makefile
+++ b/src/alpaka/Makefile
@@ -38,8 +38,7 @@ endif
 ifdef ROCM_BASE
 MY_CXXFLAGS += -DALPAKA_ACC_GPU_HIP_PRESENT -DALPAKA_ACC_GPU_HIP_ONLY_MODE
 endif
-MY_LDFLAGS := -ldl -Wl,-rpath,$(LIB_DIR)/$(TARGET_NAME)
-LIB_LDFLAGS := -L$(LIB_DIR)/$(TARGET_NAME)
+MY_LDFLAGS := -Wl,-rpath,$(LIB_DIR)/$(TARGET_NAME) -L$(LIB_DIR)/$(TARGET_NAME)
 
 ALL_DEPENDS := $(EXE_DEP)
 # Files for libraries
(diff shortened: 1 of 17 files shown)


src/alpaka/Makefile
```diff
@@ -38,8 +38,7 @@ endif
 ifdef ROCM_BASE
 MY_CXXFLAGS += -DALPAKA_ACC_GPU_HIP_PRESENT -DALPAKA_ACC_GPU_HIP_ONLY_MODE
 endif
-MY_LDFLAGS := -ldl -Wl,-rpath,$(LIB_DIR)/$(TARGET_NAME)
-LIB_LDFLAGS := -L$(LIB_DIR)/$(TARGET_NAME)
+MY_LDFLAGS := -ldl -Wl,-rpath,$(LIB_DIR)/$(TARGET_NAME) -L$(LIB_DIR)/$(TARGET_NAME)
```
makortel, line 41: I suppose strictly speaking `-ldl` is needed only for the `$(TARGET)`, but doesn't really harm for the shared libraries.

makortel (approved):
Visually looks ok

src/alpaka/Makefile
fwyzard: I guess I can move it explicitly to the `$(TARGET)` line - though of course it doesn't harm elsewhere.

makortel: I don't have preferences to either way.

fwyzard: ... we already have `-ldl` in the global `$(LDFLAGS)`, so I will just remove it from the `$(MY_LDFLAGS)`

PR #412, opened 2025-04-25
[julia] Port the serial version to Julia

fwyzard:
Implement a Julia version of the serial implementation of the CMS Patatrack pixel track reconstruction.

Include additional documentation of the reconstruction and its Julia implementation.

Corresponds to the code from https://github.com/Maya-Ali0/CMS-Julia as of commit 7e66e6afeb78cf6b0e62fa172ce104d6e2ecf033 .

PR #413, opened 2025-04-28
[julia] fix some obvious code that was not runnable, and format a bit for the files touched

Moelf:
this is a follow up of comments in #412


src/julia-serial/DataFormats/data_formats.jl
```diff
@@ -17,13 +17,9 @@ module dataFormats
     const MAX_FED_ID = 4096
 
     function inrange(fed_id::Int)
-        if fed_id >= MIN_SiPixel_uTCA_FED_ID && Fed_id <= MAX_SiPixel_uTCA_FED_ID
```
Moelf, line 20: notice the `Fed_id` was a typo.

Moelf:
overall, it feels like we need runnable `test/runtests.jl` and look at the coverage to determine what's tested what's not tested.

Here, a few errors are simply that code won't run at all if tested, such as wrong variable name compared to input, and calling methods don't exist (`::Dict[idx, fallback]`)

Moelf:
bump

PR #416, opened 2025-07-04
Removes non-ascii characters.

ericcano:
They lead to this compilation error with CUDA 12.8:
```
ptxas fatal   : Unexpected non-ASCII character encountered on line 19136
```


fwyzard:
hi @ericcano, I can reproduce the error, but I would actually like to keep the correct comments.

How about we remove `--source-in-ptx` from the compilation flags ?
I don't think we actually gain much from it, and it seems to fix the `ptxas` issue as well.

fwyzard:
As the ACTS developers put it:

> This means that you are compiling with a version later than CUDA 12.8 which has not been fixed and you are running in debug mode. This error is due to a bug in the CUDA toolkit. Either downgrade to CUDA 12.6 or disable the debug build.

I'll remove `--source-in-ptx` until NVIDIA gets their act together and brings CUDA to the 90's.

PR #421, opened 2025-07-24
Khaled branch

MohamadKhaledCharaf:
Improve Serial Julia Version: Cleanup, Fixes, and Constants Refinement

- **Fixed CA constant initialization** when `GPU_SMALL_EVENTS` is not defined, ensuring correct behavior across configurations.
- **Disabled task-local caching** for `nearestNeighbor` and `nearestNeighborNumber` since caching was found to **degrade performance**.
- **Removed redundant `makeHits` call**, avoiding unnecessary overhead from duplicate computation.


src/julia-serial/main.jl
```diff
@@ -101,9 +101,9 @@ function julia_main()::Cint
         ed_modules = [
             "BeamSpotToPOD",
             "SiPixelRawToClusterCUDA",
-            "SiPixelRecHitCUDA",
-            "CAHitNtupletCUDA",
-            "PixelVertexProducerCUDA"
+            #"SiPixelRecHitCUDA",
+            #"CAHitNtupletCUDA",
+            #"PixelVertexProducerCUDA"
```
fwyzard, line 106: @MohamadKhaledCharaf are you sure about this change ?

PR #432, opened 2026-08-25
[Compatibility] Fixes for CUDA 13.3, GCC 16.2

kopperp:
CUDA uses the environment variable `NVCC_CCBIN` to specify the host compiler used by nvcc, e.g., in case of an incompatible system GCC version. Respect this variable in the Makefile and add support for Hopper and Blackwell in Kokkos.

Also remove an unused variable in SiPixelClusterizer/gpuClustering.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,10 +35,13 @@ HOST_CXXFLAGS := -O2 -fPIC -fdiagnostics-show-option -felide-constructors -fmess
 # Compiler flags supported by GCC but not by the LLVM-based compilers (clang, hipcc, icpx, etc.)
 LLVM_UNSUPPORTED_CXXFLAGS := --param vect-max-version-for-alias-checks=50 -Werror=format-contains-nul -Wno-non-template-friend -Werror=return-local-addr -Werror=unused-but-set-variable
 
+# In case nvcc uses a different compiler than CXX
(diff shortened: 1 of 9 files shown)


fwyzard:
@kopperp can you split this into 3 PRs
  - code clean up
  - new architectures
  - different host compiler

?

I'm actually not convinced about the last one.

kopperp:
Will do. The last one is actually needed on current Fedora 44 since CUDA 13.3 only supports GCC<=15, but `CXX` defaults to `/usr/bin/g++` (GCC 16.2.1).
```
In file included from /usr/local/cuda/include/cuda_runtime.h:82,
                 from <command-line>:
/usr/local/cuda/include/crt/host_config.h:137:2: error: #error -- unsupported GNU version! gcc versions later than 15 are not supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
```
See also the [CUDA documentation](https://docs.nvidia.com/cuda/cuda-programming-guide/02-basics/nvcc.html#host-code-compilation-notes).

fwyzard:
> The last one is actually needed on current Fedora 44 since CUDA 13.3 only supports GCC<=15, but `CXX` defaults to `/usr/bin/g++` (GCC 16.2.1).

Yes - but I would simply set `CXX` to a compiler supported by CUDA.

PR #434, opened 2026-08-25
Kokkos: Add Hopper and Blackwell architectures

kopperp:


fwyzard:
Thanks @kopperp .

Can you add also sm_89 (Lovelace, L4, L40S, etc.) ?

PR #435, opened 2026-08-25
nvcc: Allow host compiler different to CXX

kopperp:
CUDA uses the environment variable `NVCC_CCBIN` to specify the host compiler used by nvcc, e.g., in case of an incompatible system GCC version. Respect this variable in the Makefile.


fwyzard:
Can't we simple set `CXX` to a compiler supported by CUDA ?

We try to avoid mixing compilers 🤷🏻‍♂️

kopperp:
Sure, that also works. You avoid mixing compilers but you need to change `CXX` when entering/leaving the project. 

`NVCC_CCBIN` can be set globally and is only active for CUDA, with all the upsides and downsides. This MR doesn't really favor any choice, just giving you the option 🤷‍♂️

kopperp:
Actually, with the current Makefile, setting `CXX` has no effect since we hardcode to `g++`. Should we go the `CXX` route then in a new PR?

fwyzard:
`make CXX=/path/to/another/compiler` should work ?

kopperp:
> `make CXX=/path/to/another/compiler` should work ?

`make CXX='g++-15' cuda` does indeed work, the `CXX` environment variable does not.

